PR #15, opened 2016-03-14
Fix a few problems that prevented me from building correctly.

edmcman:


agurfinkel:
@edmcman @lememta Travis is broken. Somehow seahorn is not build in Release mode, while LLVM is. This causes a symbol to go missing.

lememta:
Fixed the problem.

PR #20, opened 2016-04-21
Add a regression test suite.

wuestholz:
I added a regression test suite. It uses LLVM's lit tool (also used by other tools such as Boogie, Dafny and KLEE) to run the tests.


agurfinkel:
This is fantastic! Thank you.

Do you know how to use LLVM's LNT tool for regression testing as well?
http://llvm.org/docs/lnt/

wuestholz:
@agurfinkel No, I didn't know about this. Looks interesting though. Thanks!

agurfinkel:
It supposed to help with performance regression, i.e., difference in execution time and not just expected output. Would be really cool to have this for seahorn as well.

wuestholz:
@agurfinkel Yeah, for Boogie and Dafny I used to manually check the running times of the continuous build (which includes running regression tests) once in a while. However, something more automated would be quite nice, especially since running times of the build would also increase if new tests were added.

PR #30, opened 2016-06-13
internalize global variables

jiaqi-liu:


lib/Transforms/Instrumentation/KleeInternalize.cc
```diff
@@ -151,13 +154,25 @@ namespace
...
-	// TODO: update callgraph
+          retValue = Builder.CreateTruncOrBitCast(retValue, retTy);
+  
+  // TODO: update callgraph
         Builder.CreateRet (retValue);
       }
     }
-    
+
+    void InternalizeVariables (Module& M)
```
agurfinkel: Please start method name with lower case

lib/Transforms/Instrumentation/KleeInternalize.cc
```diff
@@ -151,13 +154,25 @@ namespace
...
+    {
+       for (Module::global_iterator I = M.global_begin(), E = M.global_end();
+          I != E; ++I) {
+        GlobalVariable *GV = &*I;
+        if (GV->isConstant() || GV->hasInitializer())
+          continue;
+        GV->setInitializer(Constant::getNullValue(GV->getType()->getElementType()));
```
agurfinkel: Please add a check for whether the name is in `m_externalNames` so that some external variables can remain external.

agurfinkel:
merged. thanks!

PR #31, opened 2016-06-15
Fix casting bug in KleeInternalize

edmcman:


edmcman:
The truncating cast was failing when a pointer was being returned.

agurfinkel:
That's not the right fix. Upcasting and truncating was needed for i1 which
have store size of i8.

Give me a minute.

edmcman:
I am not sure that you as the pass writer are responsible for enforcing the store size here.

agurfinkel:
We are copying a scalar into memory, then using `klee_make_symbolic` to
initialize the memory.
The size of the scalar in memory is its store size. We have to go through
that, otherwise we might be allocating one bit on the stack, which is not
allowed.

agurfinkel:
Another fix committed.

PR #33, opened 2016-06-22
Houdini Pass

Chenguang-Zhu:
The new added Houdini pass; Fix the problem of initializing workList from Rule DB; Fix the design of weakening rule head candidate until it's true.


include/seahorn/Houdini.hh
```diff
@@ -0,0 +1,68 @@
...
+    virtual const char* getPassName () const {return "Houdini";}
+
+    void printDB (const HornClauseDB &db);
+    void printHello ();
+    Expr guessCandidate(Expr pred);
+    int testGuessCandidate(Expr pred);
+    bool validateRule(Expr cand_app, HornifyModule &hm);
+    void workListAlgo(HornClauseDB &db, HornifyModule &hm);
+
+    template<typename OutputIterator>
```
agurfinkel: Please break the methods into public and private. 
Move as much out of the header file as you can (especially support functions and structs such as `IsBVar`)

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+
+  bool Houdini::runOnModule (Module &M)
+  {
+    HornifyModule &hm = getAnalysis<HornifyModule> ();
+
+    //load the Horn clause database
+    auto &db = hm.getHornClauseDB ();
+    db.buildIndexes ();
+
+    workListAlgo(db, hm);
```
agurfinkel: rename `workListAlgo` into a more descriptive name (e.g., `run` )

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+    AU.addRequired<HornifyModule> ();
+    AU.setPreservesAll();
+  }
+
+  void Houdini::printDB (const HornClauseDB &db)
+  {
+    outs () << db << "\n";
+  }
+
+  void Houdini::printHello ()
```
agurfinkel: remove

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+    return false;
+  }
+
+  void Houdini::getAnalysisUsage (AnalysisUsage &AU) const
+  {
+    AU.addRequired<HornifyModule> ();
+    AU.setPreservesAll();
+  }
+
+  void Houdini::printDB (const HornClauseDB &db)
```
agurfinkel: remove?

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  void Houdini::printHello ()
+  {
+    outs () << "Hello there.\n";
+  }
+
+  void Houdini::getUseRuleSet(HornClauseDB &db)
+  {
+	  auto &workList = db.getRules();
```
agurfinkel: this seems wrong. The worklist should not be a reference to the rules in the database.

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+
+	return cand;
+  }
+
+  #define SAT true
+  #define UNSAT false
+
+  void Houdini::workListAlgo(HornClauseDB &db, HornifyModule &hm)
+  {
+	  //auto &workList = db.getRules();
```
agurfinkel: remove

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+
+	  while (!workList.empty())
+	  {
+		  outs() << "[WORKLIST SIZE]: " << workList.size() << "\n";
+		  HornRule r = workList.front();
+
+		  outs() << "[RULE HEAD]: " << *(r.head()) << "\n";
+		  outs() << "[RULE BODY]: " << *(r.body()) << "\n";
+
+		  //generate candidate for rule head
```
agurfinkel: I am confused by this function, starting from this point. It probably needs to be broken into more manageable pieces. 

For now, I don't understand why the candidate is being guessed in the main Houdini loop. The guessing of candidates should be factored out. It is expected that many different heuristics will be plugged in there in the future.

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+		  if(validateRule(cand_app, hm) == SAT)
+		  {
+			 //add rules in db.use(r.head()) to worklist
+			 Expr head_app = r.head();
+			 Expr head_fdecl = bind::fname(head_app);
+			 outs() << "[USE SET SIZE]: " << db.use(head_fdecl).size() << "\n";
+			 for(auto r_use : db.use(head_fdecl))
+		     {
+				 bool isExist = false;
+				 for(HornClauseDB::RuleVector::iterator itr=workList.begin(); itr!=workList.end(); ++itr)
```
agurfinkel: Use STL instead of iterating yourself.

This piece of code suggests that worklist is not indexed correctly. Adding / removing elements from the worklist should be efficient.

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+						 break;
+					 }
+				 }
+				 if(isExist == false)
+				 {
+					 workList.push_back(*r_use);
+				 }
+			 }
+			 //weaken candidate for r.head()
+			 while (validateRule(cand_app, hm) == SAT)
```
agurfinkel: You should expect three answers from an SMT solver (or any function that depends on the solver). It is always possible that the solver returns `unknown` and this possibility has to be worked around somehow.

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+					 head_cand_app = mk<TRUE>(head_cand_app->efac());
+				 }
+				 else
+				 {
+					 ExprVector new_head_vec;
+					 for(auto it = head_cand_app->args_begin (), end = head_cand_app->args_end (); it != end; ++it)
+					 {
+					 	new_head_vec.push_back(*it);
+					 }
+					 new_head_vec.erase(new_head_vec.begin());
```
agurfinkel: erasing first element of a vector is expensive. I don't understand what this code is doing.

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+	  solver.toSmtLib(outs());
+	  if(solver.solve())
+	  {
+		  outs() << "SAT\n";
+	  }
+	  else
+	  {
+		  outs() << "UNSAT\n";
+	  }
+	  return solver.solve();
```
agurfinkel: should not call `solve()`. Can be expensive and can change state.

Need a better architecture where the solver is not being destroyed and re-created each time.

I don't see where the transition relation (the body of the rule) is loaded. Perhaps `validateRule()` is a misleading name since the input is not a rule but an expression.

include/seahorn/Houdini.hh
Chenguang-Zhu: Fixed.

lib/seahorn/Houdini.cc
Chenguang-Zhu: Fixed

lib/seahorn/Houdini.cc
Chenguang-Zhu: Fixed

lib/seahorn/Houdini.cc
Chenguang-Zhu: Deleted the test function getUseRuleSet()

lib/seahorn/Houdini.cc
Chenguang-Zhu: The whole program has been refactored.

lib/seahorn/Houdini.cc
Chenguang-Zhu: Fixed the design of guessCandidate(), now it is called only once at the beginning to construct a global map from predicates to candidates.

lib/seahorn/Houdini.cc
Chenguang-Zhu: Fixed.

lib/seahorn/Houdini.cc
Chenguang-Zhu: Fixed the desgin of validateRule(), it now takes a rule as input and return whether the whole replaced expression of rule is SAT

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+	  workList.insert(workList.end(), db.getRules().begin(), db.getRules().end());
+
+	  while(!workList.empty())
+	  {
+		  HornRule r = workList.front();
+		  errs() << "RULE HEAD: " << *(r.head()) << "\n";
+		  errs() << "RULE BODY: " << *(r.body()) << "\n";
+		  while (validateRule(r, db, hm) == SAT)
+		  {
+			  workList = addUsedRulesBackToWorkList(db, workList, r.head());
```
agurfinkel: the assignment here seems wrong. `addUsedRulesBackToWorkList` probably modifies the worklist in place.

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+  void Houdini::runHoudini(HornClauseDB &db, HornifyModule &hm)
+  {
+	  guessCandidate(db);
+
+	  HornClauseDB::RuleVector workList;
+	  workList.insert(workList.end(), db.getRules().begin(), db.getRules().end());
+
+	  while(!workList.empty())
+	  {
+		  HornRule r = workList.front();
```
agurfinkel: better to erase the front right here. Probably a linked list such as `std::list` would be better for the worklist than a vector

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+		  errs() << "RULE HEAD: " << *(r.head()) << "\n";
```
agurfinkel: put under `LOG` macro. See code for usage examples.

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+	  get_all_pred_apps(ruleBody, db, std::back_inserter(body_pred_apps));
+
+	  for(ExprVector::iterator itr = body_pred_apps.begin(); itr != body_pred_apps.end(); ++itr)
+	  {
+		  body_map.insert(std::pair<Expr, Expr>(*itr, applyActualArgsToCand(*itr)));
+	  }
+
+	  Expr body_cand_app = replace(ruleBody, body_map);
+
+	  Expr whole_cand = mk<AND>(neg_ruleHead_cand_app, body_cand_app);
```
agurfinkel: You can use `solver.assertExpr` multiple times instead of conjoining all expressions.

You want to separate asserting the transition relation (part of body without relations), from asserting relation in the body, and relation in the head. It will not make a difference now, but will be useful in the future since the transition relation does not change and can be left in the solver.

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+
+	  errs() << "WHOLE CANDIDATE: " << *whole_cand << "\n";
+
+	  ZSolver<EZ3> solver(hm.getZContext ());
+	  solver.assertExpr(whole_cand);
+	  solver.toSmtLib(errs());
+	  bool isSat = solver.solve();
```
agurfinkel: solver can produce 3 possible values: true, false, indeterminate . The last one indicates that the solver given up for some reason.

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+	  		return;
+	  }
+	  if(!isOpX<AND>(ruleHead_cand))
+	  {
+	  		Expr weaken_cand = mk<TRUE>(ruleHead_cand->efac());
+	  		currentCandidates[ruleHead_rel] = weaken_cand;
+	  }
+	  else
+	  {
+	  		ExprVector head_cand_args;
```
agurfinkel: This seems wrong. The weakening should be based on the model that the solver has produced. This seems to weaken by removing an arbitrary lemma. Is that right?

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+	  			currentCandidates[ruleHead_rel] = weaken_cand;
+	  		}
+	  }
+  }
+
+  /*
+   * Given a rule head, extract all rules using it in body, then add all such rules to the end of workList
+   */
+  HornClauseDB::RuleVector Houdini::addUsedRulesBackToWorkList(HornClauseDB &db, HornClauseDB::RuleVector workList, Expr ruleHead_app)
+  {
```
agurfinkel: You probably want to implement the worklist as a set so that you can simplify deciding when something is already in the worklist.

lib/seahorn/Houdini.cc
Chenguang-Zhu: I'm a little confused about this comment, shouldn't we update the workList itself?

lib/seahorn/Houdini.cc
Chenguang-Zhu: Fixed this issue, now use std::list

lib/seahorn/Houdini.cc
Chenguang-Zhu: Thanks Arie. I think I misunderstand the design of weakenCandidate(), So I hope to ask if solver.assertExpr(e) says e is SAT. Then should I use getModel() to get the model that makes the results true? Then how do I get the lemma that I should remove from the model? Thanks.

lib/seahorn/Houdini.cc
Chenguang-Zhu: If the solver returns indeterminate, should I also weaken the candidate?

lib/seahorn/Houdini.cc
Chenguang-Zhu: Thanks Arie. I think I misunderstand the design of weakenCandidate(), So I hope to ask if solver.assertExpr(e) says e is SAT. Then should I use getModel() to get the model that makes the results true? Then how do I get the lemma that I should remove from the model? Thanks.

lib/seahorn/Houdini.cc
Chenguang-Zhu: I hope to ask how to use solver.assertExpr() multiple times in each round? In each round I pick a rule, replace all predicates with candidates, and then I should validate the negation of its head and its body. So if I want to use solver.assertExpr() multiple times, should I use it for (!head) and for body separately or should I use it for each predicate(replaced with candidate) separately and use it once for the transition relation? Thanks.

lib/seahorn/Houdini.cc
Chenguang-Zhu: std::set will sort its content automatically. I'm afraid by using that when we push some rules back to worklist it will disorder the rules.

agurfinkel: You can use set for uniqueness and list for order and update both as
needed.

lib/seahorn/Houdini.cc
agurfinkel: Yes, but you are updating it inside the function already.

Chenguang-Zhu: OK. This one is fixed now. I pass reference of worklist into it.

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,292 @@
...
+
+	  for (HornClauseDB::RuleVector::iterator it = db.getRules().begin(); it != db.getRules().end(); ++it)
+	  {
+		  HornRule r = *it;
+		  workList.push_back(r);
+	  }
+
+	  while (!workList.empty())
+	  {
+		  outs() << "[WORKLIST SIZE]: " << workList.size() << "\n";
```
Chenguang-Zhu: Fixed.

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,294 @@
...
+		cand = mknary<AND>(bins.begin(), bins.end());
+	}
+
+	return cand;
+  }
+
+  #define SAT true
+  #define UNSAT false
+
+  std::map<Expr, Expr> Houdini::currentCandidates;	//a map from predicates to candidates
```
agurfinkel: this should not be here.`currentCandidates` should not be a static field.

lib/seahorn/Houdini.cc
agurfinkel: You can evaluate the lemma in the model using `ZModel::eval`, then check whether it is `true` or `false`.

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,294 @@
...
+	  while(!workList.empty())
+	  {
+		  LOG("houdini", errs() << "WORKLIST SIZE: " << workList.size() << "\n";);
+		  HornRule r = workList.front();
+		  workList.pop_front();
+		  LOG("houdini", errs() << "RULE HEAD: " << *(r.head()) << "\n";);
+		  LOG("houdini", errs() << "RULE BODY: " << *(r.body()) << "\n";);
+		  while (validateRule(r, db, hm) == SAT)
+		  {
+			  std::list<HornRule> &ref_workList = workList;
```
agurfinkel: No need to create a dummy reference variable. Pass `worklist` directly. It will be passed by reference because that is how the argument is declared in `addUsedRulesBackToWorkList`.

lib/seahorn/Houdini.cc
agurfinkel: Think of the solver as a vector. You can add expressions to it, one at a time, using `assertExpr()`. I would use it once for each predicate, once for the body with predicates removed, and once for head. Note that you can remove predicates from the body by replacing their application with `true`, just like you replace them now with candidates.

Chenguang-Zhu: Thanks. I understand it. Now the code has been updated to use multiple assertExpr().

lib/seahorn/Houdini.cc
Chenguang-Zhu: Fixed.

lib/seahorn/Houdini.cc
Chenguang-Zhu: I tried not to use static field. Instead, using 

private:
    std::map<Expr,Expr> currentCandidates;

in Houdini.hh file.
But it causes error when destroy the map. It shows:

Program received signal SIGSEGV, Segmentation fault.
~intrusive_ptr (this=0x1f263e8, __in_chrg=<optimized out>) at /usr/local/boost_1_61_0/boost/smart_ptr/intrusive_ptr.hpp:97
97	        if( px != 0 ) intrusive_ptr_release( px );

What should I do with this error?

agurfinkel: debug it. why do you think this happens when map is destroyed. is
expression factory destroyed before all expressions are?

look at similar code. what does HornSolver do with expressions it creates.
Are there other passes that keep expressions around, etc.

cheers
arie

lib/seahorn/Houdini.cc
Chenguang-Zhu: Hi Arie,
I pass the expression (B1:INT)<=(B0:INT) to the model.eval(expr) function then it returns (!(B1:INT)<=(B0:INT)). I am confused since model should be the assignment to all variables. So I thought eval () is to apply values to expressions and see the result. But it seems I'm wrong. So I hope to ask what does the eval() do ? Thanks.

Chenguang-Zhu: Never mind. I know what I did wrong.

Chenguang-Zhu: Fixed. Now we use model to evaluate each lemma of rule head's candidate. In each round the program picks a lemma that is evaluated to false by model, removes it from the candidate, then try to validate the rule again.

lib/seahorn/Houdini.cc
Chenguang-Zhu: If solver returns indeterminate for a rule, then how to deal with it? I think maybe we should also try to weaken it until solver can say UNSAT?

agurfinkel: if the result is neither true nor false you can still use the model to try
to throw out a lemma. If nothing works, remove a lemma arbitrarily. In the
worst case, the invariant is weakened to true and the algorithm stops.

Chenguang-Zhu: Thanks Arie. Now it can deal with 3 possible return values.

include/seahorn/HornClauseDB.hh
```diff
@@ -129,9 +132,18 @@ namespace seahorn
     /// resets all indexes
     void resetIndexes ();
 
+    /// callgraph
+    typedef std::map<Expr, expr_set_type > callgraph_type;
```
agurfinkel: I think the callgraph should be factored out of `HornClauseDB`. The database is just a collection of clauses. Structures that build on it should be defined separately on the side.

lib/seahorn/HornClauseDB.cc
```diff
@@ -31,6 +36,67 @@ namespace seahorn
...
+
+    // XXX: we are looking for a predicate that corresponds to the
+    // entry block of main. It is not enough to search for a predicate
+    // with no callers since predicates from each function entry won't
+    // have callers.
+    for (auto p: getRelations ()) 
+    {
+      Expr fname = bind::fname(p);
+      std::string sname = boost::lexical_cast<std::string> (fname);
+      if (boost::starts_with(sname, "main") && callers(p).size() == 0) {
```
agurfinkel: There should be a better way. `main` is not always the entry point. 
This is ok as a workaround, as long as it is properly exposed. That is, not part of a generic call-graph functionality.

lib/seahorn/Houdini.cc
```diff
@@ -0,0 +1,662 @@
...
+	  {
+		LOG("houdini", errs() << "PRED: " << *(it->first) << "\n";);
+		LOG("houdini", errs() << "CAND: " << *(it->second) << "\n";);
+	  }
+
+	  std::list<HornRule> workList;
+	  workList.insert(workList.end(), db.getRules().begin(), db.getRules().end());
+	  workList.reverse();
+
+	  if (config == EACH_RULE_A_SOLVER)
```
agurfinkel: why not use sub-classing or some other object-oriented or template-based solution?

include/seahorn/HornClauseDB.hh
Chenguang-Zhu: Fixed. created a new class HornClauseDBCallGraph.

lib/seahorn/Houdini.cc
Chenguang-Zhu: Finished the refactoring. Now use abstract class HoudiniContext and make each configuration as its sub-class.

agurfinkel:
I'm merging this pull request. However, I don't think WTO is used in the algorithm as we envisioned it. The current algorithm is worklist based, where the worklist is processed in some ad-hoc order. With WTO, we should be able to process the list in a deterministic order, based on WTO components.

agurfinkel:
Also, please split the predicate discovery from outside Houdini. Once we finalize the algorithm, the main point of variation will be in predicate discovery. I guess it is better to create issues for these other than comments :)

PR #36, opened 2016-07-11
Output Slice

jiaqi-liu:
output slice based on bmctrace


lib/seahorn/HornCex.cc
```diff
@@ -373,5 +386,62 @@ namespace seahorn
     out.keep ();
   }
 
+  static void sliceErrorTrace(Function &F, BmcTrace &trace) {
```
agurfinkel: Why could you not use reduceToRegion instead of re-implementing it?

It would also be good to have a slice through cut-points only as well.

agurfinkel: That is, use `cpTrace` to construct the slice.

jiaqi-liu: I was aware of the code-reuse issue here. However, sliceErrorTrace has different semantics than reduceToRegion, for example, reduceToRegion assumes the input is a precise region while BmcTrace may omit some basic blocks due to abstraction. Not sure of the impact on changing the semantics of reduceToRegion directly.

agurfinkel: I am not sure I see the difference in semantics. I think at this point, there is no abstraction done by the BMC trace. How does the difference manifest itself in the code? If the change is very small, we might merge it with `reduceToRegion`  and use a flag to separate between the two semantics.

jiaqi-liu: Please checkout the new commit

lib/seahorn/Bmc.cc
```diff
@@ -216,6 +216,14 @@ namespace seahorn
     return m_model.eval (v, complete);
   }
   
+  DenseSet<const BasicBlock*> BmcTrace::getRegion()
```
agurfinkel: This does not look right. `m_bbs` is the sequence of all basic blocks in the bmc trace. 
There is no need to return them as a set. Moreover, passing sets by value in expensive.
If you need a set of bbs, traverse the trace and put them into the set in the client.

lib/seahorn/HornCex.cc
```diff
@@ -220,11 +227,25 @@ namespace seahorn
...
+    if (!BmcSliceOutputFile.empty()) {
+      llvm::DenseSet<const llvm::BasicBlock*> region(trace.getRegion());
+      reduceToRegion(F, region, false);
+      dumpLLVMBitcode(M, BmcSliceOutputFile.c_str());
       return true;
     }
+
+    if (!CpSliceOutputFile.empty())
+    {
+      DenseSet<const BasicBlock*> region;
```
agurfinkel: this looks good, but please indent properly

lib/seahorn/HornCex.cc
```diff
@@ -220,11 +227,25 @@ namespace seahorn
...
+      for (const CutPoint* cp:cpTrace)
+      for (CutPoint::const_iterator it=cp->succ_begin();it!=cp->succ_end();it++)
+      for (CpEdge::iterator ite=(*it)->begin();ite!=(*it)->end();ite++)
+        region.insert(&*ite);
+      reduceToRegion(F,region,false);
```
agurfinkel: what is the reason for not deleting the basic blocks that are outside of the trace? Deleting them will enable additional optimizations. Moreover, the blocks are unreachable after the reduction and will be removed by dead code elimination anyways.

lib/Transforms/Utils/Local.cc
```diff
@@ -43,33 +43,36 @@ namespace seahorn
...
+          continue;
+           
         Builder.SetInsertPoint (&BB, br);
         CallInst *ci = Builder.CreateCall
           (kill == s1 ? assumeFn : assumeNotFn, br->getCondition ());
         ci->setDebugLoc (br->getDebugLoc ());
         br->eraseFromParent ();
         Builder.SetInsertPoint (&BB);
         Builder.CreateBr (kill == s1 ? s0 : s1);
+        kill->removePredecessor(&BB);
```
agurfinkel: I think this has to be done before the branch instruction is deleted. The documentation says `Note that this should be called while the predecessor still refers to this block`. I think this means that `removePredecessor` is called first, then the branch is removed. Not sure it is necessary if we are deleting `kill` basic block anyways.

lib/seahorn/HornCex.cc
jiaqi-liu: Good question. I debugged it for a while and this is the final code I came up with. First of all, the original code of directly deleting basic block crashes. One of the reasons is bad PHI node in successors. To resolve the imprecise implementation of deleteing, I changed it to a existing util function (DeleteDeadBlock) in a commit. This solved the phi node issue, but still crashes in some files. I had a few speculations on the causes. Overall, I think it is better to leave the optimization to an additional pass.

lib/Transforms/Utils/Local.cc
jiaqi-liu: It doesn't matter. I will change it anyway.

lib/seahorn/Bmc.cc
jiaqi-liu: I was aware of the performance issue. I thought it was not a big deal since it is only called once and not in critical part. But since you brought this up, I'll change it.

lib/seahorn/HornCex.cc
jiaqi-liu: sure.

lib/seahorn/HornCex.cc
agurfinkel: I'd like to get to the bottom of it. If the function fails here, it might be failing somewhere else as well, but we have not noticed.

Can you send me a failing test case so that I can debug it as well?

jiaqi-liu: Of course. By the way, a concern is that some basic blocks may be omitted in cpTrace due to abstraction. In that case, we cannot aggressively delete them all. Do you think it is possible?

agurfinkel: I looked at the original `reduceToRegion` again. I don't see how it can fail because of `PHI` nodes. Only if the region argument to it is not really a region. Using `DeleteDeadBlock` is not right since we might be deleting regions with loops. Not deleting basic blocks probably only masks the actual problem and if we won't fix it here, it will be harder to debug later on.

I do not understand your comment about *abstraction*. There is no abstraction happening here since we are working on the exact code being analyzed. That code might be an abstraction of something else, but that does not matter. 

Please send me an example that fails in `reduceToRegion` when basic blocks are removed. Maybe I can come up with an alternative explanation.

PR #40, opened 2016-08-02
Predabs

Chenguang-Zhu:


include/seahorn/Houdini.hh
```diff
@@ -28,22 +28,32 @@ namespace seahorn
     virtual const char* getPassName () const {return "Houdini";}
   };
 
+  class GuessCandidates
```
agurfinkel: @Chenguang-Zhu , please re-factor `GuessCandidates` so that each guessing strategy is a separate class and that the choice of the strategy can be selected at runtime.

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,370 @@
...
+
+		for(Expr rel : db.getRelations())
+		{
+			predToBiNumMap.insert(std::pair<Expr, int>(rel, index));
+			index++;
+		}
+
+		for(Expr rel : db.getRelations())
+		{
+			outs() << "OLD PRED: " << *rel << "\n";
```
agurfinkel: @Chenguang-Zhu , hide under `LOG` macro

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,370 @@
...
+			//Push boolean types
+			for(int i=0; i < db.getRelations().size(); i++)
+			{
+				new_args.push_back(mk<BOOL_TY>(rel->efac()));
+			}
+			Expr new_rel = mknary<FDECL>(new_args);
+			outs() << "NEW PRED: " << *new_rel << "\n";
+			new_DB.registerRelation(new_rel);
+			oldToNewPredMap.insert(std::pair<Expr, Expr>(rel, new_rel));
+		}
```
agurfinkel: @Chenguang-Zhu , this loop seems wrong. You should iterate over number of predicates, not number of relations in the database. 

It is useful to give the new relation a unique name so that, if desired, old and new relations can be mixed together in one database. You can use `variant::variant (bind::fname(rel), 1)` for example.

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,370 @@
...
+		outs() << "NEW DB: \n";
+		outs() << new_DB << "\n";
+
+		for(HornClauseDB::RuleVector::iterator it = db.getRules().begin(); it != db.getRules().end(); ++it)
+		{
+			HornRule r = *it;
+			outs() << "OLD RULE HEAD: " << *(r.head()) << "\n";
```
agurfinkel: hide all debug print statements under `LOG` macro.

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,370 @@
...
+					int occur_num = relOccurrenceTimesMap[bind::fname(rule_head)];
+					assert(occur_num >= 1);
+					int bool_var_index = it->second;
+					std::ostringstream oss;
+					oss << "b" << bool_var_index;
+					for(int i=0; i<occur_num - 1; i++)
+					{
+						oss << "\'";
+					}
+					Expr bi = bind::boolVar(mkTerm<std::string>(oss.str(), db.getExprFactory ()));
```
agurfinkel: The use of  `bind::boolVar` is deprecated. Use `bind::boolConst` instead.

agurfinkel: Avoid using string names. Use structured expressions instead. In this case, you probably want to use `variant::variant` to create a variant of a boolean variable.

Name the boolean variables based on the name of the predicate. This will be easier to debug. You can number them based on their occurrence in the rule (i.e., first occurrence is 0, second occurrence is 1, etc. Independently of the name).

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,370 @@
...
+				Expr rule_body_cand_app = fAppToCandApp(rule_body_pred);
+				Expr rule_body_bool_var = predAppToBoolVarMap[rule_body_pred];
+				Expr rule_body_iff = mk<IFF>(rule_body_bool_var, rule_body_cand_app);
+				new_body_exprs.push_back(rule_body_iff);
+			}
+			//Extract the constraints
+			Expr constraints = extractTransitionRelation(r, db);
+			new_body_exprs.push_back(constraints);
+			//Construct new body
+			Expr new_rule_body = mknary<AND>(new_body_exprs.begin(), new_body_exprs.end());
```
agurfinkel: Can `new_body_exprs` be empty? Use the other `mknary` that takes the base result as an extra argument.

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,370 @@
...
+					new_rule_head_args.push_back(bi);
+				}
+			}
+
+			Expr new_rule_head = bind::fapp(new_rule_head_rel, new_rule_head_args);
+			outs() << "NEW RULE HEAD: " << *new_rule_head << "\n";
+
+			//construct head iff
+			Expr rule_head_cand_app = fAppToCandApp(rule_head);
+			Expr rule_head_iff = mk<IFF>(predAppToBoolVarMap[rule_head], rule_head_cand_app);
```
agurfinkel: Not sure about this `IFF`. I was expecting a conjunction of definitions.

include/seahorn/PredicateAbstraction.hh
```diff
@@ -14,6 +14,45 @@
 namespace seahorn
 {
 	using namespace llvm;
+
+	class HornDbModel
```
agurfinkel: `HornDbModel` should not depend on `ZFixedPoint`. We might want to use it with other solvers. 
There should be a separate method to convert `ZFixedPoint` answer to a `HornDbModel`. 

I'd like to preserve the old API. To query the model, the use provides an `fapp` and gets back a definition for the predicate based on `fapp` arguments. You can use bound-variables `bind::bvar()` to store the solution internally with `bvar(0)` being the first argument, etc.

agurfinkel: `HornDbModel` should provide construct api such as `addDef (fapp, expr)` that takes an `fapp` and an expression and remembers that `expr` is the definition of the predicate in `fapp`.  Then model converter will be easier to write, no need to break encapsulation.

include/seahorn/PredicateAbstraction.hh
```diff
@@ -14,6 +14,45 @@
...
+		HornDbModel(ExprMap model) : predToSolutions(model) {}
+		virtual ~HornDbModel(){}
+		ExprMap& getRelToSolutionMap() {return predToSolutions;}
+		void setRelToSolutionMap(ExprMap map) {predToSolutions = map;}
+	};
+
+	class HornModelConverter
+	{
+	private:
+		ExprMap boolVarToTermMap;
```
agurfinkel: It probably makes sense to have a map `relation -> bool_const -> term` so that it is possible to use the same boolean constant (such as `p0`) to represent different predicates for different relations.

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -63,22 +66,72 @@ namespace seahorn
...
 		outs () << "\n";
-//
-		printInvars (M);
 
 		return false;
 	}
 
+	bool HornModelConverter::convert (HornDbModel &in, HornDbModel &out, std::map<Expr, Expr> &newToOldPredMap)
+	{
+		ExprMap inRelToSolMap = in.getRelToSolutionMap();
```
agurfinkel: Don't expose solution map. Don't make assumptions on how `HornDbModel` is implemented. I am commenting on this line, but the comment applies throughout the code.

Chenguang-Zhu: Sorry I'm a little confused about this comment. In the convert function we want to convert the abstract relation to the original relation, and convert the boolean variables to the terms. It seems we need such a data structure for iterating over the relations in abstractDB, no matter what the internal implementation is. One alternative design that I can think out is not to return the map, but a list of abs_relations, and for each relation, I use a method such as HornDbModel.getBv(rel) to get its boolean variable. Is this reasonable?

agurfinkel: You should be able to get the list of relations from the database. I think there is already an api for that.
You can then get definition from an abstract model, construct the corresponding definition for the concrete relation and enter it into concrete model.

You definitely do not want anything that is specific to type of relation or arguments. `HornDbModel` should provide the same API as `getDeltaCover` since it is a generic representation of what we already have in the fixedpoint engine.

include/seahorn/PredicateAbstraction.hh
```diff
@@ -0,0 +1,151 @@
...
+		ExprMap relToDefMap;
+		std::map<Expr, ExprVector> currentCandidates;
+	public:
+		HornDbModel() {}
+		void initModelFromFP(HornClauseDB &db, ZFixedPoint<EZ3> &fp);
+		void addDef(Expr fapp, Expr def);
+		Expr getDef(Expr fapp);
+		HornDbModel(ExprMap model) : relToDefMap(model) {}
+		virtual ~HornDbModel(){}
+		void setCurrentCandidates(std::map<Expr, ExprVector> currentCandidates) { this->currentCandidates = currentCandidates; }
```
agurfinkel: map should never be passed not by reference.

include/seahorn/PredicateAbstraction.hh
```diff
@@ -0,0 +1,151 @@
...
+#include "ufo/Expr.hpp"
+#include "ufo/Smt/Z3n.hpp"
+#include "ufo/Smt/EZ3.hh"
+#include "seahorn/HornClauseDBWto.hh"
+
+namespace seahorn
+{
+	using namespace llvm;
+
+	class HornDbModel
```
agurfinkel: `HornDbModel` must be in its own header file. It will be used outside of predicate abstraction as well.

include/seahorn/PredicateAbstraction.hh
```diff
@@ -0,0 +1,151 @@
...
+	{
+	private:
+		ExprMap relToDefMap;
+		std::map<Expr, ExprVector> currentCandidates;
+	public:
+		HornDbModel() {}
+		void initModelFromFP(HornClauseDB &db, ZFixedPoint<EZ3> &fp);
```
agurfinkel: `initModelFromFP` should probably be a function outside of `HornDbModel`

include/seahorn/PredicateAbstraction.hh
```diff
@@ -0,0 +1,151 @@
...
+		void addDef(Expr fapp, Expr def);
```
agurfinkel: add comments for how the function should be used. Applies to other functions in this class as well.

include/seahorn/PredicateAbstraction.hh
```diff
@@ -0,0 +1,151 @@
...
+		Expr getDef(Expr fapp);
+		HornDbModel(ExprMap model) : relToDefMap(model) {}
```
agurfinkel: `ExprMap` should be passed by reference

include/seahorn/PredicateAbstraction.hh
agurfinkel: `map` must be passed by reference.

include/seahorn/PredicateAbstraction.hh
```diff
@@ -0,0 +1,151 @@
...
+		virtual ~HornDbModel(){}
+		void setCurrentCandidates(std::map<Expr, ExprVector> currentCandidates) { this->currentCandidates = currentCandidates; }
+//		ExprMap& getRelToSolutionMap() {return relToDefMap;}
+//		void setRelToSolutionMap(ExprMap map) {relToDefMap = map;}
+	};
+
+	class HornModelConverter
```
agurfinkel: Should be in the same header as `HornDbModel`

include/seahorn/PredicateAbstraction.hh
```diff
@@ -0,0 +1,151 @@
...
+	{
+	private:
+		std::map<Expr, ExprMap> relToBoolToTermMap;
+		std::map<Expr, Expr> newToOldPredMap;
+		HornClauseDB* abs_db;
+	public:
+		PredAbsHornModelConverter() {}
+		bool convert (HornDbModel &in, HornDbModel &out);
+		std::map<Expr, ExprMap>& getRelToBoolToTermMap() {return relToBoolToTermMap;}
+		void setNewToOldPredMap(std::map<Expr, Expr> &newToOldMap) {newToOldPredMap = newToOldMap;}
```
agurfinkel: `newToOldMap` should be marked `const`

include/seahorn/PredicateAbstraction.hh
```diff
@@ -0,0 +1,151 @@
...
+	class PredAbsHornModelConverter : public HornModelConverter
+	{
+	private:
+		std::map<Expr, ExprMap> relToBoolToTermMap;
+		std::map<Expr, Expr> newToOldPredMap;
+		HornClauseDB* abs_db;
+	public:
+		PredAbsHornModelConverter() {}
+		bool convert (HornDbModel &in, HornDbModel &out);
+		std::map<Expr, ExprMap>& getRelToBoolToTermMap() {return relToBoolToTermMap;}
```
agurfinkel: the only public method should be `convert`. The rest are implementation detail that should not be visible to clients.

include/seahorn/PredicateAbstraction.hh
```diff
@@ -0,0 +1,151 @@
...
+	};
+
+	class PredicateAbstractionAnalysis
+	{
+	private:
+	    std::map<Expr, Expr> oldToNewPredMap;
+	    std::map<Expr, Expr> newToOldPredMap;
+	    std::map<Expr, ExprVector> currentCandidates;
+
+	    std::unique_ptr<ufo::ZFixedPoint <ufo::EZ3> >  m_fp;
```
agurfinkel: How is this initialized? We cannot use default parameters on `ZFixedPoint`. It should be properly initialized as in `HornSolver`. Ideally, `HornSolver` should be used to solve the system, not `PredicateAbstractionAnalysis`

include/seahorn/PredicateAbstraction.hh
```diff
@@ -0,0 +1,151 @@
...
+	public:
+	    PredicateAbstractionAnalysis(HornifyModule &hm) : m_hm(hm) {}
+	    ~PredicateAbstractionAnalysis() {m_fp.reset (nullptr);}
+
+	    ufo::ZFixedPoint<ufo::EZ3>& getZFixedPoint () {return *m_fp;}
+
+	    void runAnalysis();
+		void guessCandidate(HornClauseDB &db);
+		ExprVector relToCand(Expr fdecl);
+		HornClauseDB generateAbstractDB(HornClauseDB &db, PredAbsHornModelConverter &converter);
```
agurfinkel: most of these methods should not be public.

include/seahorn/PredicateAbstraction.hh
```diff
@@ -0,0 +1,151 @@
...
+	    static char ID;
+
+	    PredicateAbstraction() : ModulePass(ID) {}
+	    virtual ~PredicateAbstraction() {}
+	    virtual bool runOnModule (Module &M);
+	    virtual void getAnalysisUsage (AnalysisUsage &AU) const;
+	    virtual const char* getPassName () const {return "PredicateAbstraction";}
+	};
+
+	class HornDbUtils
```
agurfinkel: seems generic enough to be factored out of pred abstraction

include/seahorn/PredicateAbstraction.hh
```diff
@@ -0,0 +1,151 @@
...
+	{
+	public:
+		static Expr applyArgsToBvars(Expr cand, Expr fapp, std::map<Expr, ExprVector> currentCandidates);
```
agurfinkel: why static class members? This is C++, we can use functions not wrapped in any class.

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,783 @@
...
+		HornDbModel oldModel;
+		oldModel.setCurrentCandidates(currentCandidates);
+
+		PredAbsHornModelConverter converter;
+
+		//run main algorithm
+		HornClauseDB new_db = generateAbstractDB(db, converter);
+
+		//initialize spacer based on new DB
+		m_fp.reset (new ZFixedPoint<EZ3> (m_hm.getZContext ()));
```
agurfinkel: I see how `fp` is initialized. This is enough for a quick experiment, but not sustainable in the long run.

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,783 @@
...
+		new_db.loadZFixedPoint (fp, false);
+		boost::tribool result = fp.query ();
+
+		if (result) outs () << "sat";
+		else if (!result)
+		{
+			outs() << "unsat";
+			HornDbModel absModel;
+			absModel.initModelFromFP(new_db, fp);
+			converter.setNewToOldPredMap(newToOldPredMap);
```
agurfinkel: I am not clear where `newToOldPredMap` is coming from. The client should only apply the converter. It should not be responsible for setting the state. This should be done inside `generateAbstractDB`.

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,783 @@
...
+			LOG("pabs-debug", outs() << "FINAL RESULT:\n";);
+			//Print invariants
+			printInvars(db, oldModel);
+		}
+		else outs () << "unknown";
+		outs () << "\n";
+		Stats::stop("Pabs solve");
+	}
+
+	void HornDbModel::addDef(Expr fapp, Expr def)
```
agurfinkel: `HornDbModel` and `HornModelConverter` should be factored out of predicate abstraction. They will be used all over the code even when predicate abstraction is not used.

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,783 @@
...
+	{
+		if(isOpX<TRUE>(def) || isOpX<FALSE>(def))
+		{
+			relToDefMap.insert(std::pair<Expr, Expr>(bind::fname(fapp), def));
+			return;
+		}
+
+		std::ostringstream oss;
+		oss << bind::fname(bind::fname(fapp));
```
agurfinkel: no idea what is happening here, but this does not follow our programming style. We never convert expressions to strings and search through them. That is why the expressions are structured!

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,783 @@
...
+		//For AbsModel
+		if(oss.str().find("_pabs") != -1)
```
agurfinkel: Why is predicate abstraction mixed with `HornModel`? I don't understand this code at all :(

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,783 @@
...
+				}
+			}
+			def = replace(def, boolVarsToBvarsMap);
+			relToDefMap.insert(std::pair<Expr, Expr>(bind::fname(fapp), def));
+		}
+		//For OrigModel
+		else
+		{
+			ExprVector vars;
+			HornDbUtils::get_all_integers(def, std::back_inserter(vars));
```
agurfinkel: No idea what is happening here. Why only integer variables? What about all other sorts? Arrays, reals, booleans?

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,783 @@
...
+			relToDefMap.insert(std::pair<Expr, Expr>(bind::fname(fapp), orig_def));
+		}
+	}
+
+	Expr HornDbModel::getDef(Expr fapp)
+	{
+		std::ostringstream oss;
+		oss << bind::fname(bind::fname(fapp));
+		//AbsModel
+		if(oss.str().find("_pabs") != -1)
```
agurfinkel: Same comment as before. Something is wrong since unrelated code is mixed together.

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,783 @@
...
+			Expr def = relToDefMap.find(rel)->second;
+			Expr def_app = HornDbUtils::applyArgsToBvars(def, fapp, currentCandidates);
+			return def_app;
+		}
+	}
+
+	void HornDbModel::initModelFromFP(HornClauseDB &db, ZFixedPoint<EZ3> &fp)
+	{
+		ExprVector all_preds_in_DB;
+		for(HornRule r : db.getRules())
```
agurfinkel: seems like a wrong thing to rely on the rules. There can be predicates without rules. 
This also might visit the same predicate multiple times if there are multiple rules. 
Should just iterate over predicates in db and create variable names on the fly.

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,783 @@
...
+		}
+		for(Expr pred : all_preds_in_DB)
+		{
+			LOG("pabs-debug", outs() << "REL: " << *(bind::fname(pred)) << "\n";);
+			Expr solution = fp.getCoverDelta(pred);
+			LOG("pabs-debug", outs() << "SOLUTION: " << *solution << "\n";);
+			addDef(pred, solution);
+		}
+	}
+
```
agurfinkel: Stopped here since the above really does not make sense. Will wait until another round.

agurfinkel:
@Chenguang-Zhu, is it ok to close this pull request? I think we merged everything that is in it already.

Chenguang-Zhu:
Yes, we can close it.

PR #43, opened 2016-08-24
Add predicate abstraction

Chenguang-Zhu:


include/seahorn/HornClauseDB.hh
```diff
@@ -356,6 +356,6 @@ namespace seahorn
   bool util_hasBvarInRule(HornRule r, HornClauseDB &db,
                           std::map<Expr, ExprVector> currentCandidates);
 
-
+  std::vector<std::string> util_split(std::string str,std::string pattern);
```
agurfinkel: please use `boost::tokenizer` instead: http://stackoverflow.com/questions/7941725/boosttokenizer-comma-separated-c

include/seahorn/PredicateAbstraction.hh
```diff
@@ -0,0 +1,83 @@
...
+		std::map<Expr, Expr> m_newToOldPredMap;
+		HornClauseDB* m_abs_db;
+
+		std::map<Expr, ExprMap>& getRelToBoolToTermMap() {return m_relToBoolToTermMap;}
+	public:
+		PredAbsHornModelConverter() {}
+		virtual ~PredAbsHornModelConverter() {}
+		bool convert (HornDbModel &in, HornDbModel &out);
+
+		void addRelToBoolToTerm(Expr rel, ExprMap &boolToTermMap) {m_relToBoolToTermMap.insert(std::pair<Expr, ExprMap>(rel, boolToTermMap));}
```
agurfinkel: std::make_pair

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,582 @@
...
+		generateAbstractRules(db, new_DB, converter);
+
+		generateAbstractQueries(db, new_DB);
+
+		LOG("pabs-debug", outs() << "NEW DB: \n";);
+		LOG("pabs-debug", outs() << new_DB << "\n";);
+
+		converter.setAbsDB(new_DB); //set converter
+
+		return new_DB;
```
agurfinkel: Returning objects by value can be very expensive. In C++, it is more common for a function to take an output parameter. `new_DB` should be passed by reference and populated by this function.

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,582 @@
...
+	{
+		//For each relation, generate its abstract version
+		for(Expr rel : db.getRelations())
+		{
+			LOG("pabs-debug", outs() << "OLD REL: " << *rel << "\n";);
+			ExprVector new_args;
+			//Push fdecl name
+			Expr old_fdecl_name = bind::fname(rel);
+			std::ostringstream oss;
+			oss << old_fdecl_name << "_pabs";
```
agurfinkel: do not convert expressions to strings. You probably wanted to use `tag` expressions here.

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,582 @@
...
+				{
+					new_args.push_back(bind::domainTy(rel, i));
+				}
+			}
+			//Push return type
+			new_args.push_back(bind::rangeTy(rel));
+			Expr new_rel = mknary<FDECL>(new_args);
+
+			//new pred name
+			Expr new_fdecl_name = mkTerm<std::string>(oss.str(), new_rel->efac());
```
agurfinkel: do not use strings to name predicates.

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,582 @@
...
+			new_rel = bind::rename(new_rel, new_fdecl_name);
+
+			LOG("pabs-debug", outs() << "NEW REL: " << *new_rel << "\n";);
+			new_DB.registerRelation(new_rel);
+
+			m_oldToNewPredMap.insert(std::pair<Expr, Expr>(rel, new_rel));
```
agurfinkel: if you are not using strings, there is no need for these maps.

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,582 @@
...
+		{
+			Expr old_fdecl = bind::fname(old_query);
+			Expr new_fdecl = m_oldToNewPredMap.find(old_fdecl)->second;
+			ExprVector old_args;
+			Expr new_query = bind::fapp(new_fdecl, old_args);
+			new_DB.addQuery(new_query);
+		}
+	}
+
+	void PredicateAbstractionAnalysis::guessCandidate(HornClauseDB &db)
```
agurfinkel: should not guessing of terms be factored out and shared between PredAbs and Houdini?

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,582 @@
...
+				lemmas.push_back(mk<GEQ>(bvars[i], zero));
+				lemmas.push_back(mk<GEQ>(bvars[i], two));
+				lemmas.push_back(mk<LEQ>(bvars[i], two));
+				parseLemmasFromExpFile(bvars[i], lemmas, filepath);
+			}
+		}
+		return lemmas;
+	}
+
+	void PredicateAbstractionAnalysis::parseLemmasFromExpFile(Expr bvar, ExprVector& lemmas, std::string filepath)
```
agurfinkel: Is it possible to store lemmas in some smt-lib compatible representation and use smt-lib parser instead of a custom one?

lib/seahorn/PredicateAbstraction.cc
Chenguang-Zhu: We can use tag variables to create predicates names like P!pabs. But in this situation the map seems still useful. Because when given an abstract predicate P!pabs, although we can use tag expression to easily get the name of concrete predicate P, we still need to know the arguments list to construct the concrete predicate. Without this map, we must go over all predicates in DB and find the one that has the name P. I think maybe it's reasonable to keep these maps?

agurfinkel: agreed.

you can also tag a decl instead of a name. Note that name can be an
arbitrary expression. There are no restrictions.

cheers
arie

include/seahorn/GuessCandidates.hh
```diff
@@ -0,0 +1,22 @@
...
+#include <fstream>
+#include <iostream>
+#include <boost/tokenizer.hpp>
+
+namespace seahorn
+{
+	//Simple templates
+	ExprVector relToCand(Expr pred);
+	//Load templates from file
+	ExprVector applyTemplatesFromExperimentFile(Expr fdecl, std::string filepath);
```
agurfinkel: pass const std::string by reference (everywhere)

include/seahorn/HornClauseDB.hh
```diff
@@ -302,60 +302,56 @@ namespace seahorn
...
-  {filter (e, Util_IsInteger(), out);}
+  {filter (e, IsInteger(), out);}
 
   template<typename OutputIterator>
   void get_all_booleans(Expr e, OutputIterator out)
-  {filter (e, Util_IsBoolean(), out);}
+  {filter (e, IsBoolean(), out);}
 
-  bool util_hasBvarInRule(HornRule r, HornClauseDB &db,
+  bool hasBvarInRule(HornRule r, HornClauseDB &db,
```
agurfinkel: comment

lib/seahorn/HornDbModel.cc
```diff
@@ -11,29 +11,35 @@ namespace seahorn
...
+		lemma_def = replace(lemma, actual_arg_to_bvar_map);
     }
 
-    Expr lemma_def = replace(lemma, actual_arg_to_bvar_map);
-
-    m_defs.insert(std::pair<Expr, Expr>(bind::fname(fapp), lemma_def));
+    //Key exists before
+    if(m_defs.find(bind::fname(fapp)) != m_defs.end())
+    {
+    	m_defs[bind::fname(fapp)] = lemma_def;
```
agurfinkel: You can always do that. There is no need to use `insert` method.

lib/seahorn/PredicateAbstraction.cc
```diff
@@ -0,0 +1,480 @@
...
+	{
+		//For each relation, generate its abstract version
+		for(Expr rel : db.getRelations())
+		{
+			LOG("pabs-debug", outs() << "OLD REL: " << *rel << "\n";);
+			ExprVector new_args;
+
+			Expr old_fdecl_name = bind::fname(rel);
+			//new pred name
+			Expr new_fdecl_name = variant::tag(old_fdecl_name, mkTerm<std::string>("pabs", old_fdecl_name->efac()));
```
agurfinkel: Use the variant of `tag` that takes a string instead of creating string expression using `mkTerm`

PR #44, opened 2016-08-24
Add regression tests

Chenguang-Zhu:


agurfinkel:
@Chenguang-Zhu , this pull request is failing integration build. Please take a look.

Chenguang-Zhu:
Hi Arie,

It failed because it needs command option --horn-pred-abs to build the
regression tests, but at that time the pred-abs is on a different pull
request. Now it can work well. We can merge it.

Cheers,
Chenguang

agurfinkel:
Can you please re-submit it so that we can see the check pass. I don't see a way to re-do the check in other ways.

PR #45, opened 2016-08-25
assign unique names to symbolic inputs

jiaqi-liu:


lib/Transforms/Instrumentation/KleeInternalize.cc
```diff
@@ -239,13 +243,13 @@ namespace
 
     bool runOnModule (Module &M)
     {
+      m_module = &M;
       declareKleeFunctions(M);
       internalizeVariables(M);
 
       for (Function &F : M)
       {
-        if (shouldInternalize (F)) defineFunction (F);
```
agurfinkel: why did you remove this line?

agurfinkel: Is that because now the function is defined at a call site, so that each call site has a different definition?

lib/Transforms/Instrumentation/KleeInternalize.cc
```diff
@@ -270,7 +274,22 @@ namespace
           CallInst *ninst = nullptr;
           Builder.SetInsertPoint (&inst);
           
-          if (fn->getName().equals ("verifier.assume"))
+          if (shouldInternalize(*fn))
+          {
+            int id = m_functionId[fn->getName()];
+            Function * newFn = cast<Function> (m_module->getOrInsertFunction(
+              fn->getName().str() + "_" +std::to_string(id),
```
agurfinkel: use a period instead of underscore to prevent clashes with user defined functions.

jiaqi-liu: Good point. Changed.

lib/Transforms/Instrumentation/KleeInternalize.cc
jiaqi-liu: Yes. We need to assign a unique name per call. I added a few lines of code to remove the initial declaration in the new commit.

PR #51, opened 2016-08-31
Sudo-code for implementing ICE based on our Houdini

Chenguang-Zhu:


test/sudo-code
```diff
@@ -0,0 +1,88 @@
...
+      {
+        print 'DO NOT SUPPORT THIS TYPE!'
+      }
+    }
+    write lowerInterval and upperInterval into C5.intervals file
+    lowerInterval = upperInterval
+    upperInterval = lowerInterval
+  }  
+}
+
```
agurfinkel: Assuming your interpretation of the intervals is correct, this is good so far.

test/sudo-code
```diff
@@ -0,0 +1,88 @@
...
+    {
+      get model m from solver
+      extract attributes value from m             //?
+      extract implication pairs like (dp1 -> dp2) //?
+      add attributes as counter examples into C5.data file
+      C5learn()
+      add dependent rules back to worklist, including itself
+    }
+  }
+}
```
agurfinkel: I need a concrete example to illustrate how to implement your questions. This is really easy. For each attribute, like a variable `x`, you find its name in the pre-state and use `model.eval` to find its value. Then you find its name in the post-state, and use `model.eval` again. The result is a pair, value of `x` before the rule and value of `x` after the rule. Do it for all variables and you have an implication pair.

This is easiest in flat encoding since every variable has both source and destination versions.

Note that to extend this with E- and C-examples you'll also need to form the horn system we described on skype (with all the rules at once) and solve it with some bound. If the solution is sat, you'll mine the samples from there.

test/sudo-code
```diff
@@ -0,0 +1,88 @@
...
+
+C5learn()
+{
+  generateC5DataAndImplfiles()
+  call C5 terminal command
+  parse the result tree from C5.json file
+  convert the result tree to invariant
```
agurfinkel: note that there is `boost::property_tree::read_json` that will help with parsing json in C++.

test/sudo-code
caballa: Are we assuming a particular encoding? I agree that with flat encoding is
trivial but what about LBE? In that case,I was assuming that we only do it
for the cutpoints. Is that right?

agurfinkel: Yes, `flarge`. Sorry, I did not even think about the other option :)

lememta:
Can you post how to run the ICE framework on an example? and the output that you get?

Chenguang-Zhu:
Okay. I'm still working on the implementation. I'll post it as soon as I
finish.

PR #55, opened 2016-09-11
for compilation on Debian-7 and 32-bit systems

monniaux:


agurfinkel:
@monniaux , thank you!

PR #75, opened 2017-09-13
Make sassert not hijack else statements

kuhar:
This improves the sassert macro by making it an expression and not an if statement.

Consider this snippet:
``` c
if (cond1)
  sassert(cond2);
else
  foo();
```
The problem with sassert being defined as if statement is that it hijacks else statements that follow it and changes user's intended control flow. One of the solutions would be to put the if inside a block `#define sassert(X) { if ... }`, but that makes it a compound statement and it would make it impossible to use with an else following it. The other idiomatic C solution would be to put it inside `do { if ... } while(0)`, but the problem is that the C frontend could generate a new loop when compiling with `-O0`.

This path comes with a regression test for using sassert within expressions and with else block.

### DIFF
--- a/include/seahorn/seahorn.h
+++ b/include/seahorn/seahorn.h
@@ -10,5 +10,5 @@ extern void __VERIFIER_assume (int);
 #endif
 
 #define assume __VERIFIER_assume
-#define sassert(X) if(!(X)) __VERIFIER_error ()
+#define sassert(X) (void)((X) || (__VERIFIER_error (), 0))
 #endif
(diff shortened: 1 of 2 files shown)


agurfinkel:
fantastic! I'll merge once it passes ci

caballa:
The only thing worries me is that the if+assertion is converted into a or+assertion which might make things more complicated to crab-llvm but not really sure.

agurfinkel:
@caballa , the `or` should short-circuit, i.e., it will be converted to a jump since the second argument has side-effects.

PR #76, opened 2017-09-17
use MBP based QE tactic

rainoftime:
Z3 4.5 has a new tactic[1] for quantifier elimination called "qe2", which outperforms 
the "qe" tactic.  "qe2" is built on top of a recent technique "model based projection"[2].

[1] Playing with Quantified Satisfaction, N Bjørner et al, LPAR 15
[2] SMT-based model checking for recursive programs, CAV 14


agurfinkel:
SeaHorn is compiled against Spacer, a custom fork of Z3 that includes an IC3/PDR engine. This fork does not include the same MBP code. Furthermore, quantifier elimination tactic is not used by SeaHorn. So this change would be irrelevant.

PR #89, opened 2017-10-25
Add dockerfiles for llvm 3.8 and z3

kuhar:
Second attempt.

This time, both dockerfiles are parameterized by ubuntu version (trusty / xenial) and build type (Debug / Release).

With z3, the problem is that the python build script uses `-d` for debug, instead of taking a string like cmake (`Debug`, `Release`, etc.). I also don't know how to use make to automatically package z3.

Any suggestions?


deps/llvm38/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+RUN git clone https://github.com/llvm-mirror/llvm/ ./ -b release_38 --depth=1
+WORKDIR /llvm/build
+
+# Build configuration.
+RUN cmake -GNinja \
+          -DCMAKE_BUILD_TYPE=$BUILD_TYPE \ 
+          -DLLVM_TARGETS_TO_BUILD="X86;" \
+          -DCMAKE_INSTALL_PREFIX:PATH=/llvm/out \
+          -DLLVM_ENABLE_PEDANTIC=OFF \
+  		  -DLLVM_ENABLE_PIC=ON \
```
kuhar: nit: fix whitespace here

deps/z3/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+RUN echo Build type set to: $BUILD_TYPE
+
+WORKDIR /tmp/dockerutils
+
+# Create a helper script that works as switch (VAL) { Key0 : Val0, ...}.
+# This is to work around docker limitations and pass right correct flag to the
+# python configuration script.
+
+RUN echo '#!/bin/sh' > switch.sh && \ 
+    echo 'VAL=$1;shift;while test $# -gt 0;do if [ "$1" = "$VAL" ];then echo $2;exit 0;fi;shift;shift;done' >> switch.sh && \
```
kuhar: over-complicated?

PR #99, opened 2017-11-13
Add a dockerfile for seahorn

kuhar:
This works for trusty/xenial in release configuration right now. Debug configuration will not work until I produce tarballs and upload them.

### DIFF
--- a/deps/BuildingSeaHorn.md
+++ b/deps/BuildingSeaHorn.md
@@ -1,0 +1,19 @@
+# Building SeaHorn with Docker
+
+To build SeaHorn with Docker:
+```shell
+cd seahorn/deps
+docker build --build-arg UBUNTU=xenial --build-arg BUILD_TYPE=Release -t seahorn_xenial_rel .
+docker run -it z3_xenial_rel
+```
+This will automatically download all dependencies and build SeaHorn under /seahorn/build.
+
+SeaHorn's install directory is added to PATH, and a tarball with it is ready under /seahorn/build.
+
+The possible build arguments are:
+- UBUNTU: trusty, xenial
+- BUILD_TYPE: Release, Debug
+
+NOTE: Debug build type is not supported at the moment.
+
+Note that both `UBUNTU` and `BUILD_TYPE` are required arguments.
--- a/deps/BuildingSeaHorn.md
+++ b/deps/BuildingSeaHorn.md
@@ -4,11 +4,11 @@ To build SeaHorn with Docker:
 ```shell
 cd seahorn/deps
 docker build --build-arg UBUNTU=xenial --build-arg BUILD_TYPE=Release -t seahorn_xenial_rel .
-docker run -it z3_xenial_rel
+docker run -v `pwd`:/host -it sehorn_xenial_rel
 ```
 This will automatically download all dependencies and build SeaHorn under /seahorn/build.
 
-SeaHorn's install directory is added to PATH, and a tarball with it is ready under /seahorn/build.
+SeaHorn's install directory is added to PATH, and a tarball with it is ready under the current working directory (on the host).
 
 The possible build arguments are:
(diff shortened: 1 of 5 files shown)


deps/BuildingSeaHorn.md
```diff
@@ -0,0 +1,19 @@
+# Building SeaHorn with Docker
+
+To build SeaHorn with Docker:
+```shell
+cd seahorn/deps
+docker build --build-arg UBUNTU=xenial --build-arg BUILD_TYPE=Release -t seahorn_xenial_rel .
+docker run -it z3_xenial_rel
```
agurfinkel: should this be `z3_xenial_rel`?

agurfinkel: is this a standard naming convention for docker tags?

deps/Dockerfile
```diff
@@ -0,0 +1,107 @@
...
+    export UB=$(lsb_release --a 2>&1 | cut -f2 | tail -n 1) && \
+    echo "$UB"_"$BT"_ > /tmp/dockerutils/prefix.txt && \
+    cat /tmp/dockerutils/prefix.txt
+
+RUN mkdir -p /deps
+WORKDIR /deps
+RUN export PREFIX=$(cat /tmp/dockerutils/prefix.txt) && \
+    export BOOST_LINK=$(echo https://github.com/kuhar/seahorn_deps/raw/master/"$PREFIX"boost162.tar) && \
+    wget $BOOST_LINK
+
```
agurfinkel: why are deps not zipped? (same for z3)

deps/Dockerfile
```diff
@@ -0,0 +1,107 @@
...
+    tar -xvf $Z3_TAR && rm $Z3_TAR && \
+    tar -xvf $LLVM_TAR && rm $LLVM_TAR
+
+RUN ls -al --block-size=M 1>&2
+
+RUN mkdir -p /seahorn
+WORKDIR /seahorn
+
+# Checkout SeaHorn.
+RUN git clone https://github.com/seahorn/seahorn ./ --depth=10
```
agurfinkel: tag/branch should be a parameter set to `master` by default to allow building different versions

agurfinkel: need to figure out how to make this work with travis-ci builds. In travis, an appropriate branch of seahorn will already be in the working directory and should be `ADD` to docker container

deps/Dockerfile
```diff
@@ -0,0 +1,107 @@
...
+          -DCMAKE_CXX_COMPILER=g++-5 \
+          -DCPACK_GENERATOR="TGZ" \
+          -DCMAKE_EXPORT_COMPILE_COMMANDS=1 \
+          ../
+
+RUN cmake --build . --target extra  && cmake ..
+RUN cmake --build . --target crab  && cmake ..
+RUN cmake --build . --target install
+RUN cmake --build . --target package
+
```
agurfinkel: would be good to have a way to expose the package outside of docker as well

deps/Dockerfile
```diff
@@ -0,0 +1,107 @@
...
+ENV PATH "/seahorn/build/run/bin:$PATH"
+
+WORKDIR /seahorn
+# Define default command.
+CMD ["bash"]
```
agurfinkel: We also need a docker container for seahorn itself. That is, something that includes seahorn and its runtime dependencies but not the build trail. This would be useful for running tests after the build and for users to try it out.

deps/Dockerfile
kuhar: I changed it to copy the seahorn tarball to host and run shell after that.

kuhar:
Tried to address the comments.
GitHub releases seem to work well for this usecase.

When it comes to a CI version, I'll look into it later, as I'm not sure how to use docker for that and pass all the pieces of information we need to build the correct revision and run the tests.

PR #103, opened 2017-12-20
[WIP] Simple Memory Checks

kuhar:
Work in progress


kuhar:
Replaced plain dbgs() with LOG.

Also added instrumentation for GlobalVariables (both when determined potentially safe and unsafe). This appears to work as expected for when a GV is considered uninteresting (== safe), but not when we start tracking it (as in `test/smc/list1_global_unsat3.c`). For some reason, the BAR and FOO tags don't appear in shadow mem, things don't alias when they get to the searhon solver, and sea-dsa reports at most 1 alloc site for each pointer, and it doesn't want to produce a mem-dot file...

I will try to debug it some more.

lib/Transforms/Scalar/LowerGvInitializers.cc
```diff
@@ -1,240 +1,242 @@
 #include "seahorn/Transforms/Scalar/LowerGvInitializers.hh"
 
-#include "boost/range.hpp"
 #include "boost/format.hpp"
+#include "boost/range.hpp"
```
kuhar: I should have probably split formatting changes and the fix to lowering. Is is better to leave it as-is, or make into 2 separate commits (the formatting one in a separate pull request)?

agurfinkel: 2 separate commits so that it is clear what is your change and what is formatting would be nice

kuhar:
Closing for now. I created a new PR with fewer commits and a better description.

agurfinkel:
@kuhar , the PR is not actually closed. Should I close it?

kuhar:
Yes, seems like I cannot close it with the current permissions

PR #105, opened 2017-12-26
Add initial SimpleMemoryCheck pass implementation

kuhar:
The patch adds initial implementation of the Simple Memory Safety Checks.

For each memory instruction (load or store), it gets tries to go up the def-use chain and track the pointer's origin and offset. It goes up and looks through GEPs and bitcasts, until it reaches a _barrier_ (load, function call, phi node, select instruction, etc.).
The assumption is that the barrier produces a zero-offset pointer from the allocation site.

Next, it uses Sea-DSA to gather or possible allocation sites (malloc, alloca, global variables), and divides them into two groups: ones that can cause unsafe memory access after applying the accumulated offset, and those that are known to result in a safe memory access.

The pass performs the analysis described above and emits instrumentation for each pair of <allocation site, memory instruction>, which allows us to detect whether or not it is possible to go from the allocation site to the memory instruction.

For now, the pass doesn't use any other alias analysis, and is limited to analyzing only the first 16 memory instructions, and instrumenting only the first pair <AS, MI>.

### DIFF
--- a/include/seahorn/Transforms/Instrumentation/SimpleMemoryCheck.hh
+++ b/include/seahorn/Transforms/Instrumentation/SimpleMemoryCheck.hh
@@ -1,0 +1,12 @@
+#ifndef SIMPLE_MEMORY_CHECK__HH
+#define SIMPLE_MEMORY_CHECK__HH
+
+namespace llvm {
+class ModulePass;
+}
+
+namespace seahorn {
+llvm::ModulePass *CreateSimpleMemoryCheckPass();
+}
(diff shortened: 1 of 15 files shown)


agurfinkel:
Looks really good. Tiny comments in a few places, but we are good to merge.

kuhar:
This should address everything except the suggestion to expose special functions with a separate module pass.

I put the global variable lowering changes in a separate PR. Now this PR depends on it.

PR #113, opened 2018-01-11
Add PromoteMemcpy pass

kuhar:
... and fix how SMC is scheduled

### DIFF
--- a/lib/Transforms/Scalar/PromoteMemcpy.cc
+++ b/lib/Transforms/Scalar/PromoteMemcpy.cc
@@ -1,0 +1,195 @@
+#include "llvm/Analysis/AssumptionCache.h"
+#include "llvm/Analysis/MemoryBuiltins.h"
+#include "llvm/IR/CallSite.h"
+#include "llvm/IR/DataLayout.h"
+#include "llvm/IR/Dominators.h"
+#include "llvm/IR/IRBuilder.h"
+#include "llvm/Pass.h"
+#include "llvm/Support/CommandLine.h"
+#include "llvm/Support/Debug.h"
+#include "llvm/Support/raw_ostream.h"
+#include "llvm/Transforms/Utils/BuildLibCalls.h"
+#include "llvm/Transforms/Utils/Local.h"
+#include "llvm/Transforms/Utils/SimplifyLibCalls.h"
+
+#include "avy/AvyDebug.h"
+
+#define PMCPY_LOG(...) LOG("promote-memcpy", __VA_ARGS__)
+
+using namespace llvm;
+
+namespace {
+class PromoteMemcpy : public FunctionPass {
+public:
+  static char ID;
+
+  PromoteMemcpy() : FunctionPass(ID) {}
+
+  bool runOnFunction(Function &F);
+
+  void getAnalysisUsage(AnalysisUsage &AU) const override {
+    AU.setPreservesCFG();
+    AU.addRequired<llvm::DominatorTreeWrapperPass>();
+    AU.addRequired<AssumptionCacheTracker>();
+  }
+
+  const char *getPassName() const override { return "PromoteMemcpy"; }
+
+private:
+  Module *m_M = nullptr;
+  LLVMContext *m_Ctx = nullptr;
+  const DataLayout *m_DL = nullptr;
+  DominatorTree *m_DT = nullptr;
+  AssumptionCache *m_AC = nullptr;
+
+  bool simplifyMemCpy(MemCpyInst *MCpy);
+};
+
+char PromoteMemcpy::ID = 0;
+
+bool PromoteMemcpy::simplifyMemCpy(MemCpyInst *MI) {
+  assert(MI);
+
+  unsigned DstAlign = getKnownAlignment(MI->getDest(), *m_DL, MI, m_AC, m_DT);
+  unsigned SrcAlign = getKnownAlignment(MI->getSource(), *m_DL, MI, m_AC, m_DT);
+  unsigned MinAlign = std::min(DstAlign, SrcAlign);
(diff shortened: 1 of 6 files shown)


lib/Transforms/Scalar/PromoteMemcpy.cc
```diff
@@ -0,0 +1,211 @@
...
+  unsigned DstAlign =
+      getKnownAlignment(MI->getArgOperand(0), *m_DL, MI, m_AC, m_DT);
+  unsigned SrcAlign =
+      getKnownAlignment(MI->getArgOperand(1), *m_DL, MI, m_AC, m_DT);
+  unsigned MinAlign = std::min(DstAlign, SrcAlign);
+  unsigned CopyAlign = MI->getAlignment();
+
+  if (CopyAlign < MinAlign) {
+    MI->setAlignment(ConstantInt::get(MI->getAlignmentType(), MinAlign, false));
+    return MI;
```
agurfinkel: shouldn't this be `return true`

lib/Transforms/Scalar/PromoteMemcpy.cc
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  // If MemCpyInst length is 1/2/4/8 bytes then replace memcpy with
+  // load/store.
+  ConstantInt *MemOpLength = dyn_cast<ConstantInt>(MI->getArgOperand(2));
```
agurfinkel: check that `MemCpyInst` does not provide better API to get at the arguments

lib/Transforms/Scalar/PromoteMemcpy.cc
```diff
@@ -0,0 +1,211 @@
...
+  assert(DstPtrTy);
+
+  if (SrcPtrTy != DstPtrTy)
+    return false;
+
+  auto *Buffer = dyn_cast<AllocaInst>(Dst);
+  if (!Buffer)
+    return false;
+
+  PMCPY_LOG(dbgs() << "\nSrc:\t");
```
agurfinkel: it is cleaner and easier to read when wrapped in the single `LOG`. I would do that as ```LOG("pcmpy", all-the-code;);```

agurfinkel: Is `dbgs()` available for release builds? If not, I'd use `errs()`

lib/Transforms/Scalar/PromoteMemcpy.cc
```diff
@@ -0,0 +1,211 @@
...
+  auto *SrcPtrTy = dyn_cast<PointerType>(Src->getType());
+  assert(SrcPtrTy);
+
+  auto *DstPtrTy = dyn_cast<PointerType>(Dst->getType());
+  assert(DstPtrTy);
+
+  if (SrcPtrTy != DstPtrTy)
+    return false;
+
+  auto *Buffer = dyn_cast<AllocaInst>(Dst);
```
agurfinkel: Why does it matter that `memcpy` is into `alloca`? We want to minimize the number of memory operations.

lib/Transforms/Scalar/PromoteMemcpy.cc
```diff
@@ -0,0 +1,211 @@
...
+  auto *BufferTy = dyn_cast<StructType>(SrcPtrTy->getPointerElementType());
+  if (!BufferTy)
+    return false;
+
+  IRBuilder<> Builder(MI);
+  auto *I64Ty = IntegerType::getInt64Ty(*m_Ctx);
+  auto *NullInt = Constant::getNullValue(I64Ty);
+  auto *I32Ty = IntegerType::getInt32Ty(*m_Ctx);
+
+  for (unsigned i = 0, e = BufferTy->getStructNumElements(); i != e; ++i) {
```
agurfinkel: This will not recursively copy nested structs, right?
It is fine to cut corners in that way, but would be good to add a warning so that we know this is happening.

lib/Transforms/Scalar/PromoteMemcpy.cc
```diff
@@ -0,0 +1,211 @@
...
+    auto *Idx = Constant::getIntegerValue(I32Ty, APInt(32, i));
```
agurfinkel: add a comment on what code is generated so it is easier to follow

lib/Transforms/Scalar/PromoteMemcpy.cc
```diff
@@ -0,0 +1,211 @@
...
+
+  PMCPY_LOG(dbgs() << "Removing dead memcpys in " << F.getName() << ":");
+  for (auto *MCpy : ToDeleteQueue) {
+    PMCPY_LOG(MCpy->print(dbgs() << "\n\t"));
+
+    auto *SrcPtr = dyn_cast<BitCastInst>(MCpy->getArgOperand(1));
+    auto *DstPtr = dyn_cast<BitCastInst>(MCpy->getArgOperand(0));
+
+    MCpy->eraseFromParent();
+    if (SrcPtr && SrcPtr->getNumUses() == 0) {
```
agurfinkel: Use `hasNUses(0)` instead of `getNumUses` :)

lib/Transforms/Scalar/PromoteMemcpy.cc
kuhar: Is a recursive copy really what we want to do here?

agurfinkel: if there are nested structs in llvm then yes

kuhar:
Addressed CR comments. Will add recursive memcpy promotion later, if we need it.

kuhar:
This version can look through pointer casts, which lets get rid of all ptrtoint and inttoptr instructions in CASS.

PR #115, opened 2018-01-19
Filter out vtables and alloc-like functions

kuhar:


lib/Transforms/Scalar/PromoteMemcpy.cc
```diff
@@ -114,21 +114,59 @@ bool PromoteMemcpy::simplifyMemCpy(MemCpyInst *MI) {
   // for each field_id in fields(BufferTy):
   //   *GEP(Dst, field_id) = *GEP(Src, field_id)
   //
-  for (unsigned i = 0, e = BufferTy->getStructNumElements(); i != e; ++i) {
-    auto *Idx = Constant::getIntegerValue(I32Ty, APInt(32, i));
 
-    auto *SrcGEP = Builder.CreateInBoundsGEP(nullptr, SrcPtr, {NullInt, Idx},
-                                             "src.gep.pmcpy");
-    auto *NewLoad = Builder.CreateLoad(SrcGEP, SrcPtr->getName() + ".pmcpy");
+  using Transfer = std::pair<Value *, Value *>;
```
kuhar: This new recursive copy doesn't seem to ever fire. No sure whether to keep it or to revert to the old one and emit a warning when we find a memcpy that requires it.

agurfinkel: You can add a warning that untested code is running if you want.
Or at least comment in the code that it never ran as far as we know.

Issue #117, opened 2018-02-02
Minimal docker image consuming searhorn tarball


kuhar:
This can be considered an intermediate step between #116 and #93.

We would like to have a clean, minimal docker image that consumes a prebuilt seahorn and runs all the tests. We don't want to install any build-related packages for this docker image, just grab seahorn and clang (from llvm.org), and run the lit tests.
The goal it to have a more light-weight docker container for running seahorn and running the test suite outside of the build environment. This way we can make sure there aren't any dependencies we are currently not aware of.
In #116, we'll be able use this docker file to create a docker-based seahorn distribution and put it on dockerhub.

PR #118, opened 2018-02-08
Add tests to build for CI

mrthefakeperson:
- includes tests which are currently being run in install
 - necessary for #117 (Dockerfile will fetch SeaHorn, build package, move to a separate container and run tests)

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -58,3 +58,9 @@ add_lit_testsuite(test-dsa "Regression test for DSA"
   test_dir=${CMAKE_CURRENT_BINARY_DIR}
   DEPENDS seahorn
   )
+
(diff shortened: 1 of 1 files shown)


test/CMakeLists.txt
```diff
@@ -58,3 +58,9 @@ add_lit_testsuite(test-dsa "Regression test for DSA"
   test_dir=${CMAKE_CURRENT_BINARY_DIR}
   DEPENDS seahorn
   )
+
+install (DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/simple DESTINATION test)
```
agurfinkel: please change destination to `share/searhon/test` to be consistent with typical location of files on a linux system

PR #122, opened 2018-02-17
Enable docker builds

mrthefakeperson:
- CI with Travis/Docker
 - download clang from llvm.org instead of using apt-get
 - create/push Docker images for seahorn and seahorn-build on Dockerhub (inside docker folder)


agurfinkel:
The build failed with 10m wait. I can't merge it until we resolve that problem.

.travis.yml
```diff
@@ -5,59 +5,26 @@ env:
...
-  #- /usr/bin/cmake --build .
-  - /usr/bin/cmake --build . --target extra && /usr/bin/cmake ..
-  - /usr/bin/cmake --build . --target crab && /usr/bin/cmake ..
-  - /usr/bin/cmake --build . --target install
-  - /usr/bin/cmake --build . --target test-simple
-  - /usr/bin/cmake --build . --target test-solve
-  - /usr/bin/cmake --build . --target test-abc   
-  - /usr/bin/cmake --build . --target test-dsa
+  ### build with docker
+  - docker build --build-arg UBUNTU=trusty --build-arg BUILD_TYPE=MinSizeRel --build-arg TRAVIS=true -t seahorn_trusty_build -f docker/seahorn-full-size-rel.Dockerfile .
```
agurfinkel: why are there 2 `docker build` commands? I was expecting only one

agurfinkel: is the first command for the build and the second to create a seahorn distribution to run the test?
If so, please comment this in the file to avoid similar confusion in the future.

docker/seahorn-build.Dockerfile
```diff
@@ -0,0 +1,85 @@
+#
+# Dockerfile for SeaHorn
```
agurfinkel: the comment on this dockerfile seems wrong.
It does not compile seahorn. 

Perhaps this is the base for seahorn-build image. In which case it is not clear why it is cloning seahorn.
A build image must include all dependencies except for seahorn itself.

docker/seahorn-build.Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+ARG UBUNTU
+
+# Pull base image.
+FROM buildpack-deps:$UBUNTU
+
+ARG BUILD_TYPE
+RUN echo "Build type set to: $BUILD_TYPE"
+
+# Install deps.
+RUN \
```
agurfinkel: minimize number of `RUN` commands that allocate disk space. All install/download commands must under the same `RUN`

docker/seahorn-build.Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+    tar -xvf llvm38.tar.gz && \
+    rm *.tar.gz
+
+RUN ls -al --block-size=M 1>&2
+
+RUN mkdir -p /seahorn
+WORKDIR /seahorn
+
+# clone SeaHorn and set up clang
+RUN git clone https://github.com/seahorn/seahorn ./ -b master --depth=10 && \
```
agurfinkel: why is seahorn cloned here?

docker/seahorn-full-size-rel.Dockerfile
```diff
@@ -0,0 +1,53 @@
+#
+# Dockerfile for SeaHorn
+# This produces package in /seahorn/build
```
agurfinkel: update comment. Seems like the same comment as in the previous file

docker/seahorn-build.Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+    > /tmp/dockerutils/dt_out.txt
+RUN export BT=$(cat /tmp/dockerutils/dt_out.txt) && \
+    export UB=$(lsb_release --a 2>&1 | cut -f2 | tail -n 1) && \
+    echo "$UB"_"$BT" > /tmp/dockerutils/prefix.txt && \
+    cat /tmp/dockerutils/prefix.txt
+
+RUN mkdir -p /deps
+WORKDIR /deps
+RUN export PREFIX=$(cat /tmp/dockerutils/prefix.txt) && \
+    export DEPS_LINK=$(echo https://github.com/kuhar/seahorn_deps/releases/download/v0.1/"$PREFIX".tar.gz) && \
```
agurfinkel: use https://github.com/seahorn/seahorn-ext-deps instead of *kuhar*

docker/seahorn-full-size-rel.Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+RUN if [ "$TRAVIS" = "" ] ; \
+      then git fetch --depth=10 && git reset --hard origin/master ; \
+    fi
+RUN mkdir -p /seahorn/build
+WORKDIR /seahorn/build
+
+ARG BUILD_TYPE
+
+# Build configuration.
+RUN cmake -GNinja \
```
agurfinkel: Use a single `RUN` command to avoid creating intermediate snapshots

docker/seahorn-full-size-rel.Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+RUN cmake --build . --target crab  && cmake ..
+RUN cmake --build . --target install
+RUN cmake --build . --target package
+
+ENV PATH "/seahorn/build/run/bin:$PATH"
+
+WORKDIR /seahorn
+
+RUN echo '#!/bin/sh' > /tmp/cpy.sh && \ 
+    echo 'cp /seahorn/build/*.tar.gz /host/ && bash' >> /tmp/cpy.sh && \
```
agurfinkel: what is the benefit of `bash` at the end of `cpy`?

Rename `cpy.sh` to be more descriptive (useful when looking at container history)

docker/seahorn.Dockerfile
```diff
@@ -0,0 +1,42 @@
+#
+# Minimal Dockerfile for SeaHorn
+# produces a lightweight container with SeaHorn
+# Arguments:
+#  - UBUNTU:     trusty, xenial
+#
+
+ARG UBUNTU
+FROM buildpack-deps:$UBUNTU
```
agurfinkel: should depend on a seahorn container that already includes the minimal dependencies, such as clang.
Ideally, this container has everything that is needed to just unpack seahorn binary and go

docker/seahorn.Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+    apt-get update && apt-get install -yqq software-properties-common && \
+    add-apt-repository --yes ppa:ubuntu-toolchain-r/test && \
+    apt-get update && apt-get upgrade -yqq && \
+    apt-get install -yqq python-dev python-pip libgraphviz-dev libstdc++6 && \
+    pip install lit OutputCheck && \
+    easy_install networkx pygraphviz && \
+# download clang
+    cd run/bin && \
+    if [ "$UBUNTU" = "xenial" ] ; \
+      then wget http://releases.llvm.org/3.8.0/clang+llvm-3.8.0-x86_64-linux-gnu-ubuntu-16.04.tar.xz ; \
```
agurfinkel: make `wget` be less verbose for shorter logs

test/CMakeLists.txt
```diff
@@ -62,5 +62,10 @@ add_lit_testsuite(test-dsa "Regression test for DSA"
 install (DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/simple DESTINATION share/seahorn/test)
 install (DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/solve DESTINATION share/seahorn/test)
 install (DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/abc DESTINATION share/seahorn/test)
-install (DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/dsa DESTINATION share/seahorn/test)
+install (DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/dsa DESTINATION share/seahorn/test
+  PATTERN "*/check_graphs.py" EXCLUDE
+  )
 install (FILES ${CMAKE_CURRENT_SOURCE_DIR}/lit.cfg DESTINATION share/seahorn/test)
+install (FILES ${CMAKE_CURRENT_SOURCE_DIR}/dsa/check_graphs.py DESTINATION share/seahorn/test/dsa
```
agurfinkel: use `install PROGRAMS` instead of `install FILES` to install binaries. See examples in [py/CMakeLists.txt](https://github.com/seahorn/seahorn/blob/master/py/CMakeLists.txt)

docker/seahorn-build.Dockerfile
mrthefakeperson: I thought it would be faster to run `git fetch` instead of clone in the actual build file, which is why I cloned it here.

agurfinkel: if this is `seahorn-build`, then this is a base build image for all builds of seahorn. At this point, you don't know which version of the code will be build: master, dev, custom branch, etc.

On travis-ci, travis makes the appropriate version of the code available and it has to be imported into the container (you actually have that in TRAVIS specific docker already)

docker/seahorn-full-size-rel.Dockerfile
mrthefakeperson: I'm not sure about the `&& bash`, since it was in the original Dockerfile. It's probably not necessary, so I'll remove it and see what happens.

mrthefakeperson:
I may need to disable a few tests to get the build to pass. Please do not merge until that has been resolved.

agurfinkel:
Put this in the pull request message, like DNM at the top.
It might be ok to disable some tests, but not until we triage them and find out what exactly is going on.

docker/seahorn-full-size-rel.Dockerfile
agurfinkel: @kuhar : you might know the answer to this one

docker/readme.md
```diff
@@ -0,0 +1,27 @@
...
+```
+This will automatically download all dependancies from a base image (generated by `seahorn-build.Dockerfile`) and build SeaHorn under `/seahorn/build`.
+
+SeaHorn's install directory is added to PATH, and a tarball with it is ready under the current working directory (on the host).
+
+The possible build arguments are:
+- UBUNTU: trusty, xenial
+- BUILD_TYPE: Release, Debug
+
+NOTE: Debug build type is not supported at the moment.
```
kuhar: If this comes from some of my previous files, this should not longer be a case -- we have dependencies for [Debug, Release] x [trusty, xenial] hosted in seahorn-deps.

docker/seahorn-build.Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+
+RUN export PREFIX=$(cat /tmp/dockerutils/prefix.txt) && \
+    export DEPS_TAR=$(echo "$PREFIX".tar.gz) && \
+    tar -xvf $DEPS_TAR && \
+    tar -xvf boost162.tar.gz && \
+    tar -xvf z3.tar.gz && \
+    tar -xvf llvm38.tar.gz && \
+    rm *.tar.gz
+
+RUN ls -al --block-size=M 1>&2
```
kuhar: this was just for debugging and I think we can comment it out at this point

docker/seahorn-full-size-rel.Dockerfile
kuhar: The idea was that it that the container always produces the binary and copies it outside, but it also lets you play with seahorn inside the container.

mrthefakeperson:
The tests with strange behavior have been disabled, and issue #123 was created for the newly disabled tests.

docker/seahorn-full-size-rel.Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+ENV PATH "/seahorn/build/run/bin:$PATH"
+
+WORKDIR /seahorn
+
+RUN echo '#!/bin/sh' > /tmp/cpy-tarball.sh && \ 
+    echo 'cp /seahorn/build/*.tar.gz /host/ && bash' >> /tmp/cpy-tarball.sh && \
+    chmod +x /tmp/cpy-tarball.sh
+
+# Define default command.
+CMD ["/tmp/cpy-tarball.sh"]
```
agurfinkel: I don't think we need this anymore. In the current travis configuration, the image is copied out using `docker run` instead of as a side-effect of `docker-build`

docker/seahorn-build.Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+    export BT=$(cat /tmp/dockerutils/dt_out.txt) && \
+    export UB=$(lsb_release --a 2>&1 | cut -f2 | tail -n 1) && \
+    echo "$UB"_"$BT" > /tmp/dockerutils/prefix.txt && \
+    cat /tmp/dockerutils/prefix.txt && \
+    mkdir -p /deps
+
+WORKDIR /deps
+RUN export PREFIX=$(cat /tmp/dockerutils/prefix.txt) && \
+    export DEPS_LINK=$(echo https://github.com/seahorn/seahorn-ext-deps/releases/download/v0.1/"$PREFIX".tar.gz) && \
+    wget -nv $DEPS_LINK && \
```
agurfinkel: @kuhar , is this package inside a package? Not ideal. At least the top level `gz` should be removed since zipping zipped files does not make them smaller.

@mrthefakeperson: please update to use `curl -sSOL url` instead of `wget` to minimize our dependencies

docker/seahorn-full-size-rel.Dockerfile
mrthefakeperson: I was wondering if we still want to be able to build a tarball onto the host machine with Docker. In that case, it's convenient to set the default command here so that users only need to `docker run` the image after building.

PR #130, opened 2018-03-22
Run smc tests in CI

kuhar:


agurfinkel:
only one test?

kuhar:
there are more than one, but I had to xfail one of them -- I think it gets optimized out before I run my analysis. I will investigate later

PR #135, opened 2018-04-04
Add docker tools

mrthefakeperson:
- add vim.tiny, gdb and to docker image


mrthefakeperson:
In that case, it might be better to remove lldb. Vim.tiny is 5.828 MB, lldb is 257 MB, and gdb is 31.1 MB. The new packages increased the compressed image size by about 80 MB.

agurfinkel:
is the PR ready to be merged?

I believe Jakub has permissions to merge PRs as well. If so, Jakub, please
merge when you think it is ready.

PR #175, opened 2018-11-24
Remotes/origin/types 5.0 callgraph

nhamlv-55:


lib/Transforms/Instrumentation/SimpleMemoryCheck.cc
```diff
@@ -30,12 +30,6 @@ static llvm::cl::opt<bool>
                   llvm::cl::desc("Print Simple Memory Check statistics"),
                   llvm::cl::init(false));
 
-static llvm::cl::opt<bool>
```
agurfinkel: @nhamlv-55 : did you intentionally change this file? Looks like it should not be part of the PR

lib/Transforms/Utils/DevirtFunctions.cc
```diff
@@ -461,6 +521,8 @@ namespace
 namespace seahorn
 {
   Pass* createDevirtualizeFunctionsPass () {return new DevirtualizeFunctions ();}
+  Pass* createCFLAAFunctionPass() {return llvm::createCFLAndersAAWrapperPass();}
```
agurfinkel: These two functions are not needed. Use `llvm::` versions.

agurfinkel:
@nhamlv-55 can you post the example that was not working and the command line to use

lib/Transforms/Utils/DevirtFunctions.cc
```diff
@@ -195,12 +245,11 @@ namespace
     Value* ptr = CS.getCalledValue();
     SmallVector<Type*, 8> TP;
     TP.push_back (ptr->getType ());
-    for (auto i = CS.arg_begin(), e = CS.arg_end (); i != e; ++i) 
+    for (auto i = CS.arg_begin(), e = CS.arg_end (); i != e; ++i)
```
kuhar: You can make it a bit prettier and more explicit:
`for (Use *U : CS.args())`

lib/Transforms/Utils/DevirtFunctions.cc
```diff
@@ -399,7 +448,9 @@ namespace
   
   void DevirtualizeFunctions::visitCallSite (CallSite &CS)
   {
+    errs()<<"visiting:"<<CS.getInstruction()->getParent()->getParent()->getName()<<"\n";
```
kuhar: please run clang format on your changes:
https://github.com/llvm-mirror/clang/blob/master/tools/clang-format/clang-format-diff.py

PR #181, opened 2019-01-13
Readme Update

shvamath:
Readme update for packages which I had to install

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,7 +242,6 @@ This is an example of a C program annotated with a safety property:
       }
       sassert(k>=n);
     }
-
 ```
 SeaHorn follows [SV-COMP](http://sv-comp.sosy-lab.org) convention of
 encoding error locations by a call to the designated error function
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -252,6 +251,11 @@ safe. SeaHorn returns `sat` when `__VERIFIER_error()` is reachable and
 the program is unsafe. `sassert()` method is defined in
 `seahorn/seahorn.h`.
 
+## Ubutnu 18.04 Packages ##
+Before proceeding with installation make sure you have all of the following packages:
```
kuhar: I'd say these are suggested packages you may need

README.md
```diff
@@ -252,6 +251,11 @@ safe. SeaHorn returns `sat` when `__VERIFIER_error()` is reachable and
...
+* `sudo apt install cmake build-essential ninja-build llvm-6.0 clang-6.0 lld-6.0 git-svn libboost-dev subversion g++-7-multilib gcc-multilib lib32stdc++7 libgmp-dev libgmpxx4ldbl libgraphviz-dev`
```
kuhar: git-svn is not strictly necessary, while git obviously is

README.md
```diff
@@ -252,6 +251,11 @@ safe. SeaHorn returns `sat` when `__VERIFIER_error()` is reachable and
...
+
+You may also need to install the following packages if getting warning: `sudo apt install libncurses5-dev ncurses-doc`
```
kuhar: why not merge it with the previous list?

caballa: Why not clang-5.0 rather than 6.0. Also, why do we need llvm? we install our own from sources.
ninja is optional.

kuhar: the lld linker started working around the ~6.0 release; and it makes sense to take the newest package your distro offers

kuhar: I'd also argue that the default ld.bfd linker is horribly slow and nobody should ever use it in 2019

caballa: but it can happen (maybe not with 6.0) that clang generates some bitcode that our 5.0 frontend could not digest

kuhar: the intention is this is for building seahorn itself, not for generating bitcode

README.md
kuhar: or something like: suggested packages for building seahorn

agurfinkel:
Thank you! I've merged it manually. Closing the PR here.

PR #196, opened 2019-05-15
(WIP) Initialize global variables in registers

mrthefakeperson:


lib/seahorn/BvOpSem2.cc
```diff
@@ -1488,7 +1490,24 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemBase {
...
-      WARN << "Skipping initialization of a global: " << inst << "\n";
+      WARN << "initializing a global: " << inst << "\n";
+      Expr memOut = m_ctx.mkRegister(inst);
+      Expr memIn = m_ctx.getRegister(*CS.getArgument(1));
+      m_ctx.read(memIn);
+      setValue(inst, havoc(inst));
+
+      m_ctx.setMemReadRegister(memIn);
+      m_ctx.setMemWriteRegister(memOut);
+      m_ctx.setMemScalar(extractUniqueScalar(CS) != nullptr);
```
agurfinkel: don't do that. I am reusing the scalar slot for global. Just delete this line

lib/seahorn/BvOpSem2.cc
```diff
@@ -1488,7 +1490,24 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemBase {
...
+
+      llvm::GlobalVariable *gv = (GlobalVariable *)((*CS.getArgument(2)).stripPointerCasts());
```
agurfinkel: to cast, use `dyn_cast`. See examples in code, and LLVM documentation on the web.
`dyn_cast` returns `nullptr` if the argument cannot be safely converted to requested type.

lib/seahorn/BvOpSem2.cc
```diff
@@ -1488,7 +1490,24 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemBase {
...
+      Expr x = op::array::select(m_ctx.read(memIn), lookup(*gv));
+      WARN << "x: ";
+      x->dump();
```
agurfinkel: `Expr` can be printed. `dump` is only to be used from within a debugger. You can do `WARN << "x : " << *x << "\n"`

lib/seahorn/BvOpSem2.cc
```diff
@@ -1488,7 +1490,24 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemBase {
...
+      //m_ctx.MemCpy(lookup(*CS.getArgument(1)), lookup(*gv), 4, 4);
```
agurfinkel: This should be something like
```
auto gvVal = m_ctx.getGlobalVariableInitValue(gv)
if (gvVal.first) {
  m_ctx.MemCpy(lookup(*gv), gvVal.first, gvVal.second, 4);
}
```
Alignment is determined by the semantic machine since it decides how to allocate and align globals.
You may need to add API for `getGlobalVariableInitValue`. I've only added it to the memory manager, 
and it is probably not accessible directly from the context.

agurfinkel:
the code looks good, but a test fails with a memory fault...

lib/seahorn/BvOpSem2.cc
```diff
@@ -1503,7 +1511,7 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemBase {
       setValue(inst, havoc(inst));
 
       m_ctx.setMemReadRegister(memIn);
-      m_ctx.setMemWriteRegister(memOut);
+      m_ctx.setMemWriteRegister(memIn);
```
agurfinkel: the old code was correct. The current one  will generate a constraint `MEM_IN = MEM_IN with a change`. The equality is trivially false, which will cause everything else to be unsat

lib/seahorn/BvOpSem2.cc
```diff
@@ -889,8 +889,16 @@ class OpSemMemManager {
   /// \brief Executes symbolic memcpy from physical memory with concrete length
   Expr MemCpy(PtrTy dPtr, char* sPtr, unsigned len, uint32_t align) {
     assert(len % align == 0);
-    WARN << "skipping MemCpy range";
-    return nullptr;
+    Expr res = m_ctx.read(m_ctx.getMemReadRegister());
+    for (int i = 0; i < len; i++) {
+      Expr offset = bv::bvnum(i, ptrSzInBits(), m_efac);
+      Expr dIdx = i == 0 ? dPtr : mk<BADD>(dPtr, offset);
+      Expr val = bv::bvnum(sPtr[i], ptrSzInBits(), m_efac);
```
agurfinkel: this writes an `8` bit value into `ptrSzInBits()` wide memory word. This is not quite what you want.
You want to read `sPtr` one semantic machine word at a time, create `bvnum`, as  you do, and store this.

In the current configuration, both pointers and machine words are `32` bits, so that's why the `store` succeeds.
But you are still updating memory incorrectly. If `sPtr` has one 32 bit word with octets `abcd`, you are writing `000a000b000c000d` into memory. Actually not quite due to correct computation of byte addresses.
Hope you get the picture

test/opsem/muchang.gstr.ll
```diff
@@ -1,6 +1,5 @@
-; RUN: %seabmc "%s" 2>&1 | %oc %s
+; RUN: %seabmc --horn-bv2-word-size=1 "%s" 2>&1 | %oc %s
```
mrthefakeperson: TODO: load/store instructions are not word-aligned, this arg will be removed once that is implemented

lib/seahorn/BvOpSem2.cc
```diff
@@ -885,6 +886,26 @@ class OpSemMemManager {
...
 
+  /// \brief Executes symbolic memcpy from physical memory with concrete length
+  Expr MemCpy(PtrTy dPtr, char* sPtr, unsigned len, uint32_t align) {
+    Expr res = m_ctx.read(m_ctx.getMemReadRegister());
+    unsigned sem_word_sz = wordSzInBytes();
+    for (int i = 0; i < len; i += sem_word_sz) {
+      Expr offset = bv::bvnum(i, ptrSzInBits(), m_efac);
+      Expr dIdx = i == 0 ? dPtr : mk<BADD>(dPtr, offset);
+      unsigned val32 = 0;
+      for (int byte = sem_word_sz - 1; byte >= 0; byte--) {
```
agurfinkel: I think you can do this with `memcpy` without unaligned memory accesses. Minor, but will look cleaner

lib/seahorn/BvOpSem2.cc
```diff
@@ -885,6 +886,26 @@ class OpSemMemManager {
...
   }
 
+  /// \brief Executes symbolic memcpy from physical memory with concrete length
+  Expr MemCpy(PtrTy dPtr, char* sPtr, unsigned len, uint32_t align) {
+    Expr res = m_ctx.read(m_ctx.getMemReadRegister());
+    unsigned sem_word_sz = wordSzInBytes();
+    for (int i = 0; i < len; i += sem_word_sz) {
+      Expr offset = bv::bvnum(i, ptrSzInBits(), m_efac);
+      Expr dIdx = i == 0 ? dPtr : mk<BADD>(dPtr, offset);
+      unsigned val32 = 0;
```
agurfinkel: this should be `val`, not `val32` since different word sizes are possible

lib/seahorn/BvOpSem2.cc
```diff
@@ -1488,8 +1509,29 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemBase {
...
+      LOG("opsem.mem.global.init", errs() << "mem.global.init: " << inst << "\n";
+          errs() << "arg1: " << *CS.getArgument(1) << "\n";
+          errs() << "memIn: " << *memIn << ", memOut: " << *memOut << "\n";
+          );
+
+      Value *gVal = (*CS.getArgument(2)).stripPointerCasts();
+      if (auto *gv = dyn_cast<llvm::GlobalVariable>(gVal)) {
+        auto gvVal = m_ctx.getGlobalVariableInitValue(*gv);
+        if (gvVal.first) {
+          m_ctx.MemCpy(lookup(*gv), gvVal.first, gvVal.second, 4);
```
agurfinkel: alignment probably depends on the machine. This is the alignment of symbolic pointer corresponding to `gv`. 
Depends on what `galloc()` does. Probably good idea to make it `4` or even `8` aligned on all architectures. Sync with what `galloc()` does right now and comment about the relationship

lib/seahorn/BvOpSem2.cc
```diff
@@ -885,6 +886,26 @@ class OpSemMemManager {
     return res;
   }
 
+  /// \brief Executes symbolic memcpy from physical memory with concrete length
+  Expr MemCpy(PtrTy dPtr, char* sPtr, unsigned len, uint32_t align) {
+    Expr res = m_ctx.read(m_ctx.getMemReadRegister());
+    unsigned sem_word_sz = wordSzInBytes();
+    for (int i = 0; i < len; i += sem_word_sz) {
```
kuhar: nit: since both `len` and `sem_word_sz` are `unsigned`, `i` should also be `unsigned`

lib/seahorn/BvOpSem2.cc
mrthefakeperson: It looks like currently, galloc defaults to the word size of the machine and only considers an alignment when it is greater than that size. Updated and comment added to MemCpy

agurfinkel:
you can setup CLion to run clang-format automatically as you edit

PR #207, opened 2019-07-23
Alignment Fix for OpSem Load

mrthefakeperson:
#202 merge into master

### DIFF
--- a/lib/seahorn/BvOpSem2.cc
+++ b/lib/seahorn/BvOpSem2.cc
@@ -458,7 +458,7 @@ public:
         m_wordSz(WordSize), m_alignment(m_wordSz),
         m_allocaName(mkTerm<std::string>("sea.alloca", m_efac)),
         m_freshPtrName(mkTerm<std::string>("sea.ptr", m_efac)), m_id(0) {
-    assert((m_wordSz == 1 || m_wordSz == 4) && "Untested word size");
+    assert((m_wordSz == 1 || m_wordSz == 4 || m_wordSz == 8) && "Untested word size");
     assert((m_ptrSz == 4) && "Untested pointer size");
     m_nullPtr = bv::bvnum(0, ptrSzInBits(), m_efac);
     m_sp0 = bv::bvConst(mkTerm<std::string>("sea.sp0", m_efac), ptrSzInBits());
@@ -520,7 +520,6 @@ public:
   /// \brief Allocates memory on the stack and returns a pointer to it
   /// \param align is requested alignment. If 0, default alignment is used
   PtrTy salloc(unsigned bytes, uint32_t align = 0) {
-
     align = std::max(align, m_alignment);
     unsigned start = m_allocas.empty() ? 0 : m_allocas.back().m_end;
     start = llvm::alignTo(start, align);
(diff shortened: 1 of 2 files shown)


lib/seahorn/BvOpSem2.cc
```diff
@@ -765,11 +770,17 @@ class OpSemMemManager {
     Expr val = _val;
     Expr mem = m_ctx.read(memReadReg);
 
+    unsigned offsetBits = getByteAlignmentBits();
+    bool wordAligned = offsetBits == 0 || align % wordSzInBytes() == 0;
```
agurfinkel: are you sure `align % wordSzInBytes() == 0` is correct test?

lib/seahorn/BvOpSem2.cc
```diff
@@ -789,6 +800,40 @@ class OpSemMemManager {
...
+      PtrTy byteOffset = bv::extract(offsetBits - 1, 0, ptrAdd(ptr, i));
+      byteOffset = bv::zext(byteOffset, wordSzInBits() - 3);
+      // (x << 3) to get bit offset; zero extend to maintain word size
+      PtrTy bitOffset = bv::concat(byteOffset, bv::bvnum(0, 3, ptr->efac()));
+
+      // existing word at address
+      PtrTy alignedPtr = bv::concat(wordAddress, bv::bvnum(0L, offsetBits, ptr->efac()));
+      Expr existingWord = loadAlignedWordFromMem(alignedPtr, mem);
+
+      // set a byte of existing word to 0, byte should be address aligned
```
agurfinkel: this should be factored out into a separate method since there are multiple ways to implement it

lib/seahorn/BvOpSem2.cc
```diff
@@ -789,6 +800,40 @@ class OpSemMemManager {
...
+      Expr lowestByteMask = bv::bvnum(0xff, wordSzInBits(), ptr->efac());
+      Expr addressByteMask = mk<BNOT>(mk<BSHL>(lowestByteMask, bitOffset));
+      existingWord = mk<BAND>(existingWord, addressByteMask);
+
+      unsigned lowBit = i * 8;
+      Expr byteToStore = bv::extract(lowBit + 7, lowBit, val);
+
+      // shift into position for zeroed part of existing word; mask and rewrite
```
agurfinkel: factor into a separate method

lib/seahorn/BvOpSem2.cc
```diff
@@ -789,6 +800,40 @@ class OpSemMemManager {
...
+      Expr shiftedByte = mk<BSHL>(bv::zext(byteToStore, wordSzInBits()), bitOffset);
+      Expr updatedWord = mk<BOR>(existingWord, shiftedByte);
+      res = storeAlignedWordToMem(updatedWord, alignedPtr, mem);
+      mem = res;
+    }
```
agurfinkel: Add a comment on the basic idea: 
  for each byte i
      1. find word w of i
      2. load w
      3. update a byte in w
      4. write result back to memory
     
So this is quite a lot of unaligned reads and writes. 

This is ok for now. We will improve it as it becomes a performance bottleneck. But a comment about how it is working will be useful for future code reviews.

mrthefakeperson:
> Almost there. Just a few little nit picks.
> 
> Do you have any way to test the new unaligned store code?

So far, I've been hardcoding the alignment condition to false and running the tests.

lib/seahorn/BvOpSem2.cc
mrthefakeperson: `align` is the known alignment of pointers (given pointers are divisible by `align`, they are also divisible by all factors of `align`), so I believe it should be aligned to the word size (pointers are divisible by word size) as long as `align` is divisible by the word size

agurfinkel: what I am worried about is what happens when `wordSzInBytes() > align`

mrthefakeperson: as long as align isn't zero, the check will evaluate to false, which is intended in that case since a larger word size than known alignment means that the address is not necessarily word aligned

agurfinkel: got it

PR #210, opened 2019-09-10
Coverage build for seahorn

danblitzhou:
- added `Coverage.cmake` module for detecting / setting up coverage build 
- modified `CMakeList.txt` to handle building with `-DCMAKE_BUILD_TYPE=Coverage`
- added some instructions & clarifications in README.

Test: 
Ran coverage build and release build on mymble server, build was successful, example html report can be found on slack:coverage


README.md
```diff
@@ -108,6 +108,27 @@ $ export SEAHORN=<install_dir>/bin/sea
 $ cmake --build . --target test-all
 ```
 
+# Coverage #
+We can use `gcov` and `lcov` to generate test coverage information for SeaHorn. To build with coverage enabled,
+we need to run build under a different directory and use __`gcc`__ / __`g++`__ as compiler to avoid [gcc/gcov mismatching](https://stackoverflow.com/questions/12454175/gcov-out-of-memory-mismatched-version).
+
+Example steps for `opsem` tests coverage:
+1. `mkdir coverage; cd coverage`  create and enter coverage build directory
+2. `cmake -DCMAKE_INSTALL_PREFIX=run -DCMAKE_BUILD_TYPE=Coverage -DCMAKE_CXX_COMPILER=g++ -DCMAKE_C_COMPILER=gcc ../`
```
agurfinkel: why can't we get coverage with `clang` That is our preferred compiler and I thought everything should be supported with both `gcc` and `clang`

PR #212, opened 2019-09-20
Added tests to cover all the cases for visitICmpInst and visitBinaryO…

SANKEERTH26:
test_bo covers all the cases in visitBinaryOperator and test_icmpinst covers all the cases in visitICmpInst


test/opsem/test_bo.ll
```diff
@@ -0,0 +1,225 @@
...
+  store i32 %17, i32* %2, align 4, !tbaa !5
+  br label %18
+
+; <label>:18:                                     ; preds = %14, %0
+  %19 = load i32, i32* %2, align 4, !tbaa !5
+  %20 = icmp eq i32 %19, 5
+  br i1 %20, label %22, label %21
+
+; <label>:21:                                     ; preds = %18
+  call void @__VERIFIER_error() #5
```
agurfinkel: Something is wrong with this input. I don't expect any calls to `__VERIFIER_error`. One might think that this function indicates something to seahorn. Yet, under current set of options it is treated as `nd()`. It should either be removed, or this indicates some underlying issue with this test

agurfinkel:
PR title is too long. Treat PR titles / comments like git commit logs

test/opsem/test_bo.ll
SANKEERTH26: Ok, I will check the test and make changes accordingly.

SANKEERTH26:
Sorry, will take care of it from now.

agurfinkel:
This is great, but it would be better if the tests exercise more of the behaviour of the instructions. For example, the test values for `sgt` should be such that it fails if `sgt` is interpreted as `ugt`. Would be great if `add` also tests the overflow behavior. Basically, if the test would try to break the functionality, not just exercise the most common case.

PR #218, opened 2019-10-04
[expr] Update conversion from Expr to z3::ast

agurfinkel:
Updated conversion to z3::ast to be less generic, but (hopefully) more efficient
and more readable.

Refactored the code to improve compilation times. Conversion code is now
compiled into an object file, so changes to it do not require recompilation of
the whole system. Similar strategies should be applied everywhere we have
excessive use of templates and excessive amount of code in header files.

### DIFF
--- a/include/seahorn/Expr/Smt/ExprToZ.def
+++ b/include/seahorn/Expr/Smt/ExprToZ.def
@@ -1,0 +1,517 @@
+#pragma once
+
+#include "seahorn/Expr/ExprLlvm.hh"
+#include "seahorn/Expr/Smt/ExprToZ.hh"
+#include "seahorn/Support/SeaLog.hh"
+#include "seahorn/Support/SeaDebug.h"
+
+#include "boost/lexical_cast.hpp"
+#include "llvm/Support/raw_ostream.h"
+
+namespace {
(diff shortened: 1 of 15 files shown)


include/seahorn/Expr/Smt/ExprToZ.def
```diff
@@ -0,0 +1,512 @@
+#pragma once
```
caballa: Why not to move the whole file ExprToZ.def to  lib/seahorn/Smt/ExprToZ.cc?

agurfinkel: flexibility. You can include `.def` and pay price for templates, or link `ExprToZ` but then you are stuck with particular type for the cache

caballa:
I think you just need to link the unit tests with `SeaSmt`

agurfinkel:
I did, and that worked on my mac. But `gcc` on ubuntu still does not like something. Will try to reproduce locally

caballa:
It looks good now. Should we merge it?

agurfinkel:
I will do it. Don't worry. Using PR to make sure that tests pass before merging

PR #225, opened 2019-10-09
Decoupling PathBmcEngine from BmcEngine

caballa:
This will allow us to play more freely with different solvers and encodings in the PathBmc.
The negative side is that for now we cannot generate executable counterexamples if PathBmc is used.


lib/seahorn/BmcPass.cc
```diff
@@ -172,113 +186,143 @@ class BmcPass : public llvm::ModulePass {
...
-#endif
-      break;
-    }
-    case mono_bmc:
-    default:
+    if (m_engine == mono_bmc){
+
+      std::unique_ptr<OperationalSemantics> sem;
+      if (HornBv2)
+	sem = llvm::make_unique<Bv2OpSem>(efac, *this,
```
agurfinkel: something is wrong with indentation. Ran `clang-format` on everything.

lib/seahorn/BmcPass.cc
```diff
@@ -87,23 +99,25 @@ class BmcPass : public llvm::ModulePass {
   }
 
   void getAnalysisUsage(AnalysisUsage &AU) const {
-#ifdef HAVE_CRAB_LLVM
+#ifdef HAVE_CRAB_LLVM
```
agurfinkel: The logic for `getAnalysisUsage` seems confusing. Maybe it is only in the diff. 
What is included and when. Code that is commented should be removed. Why do we want to lower constant expressions at this point? Seems like a hack that has outlived its purpose.

lib/seahorn/BmcPass.cc
```diff
@@ -172,113 +186,143 @@ class BmcPass : public llvm::ModulePass {
...
+	  llvm::errs() << "VC:\n"
+	               << z3_to_smtlib(bmc.zctx(), vc) << "\n~~~~\n"
+	               << "Simplified VC:\n"
+	               << z3_to_smtlib(bmc.zctx(), vc_simpl) << "\n");
+      
+      if (m_out)
+	bmc.toSmtLib(*m_out);
+
+      if (!m_solve) {
+	LOG("bmc", errs() << "Stopping before solving\n";);
```
agurfinkel: indentation again

lib/seahorn/BmcPass.cc
```diff
@@ -172,113 +186,143 @@ class BmcPass : public llvm::ModulePass {
...
-        });
-
-    LOG("cex", if (res) {
-      errs() << "Analyzed Function:\n" << F << "\n";
-      BmcTrace trace(bmc->getTrace());
-      errs() << "Trace \n";
-      trace.print(errs());
-    });
 
+      if (res)
```
agurfinkel: indentation all the way from here down

tools/seahorn/seahorn.cpp
```diff
@@ -143,13 +142,19 @@ static llvm::cl::opt<bool>
             "Use BMC. Currently restricted to intra-procedural analysis"),
         llvm::cl::init(false));
 
-static llvm::cl::opt<seahorn::bmc_engine_t>
+// Available BMC engines
+typedef enum {
```
agurfinkel: I realy like new class enums of C++11. Do you mind switching to those. Otherwsie enum values pollute the global namespace.

PR #232, opened 2019-10-21
Fix compilation error in Debug mode

igcontreras:


agurfinkel:
Better fix is to remove the call to `llvm_unreachable`. I will commit a fix shortly. Thanks!

PR #234, opened 2019-10-23
Improve vcgen for inter-procedural analysis

igcontreras:


include/seahorn/Transforms/Instrumentation/ShadowMemSeaDsa.hh
```diff
@@ -1,22 +1,51 @@
 /// This file should be completely moved into C++ file.
+// TODO: remove this comment now? we want to use this pass in the vcgen
```
agurfinkel: yes

include/seahorn/Transforms/Instrumentation/ShadowMemSeaDsa.hh
```diff
@@ -1,22 +1,51 @@
...
+  ~ShadowMemSeaDsa();
 
   bool runOnModule(llvm::Module &M) override;
 
   bool runOnFunction(llvm::Function &F);
   void getAnalysisUsage(llvm::AnalysisUsage &AU) const override;
   llvm::StringRef getPassName() const override { return "ShadowMemSeaDsa"; }
+
+  // TODO: should this be a cell?
+  bool hasShadowId(const sea_dsa::Node *);
```
agurfinkel: I am not sure you can/need to go from DSA nodes to shadow ids. But you do want the opposite, 
given a `shadow.mem` call site, you want to know what DSA nodes it corresponds to. I would do this 
by providing a class, like CallSite does, that wraps a shdow.mem instruction and provides the necessary information to clients.

include/seahorn/Transforms/Instrumentation/ShadowMemSeaDsa.hh
```diff
@@ -1,22 +1,51 @@
...
+  unsigned getNodeShadowId(const sea_dsa::Node *);
+
+  sea_dsa::Graph & getSummaryGraph(const llvm::Function &F);
```
agurfinkel: What is the difference between summary graph and DsaGraph?

include/seahorn/UfoOpMemSem.hh
```diff
@@ -0,0 +1,83 @@
+#pragma once
```
agurfinkel: Why is `UfoOpMemSem` has been pulled into a header file?

lib/Transforms/Instrumentation/ShadowMemSeaDsa.cc
```diff
@@ -48,6 +48,15 @@ llvm::cl::opt<bool>
                      llvm::cl::desc("Use TypeBasedAA in the MemSSA optimizer"),
                      llvm::cl::init(true));
 
+namespace seahorn {
+bool InterProcMemFlag;
+}
+
+static llvm::cl::opt<bool,true>  InterProcMem("horn-inter-proc-mem",
```
agurfinkel: you do not need `1InterProcMemFlag`. You can use `InterProcMem` as a boolean directly.

lib/Transforms/Instrumentation/ShadowMemSeaDsa.cc
```diff
@@ -163,6 +172,13 @@ class LocalAAResultsWrapper {
 };
 
 class ShadowDsaImpl : public InstVisitor<ShadowDsaImpl> {
+public:
+  using NodeIdMap = llvm::DenseMap<const sea_dsa::Node *, unsigned>;
+
+  /// \brief A map from DsaNode to its numeric id
+  NodeIdMap m_nodeIds;
```
agurfinkel: `m_nodeIds` should be private.

I am not sure why `NodeIdMap` type should be public either

lib/Transforms/Instrumentation/ShadowMemSeaDsa.cc
```diff
@@ -478,6 +491,20 @@ class ShadowDsaImpl : public InstVisitor<ShadowDsaImpl> {
     return *ci;
   }
 
+  CallInst &mkArgNewModNode(IRBuilder<> &B, Constant *argFn, const dsa::Cell &c,
```
agurfinkel: We decided to revert this, to not include extra arguments in `shadow.mem` calls

lib/Transforms/Instrumentation/ShadowMemSeaDsa.cc
```diff
@@ -1357,11 +1396,17 @@ void ShadowDsaImpl::solveUses(Function &F) {
...
+}
+
+ShadowMemSeaDsa::~ShadowMemSeaDsa() {}
+
 bool ShadowMemSeaDsa::runOnModule(llvm::Module &M) {
   if (M.begin() == M.end())
     return false;
 
-  dsa::GlobalAnalysis &dsa = getAnalysis<dsa::DsaAnalysis>().getDsaAnalysis();
+  m_dsa = &getAnalysis<dsa::DsaAnalysis>().getDsaAnalysis();
```
agurfinkel: use `getAnalysisIfAvailable<>` which only returns analysis if it has been scheduled by the pass manager.
On the other hand, `getAnalysis<>` will schedule an additional run of the analysis if necessary

lib/Transforms/Instrumentation/ShadowMemSeaDsa.cc
```diff
@@ -1374,8 +1419,8 @@ bool ShadowMemSeaDsa::runOnModule(llvm::Module &M) {
   LOG("shadow_verbose", errs() << "Module before shadow insertion:\n"
                                << M << "\n";);
 
-  ShadowDsaImpl impl(dsa, asi, tli, cg, *this, SplitFields, LocalReadMod);
-  bool res = impl.runOnModule(M);
+  m_shadow = std::unique_ptr<ShadowDsaImpl>(new ShadowDsaImpl(*m_dsa, asi, tli, cg, *this, SplitFields, LocalReadMod));
```
agurfinkel: use `llvm::make_unique<>` instead of `std::unique_ptr`

include/seahorn/Transforms/Instrumentation/ShadowMemSeaDsa.hh
igcontreras: The summary graph is the bottom up graph. Should I call it directly BUDsaGraph?

agurfinkel: maybe just a comment on the function will do. I don't have a better name. I prefer current name to `BUDsa`.

include/seahorn/UfoOpMemSem.hh
igcontreras: There were originally an `UfoOpSem.hh` (and `UfoOpSem.cc`), so I used the same structure in my copy. Should I move `UfoOpMemSem.hh` to `UfoOpSem.hh`?

lib/Transforms/Instrumentation/ShadowMemSeaDsa.cc
agurfinkel: I see that the flag is needed because it is accessed elsewhere. Not ideal, but ok for now. Should add comment to that effect here where the variable is created.

include/seahorn/UfoOpMemSem.hh
agurfinkel: No. This is good. Just add a comment that this is a clone of `UfoOpSem` for the purpose of prototyping. I got confused and thought this was an existing class.

igcontreras: OK!

include/seahorn/Transforms/Instrumentation/ShadowMemSeaDsa.hh
igcontreras: OK, I will add comments.

include/seahorn/Transforms/Instrumentation/ShadowMemSeaDsa.hh
igcontreras: I think that this is needed for efficiency. When traversing the BU graph, it's useful to get directly the id of all the reachable nodes from each actual parameter in the callsite, so that we know where to copy.

lib/Transforms/Instrumentation/ShadowMemSeaDsa.cc
igcontreras: This was needed for https://github.com/seahorn/seahorn/pull/234#discussion_r338634001

include/seahorn/Transforms/Instrumentation/ShadowMemSeaDsa.hh
agurfinkel: I need to understand the use case better. Note that there is no one-to-one correspondence between shadow ids and nodes. Different fields of a node might have different ids. Of course we can ignore this in the first implementation, but it would be nicer if we have an interface that makes it simpler.

igcontreras: Yes, in fact, I think this should be in terms of cells, as you mention, different fields may have different ids.
I was considering first the case in which only one logical array is created per dsa node (that is why I wrote the comment) because I thought that it was easier, but I will change it to cell.

lib/Transforms/Instrumentation/ShadowMemSeaDsa.cc
```diff
@@ -1146,7 +1164,7 @@ CallInst &ShadowDsaImpl::getParentDef(CallInst &memOp) {
 
   auto *fn = memOp.getCalledFunction();
   assert(fn);
-  assert(fn == m_memLoadFn || fn == m_memStoreFn || fn == m_memTrsfrLoadFn);
+  //assert(fn == m_memLoadFn || fn == m_memStoreFn || fn == m_memTrsfrLoadFn);
```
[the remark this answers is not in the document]
igcontreras: I will uncomment it, sorry

lib/Transforms/Instrumentation/ShadowMemSeaDsa.cc
```diff
@@ -48,9 +48,18 @@ llvm::cl::opt<bool>
...
+
+static llvm::cl::opt<bool,true>  InterProcMem("horn-inter-proc-mem",
+                 llvm::cl::desc("Use inter-procedural encoding with memory"),
+                                              llvm::cl::location(seahorn::InterProcMem),
+                                              llvm::cl::init(false));
+
 using namespace llvm;
 namespace dsa = sea_dsa;
-namespace {
+namespace seahorn {
```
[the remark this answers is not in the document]
igcontreras: It was intentional but because I thought it was needed, since it's not I will undo this.

CMakeLists.txt
```diff
@@ -68,7 +68,7 @@ if (GIT_FOUND)
     CACHE STRING "seahorn-llvm repo")
   set (LLVM_DSA_REPO "https://github.com/seahorn/llvm-dsa"
     CACHE STRING "llvm-dsa repo")
-  set (SEA_DSA_REPO "https://github.com/seahorn/sea-dsa"
+  set (SEA_DSA_REPO "https://github.com/igcontreras/sea-dsa"
```
igcontreras: This is just a comment to remember to undo this change

lib/Transforms/Instrumentation/ShadowMemSeaDsa.cc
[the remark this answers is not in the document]
agurfinkel: It would be good to fix the code as well. Please log an issue in `sea-dsa` with code on which the assertion fails. Remember that until it is fixed, you should disable this `sea-dsa` optimization.

tools/seahorn/seahorn.cpp
```diff
@@ -313,8 +313,13 @@ int main(int argc, char **argv) {
...
   // -- initialize any global variables that are left
   pass_manager.add(new seahorn::LowerGvInitializers());
+
+  // TODO: get rid of the deprecaed dsa pass??
+
+  // the ShadowMemSeaDsa pass is now added automatically because it is required
+  // by HornifyModule
   if (SeaHornDsa) {
-    pass_manager.add(seahorn::createShadowMemSeaDsaPass());
+    //    pass_manager.add(seahorn::createShadowMemSeaDsaPass());
```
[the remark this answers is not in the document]
igcontreras: This is now uncommented

lib/seahorn/UfoOpMemSem.cc
```diff
@@ -0,0 +1,1377 @@
...
+    Expr origE = getInArraySymbol(node_id);
+
+    auto it = m_rep.find(node_id);
+    if (it == m_rep.end()) { // not copied yet
+      assert(bind::isArrayConst(origE));
+      Expr name = bind::fname(origE);
+      Expr rTy = bind::rangeTy(name);
+
+      Expr copyE = bind::mkConst(
+          mkTerm<std::string>("copy_" + std::to_string(++m_copy_count), m_efac),
```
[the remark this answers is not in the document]
igcontreras: Done

lib/Transforms/Instrumentation/ShadowMemSeaDsa.cc
```diff
@@ -1394,6 +1426,66 @@ void ShadowMemSeaDsa::getAnalysisUsage(llvm::AnalysisUsage &AU) const {
...
+
+const llvm::StringRef ShadowMemSeaDsa::ARG_NEW = "shadow.mem.arg.new";
+const llvm::StringRef ShadowMemSeaDsa::ARG_READ = "shadow.mem.arg.ref";
+const llvm::StringRef ShadowMemSeaDsa::ARG_MOD = "shadow.mem.arg.mod";
+
+// bool ShadowMemSeaDsa::hasShadowId(const sea_dsa::Cell *c) {
+//   auto it = m_shadow->m_nodeIds.find(c->getNode());
+//   return it != m_shadow->m_nodeIds.end();
+// }
+unsigned ShadowMemSeaDsa::getCellShadowId(const dsa::Cell &c) {
```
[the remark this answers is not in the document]
igcontreras: Done

igcontreras:
The small examples are now working with the transformation!

include/seahorn/Transforms/Instrumentation/ShadowMemSeaDsa.hh
```diff
@@ -1,23 +1,66 @@
...
+  bool shadowInstrIsCallSiteParam(llvm::CallSite &cs);
+  bool shadowInstrWrites(llvm::CallSite &cs);
+  bool shadowInstrReads(llvm::CallSite &cs);
+  llvm::Value * getShadowInAlloc(llvm::CallSite &cs);
+  llvm::Value *getShadowOutAlloc(llvm::CallSite &cs);
+
+  const sea_dsa::Cell getCellShadowInstr(llvm::CallSite &cs);
+
+  // API of graphs/shadow ids
+  bool hasShadowId(const sea_dsa::Cell *);
```
agurfinkel: should probably be `const sea_dsa::Cell &`. I don't immediately see what is the benefit of passing `nullptr` to `hasShadowId`

lib/Transforms/Instrumentation/ShadowMemSeaDsa.cc
```diff
@@ -1353,15 +1355,60 @@ void ShadowDsaImpl::solveUses(Function &F) {
...
+  const sea_dsa::Node *n = c.getNode();
+  unsigned offset = c.getOffset();
+  auto it = m_nodeIds.find(n);
+  if (it != m_nodeIds.end()) {
+    id = it->second + offset;
+  }
+  else{
+    m_nodeIds[n] = id;
+    if (n->size() == 0) {
+      // XXX: what does it mean for a node to have 0 size?
```
agurfinkel: all properties of dsa-nodes are by use. So if there is a pointer that appears in the program but is never used (i.e., read/stored) then the size is 0.

Size is also 0 if a node is collapsed -- we are not sure what size it is at all.

lib/Transforms/Instrumentation/ShadowMemSeaDsa.cc
```diff
@@ -1353,15 +1355,60 @@ void ShadowDsaImpl::solveUses(Function &F) {
...
+      ++m_maxId;
+    } else {
+      // -- allocate enough ids for the node and all its fields
+      m_maxId += n->size();
+      id = id + offset;
+    }
+  }
+  auto it2 = m_shadowsReverse.find(id);
+  if(it2 == m_shadowsReverse.end()){ // add to the reverse map if needed
+    m_shadowsReverse.insert(std::make_pair(id, std::make_pair(c.getNode(), c.getRawOffset())));
```
agurfinkel: You should be able to use `{c.getNode(), c.getRawOffset()}` instead of `std::make_pair`

tools/seahorn/seahorn.cpp
```diff
@@ -312,6 +312,11 @@ int main(int argc, char **argv) {
 
   // -- initialize any global variables that are left
   pass_manager.add(new seahorn::LowerGvInitializers());
+
+  // TODO: get rid of the deprecaed dsa pass??
```
agurfinkel: yes! The new pass has been around for a while and seems to work just fine

tools/seahorn/seahorn.cpp
```diff
@@ -312,6 +312,11 @@ int main(int argc, char **argv) {
...
+
+  // the ShadowMemSeaDsa pass is now added automatically because it is required
+  // by HornifyModule
```
agurfinkel: Not sure what this comment means. But note that `HornifyModule` is not always loaded -- for example when we do Bmc. While Dsa and Shadow are still required.

igcontreras: This comment is outdated, I forgot to remove it.

include/seahorn/Transforms/Instrumentation/ShadowMemSeaDsa.hh
igcontreras: Yes, in fact, this is now not used since the shadow internal id is not accessible, I will remove it.

lib/Transforms/Instrumentation/ShadowMemSeaDsa.cc
igcontreras: Thanks! I added a comment for this!

PR #240, opened 2019-11-05
Do not remove the content of --temp-dir if --save-temps was not specified

igcontreras:
I accidentally misspelled the option `--save-temps` and, as a result, all the content of the `--temp-dir` was removed, without any messages. This is really dangerous when running the option as, e.g., `--temp-dir .`, because `shutil.rmtree` will recursively remove everything in that directory.

### DIFF
--- a/py/sea/__init__.py
+++ b/py/sea/__init__.py
@@ -90,8 +90,10 @@ def createWorkDir (dname=None, save=False, prefix='tmp-'):
         if not os.path.isdir (dname): os.mkdir (dname)
         workdir = dname
 
-    if not save:
+    if dname is None:
         atexit.register (shutil.rmtree, path=workdir)
+    else:
+        print "Warning: --temp-dir specified without the --save-temps option"
     return workdir
 
 def add_help_arg (ap):


agurfinkel:
I'm reluctant to accept this PR. It is strange to specify temp directory without saving it. The only immediate use case is to force temp files to reside on some special partition. I recall that I took those options from some other tools I was using at the time. I'm worried that we have some portfolio scripts that might depend on it. 

The options are only used for debugging. You should always specify a throwaway directory as temp. Never `.`. 

If this is highly desired, I would instead remove `--save-temps` and say that if `--temp-dir` is given, then it is not deleted, if temp directory is created by the script, then it is deleted.

caballa:
I always use `.` ;-)

caballa:
But I agree that 

"If this is highly desired, I would instead remove --save-temps and say that if --temp-dir is given, then it is not deleted, if temp directory is created by the script, then it is deleted."

is a better solution

igcontreras:
In the pull request the behavior of the script is what you suggest (except for the `--save-temps` option):
>  if `--temp-dir` is given, then it is not deleted, if temp directory is created by the script, then it is deleted

Another option is, if the use case is "specify the temp dir but leave it clean after the execution", then add an option to remove the content of the temp file even if specified.

Issue #252, opened 2019-11-19
[seaopt] type of global variable changed unexpectedly by seaopt


danblitzhou:
During the process of `sea pf` pipeline, `seaopt` converted the type of a global variable from `i32` to `i1`, which caused other problems down the line of the pipeline: (possibly during BmcPass)
```
Error: failed to convert to z3: (select(main@%sm67, @msg_buf_size)_0=(1:bv(1)))

Error: Sorts Bool and (_ BitVec 1) are incompatible
```
Example LLVM IR file for repro https://gist.github.com/danblitzhou/0a81e0b9cfb5b7fd8b6145c3c2924a95
command: 
`seaopt -f -funit-at-a-time --simplifycfg-sink-common=false -o  int_bool.out.bc  -O3 --enable-indvar=false --enable-loop-idiom=false --enable-nondet-init=false --unroll-threshold=150 --disable-loop-vectorization=true --disable-slp-vectorization=true int_bool.bc -S`

agurfinkel:
@danblitzhou is this still an issue? I fixed the conversion failure error. Do you still believe that the optimization is incorrect / suspicious?

danblitzhou:
I think the optimization is correct; can no longer reproduce the crash after the fix in opsem. Closing.

PR #253, opened 2019-11-20
Cover loading from / storing into i1 type 

danblitzhou:
Added lit test cases that covers a z3/llvm bool conversion issue introduced by load / store instruction on i1 type variables described in #252 , fixed by [cb860197](https://github.com/seahorn/seahorn/commit/cb860197ba62895b56cb1777c490cf380e35e040)

Original C code: 
```
#include "seahorn/seahorn.h"
extern int nd(void);

int un_init;

int get_y(int x) {
        if (x > un_init) {
                un_init = x;
        }
        return 0;
}

int main(void) {
 int x=1;
 int y = get_y(x);

 while(nd()) {
   int t1 = x;
   int t2 = y;
   if (t1 > un_init) {
        x = t1+ t2;
        y = t1 + t2;
   }
  }
   sassert(x > y);
   return 0;
 }
```

Note the type conversion of `un_init` (i32 to i1) done by the optimization process


agurfinkel:
thank you!

PR #257, opened 2019-11-25
Generate LLVM harness from BMC counterexample

caballa:


lib/seahorn/BmcPass.cc
```diff
@@ -243,12 +254,29 @@ class BmcPass : public llvm::ModulePass {
...
-        trace.print(errs());
-      });
+	  errs() << "Analyzed Function:\n" << F << "\n";
+	  errs() << "Trace \n";
+	  BmcTrace trace(bmc.getTrace());	  
+	  trace.print(errs());
+	});
+
+      if (res) {
+	StringRef CexFileRef(HornCexFile);
```
agurfinkel: formatting

caballa: This is some issue with git diff. In any case, I prefer the first PR without clang-format and then after approval to run clang-format.

agurfinkel: Not sure it is just git diff. I think your emacs is configured differently from mine. I often get small formatting issues like this. As long as you run clang-format eventually, this is all good.

I haven't seen the corresponding option in `sea` script. Did you forget to commit it?

PR #265, opened 2019-12-20
Improve vcgen for inter-procedural analysis

igcontreras:


include/seahorn/InterMemPreProc.hh
```diff
@@ -0,0 +1,39 @@
...
+
+  NodesCSMap m_safen_cs_callee; // set of unsafe nodes in the callee of callsite
+  NodesCSMap m_safen_cs_caller; // set of unsafe nodes in the caller of callsite
+
+  using NodeFMap = DenseMap<const Function *, std::unique_ptr<NodeSet>>;
+
+  NodeFMap m_safen_f_callee; // set of unsafe nodes in the callee of a function
+
+public:
+  InterMemPreProc(CompleteCallGraph *ccg, ShadowMem &shadowDsa)
```
agurfinkel: why is `CompleteCallGraph` passed by pointer but `ShadowMem` by reference?

include/seahorn/UfoOpMemSem.hh
```diff
@@ -0,0 +1,90 @@
+#pragma once
+// This is a copy of UfoOpSem for prototyping purposes
```
agurfinkel: `MemUfoOpSem` would be a better name. `OpSem` stands for *operational semantics*, so puting `Mem` in between does not work

agurfinkel: Can this be refactored into a base template and two implementations so that code that is common between `UfoOpSem` and `UofOpMemSem` is shared

lib/seahorn/HornifyModule.cc
```diff
@@ -92,6 +98,27 @@ static llvm::cl::list<std::string>
                       llvm::cl::desc("Abstract all calls to these functions"),
                       llvm::cl::ZeroOrMore, llvm::cl::CommaSeparated);
 
+static llvm::cl::opt<bool, true> InterProcMem(
+    "horn-inter-proc-mem",
+    llvm::cl::desc("Use inter-procedural encoding with memory"), llvm::cl::init(false));
+
+namespace seahorn {
+// counters for copying the new inter-proc vcgen
+// only updated if the log "inter_mem_counters" is active
```
agurfinkel: these seem like global variables. So names should be `g_`. 

However, this brings another question, why do we need global variables? Why can't they be members of `HornifyModule`?

lib/seahorn/HornifyModule.cc
```diff
@@ -142,6 +169,19 @@ bool HornifyModule::runOnModule(Module &M) {
   if (Step == hm_detail::CLP_SMALL_STEP ||
       Step == hm_detail::CLP_FLAT_SMALL_STEP)
     m_sem.reset(new ClpOpSem(m_efac, *this, M.getDataLayout(), TL));
+  else if (InterProcMem) {
+    ShadowMemPass * smp = getAnalysisIfAvailable<sea_dsa::ShadowMemPass>();
+    assert(smp);
+    ShadowMem &shadowmem_analysis =smp->getShadowMem();
```
agurfinkel: space after `=`. Looks like `clang-format` was not run

lib/seahorn/HornifyModule.cc
```diff
@@ -142,6 +169,19 @@ bool HornifyModule::runOnModule(Module &M) {
...
+
+    CompleteCallGraph *ccg = getAnalysisIfAvailable<sea_dsa::CompleteCallGraph>();
+    std::shared_ptr<InterMemPreProc> preproc(new InterMemPreProc(ccg, shadowmem_analysis));
```
agurfinkel: use `std::make_shared`

agurfinkel: or maybe `std::unique`?

lib/seahorn/HornifyModule.cc
```diff
@@ -142,6 +169,19 @@ bool HornifyModule::runOnModule(Module &M) {
...
+
+    preproc->runOnModule(M);
+
+    m_sem.reset(new UfoOpMemSem(m_efac, *this, M.getDataLayout(), preproc, TL,
+                                abs_fns, &shadowmem_analysis));
```
agurfinkel: A better design would be to add setters on `UfoOpMemSem` to communicate extra analyses that it needs.
Then, `UfoOpMemSem` works like `UfoOpSem` when the extra fields are not set, and like itself when they are.
This will also eliminate the need for two versions of `UfoOpSem`

lib/seahorn/HornifyModule.cc
```diff
@@ -279,12 +319,40 @@ bool HornifyModule::runOnModule(Module &M) {
...
+      });
+
   if (!m_db.hasQuery()) {
     // --- This may happen if the exit block of main is unreachable
     //     but still the main function can fail.
     m_db.addQuery(mk<TRUE>(m_efac));
   }
 
+  // DEBUG: printing clauses
+  LOG("print_clauses", errs() << "------- PRINTING CLAUSE DB ------\n";);
```
agurfinkel: Use one `LOG` statement with many instructions instead of multiple `LOG`

lib/seahorn/HornifyModule.cc
```diff
@@ -372,9 +440,35 @@ bool HornifyModule::runOnFunction(Function &F) {
   /// -- allocate LiveSymbols
   auto r = m_ls.insert(std::make_pair(&F, LiveSymbols(F, m_efac, *m_sem)));
   assert(r.second);
+
+  // HACK because reset counters because "run()" calls VisitCallSite
+  // TODO: store part of what is computed by LiveSymbols?
+  unsigned tmp_m_n_params = m_n_params;
```
agurfinkel: create a class for all the counters, say called `stats` and manipulate that class instead of the individual counters.

lib/seahorn/InterMemPreProc.cc
```diff
@@ -0,0 +1,78 @@
...
+
+  dsaCG.buildDependencies(); // TODO: this is already done already in other passes
+  const GlobalAnalysis &ga = m_shadowDsa.getDsaAnalysis();
+
+  for (auto &F : M) {
+    if (!ga.hasGraph(F))
+      continue;
+
+    // store the contex insensitive unsafe nodes
+    std::unique_ptr<NodeSet> ci_unsafe_callee(new NodeSet());
```
agurfinkel: `llvm::make_unique`

lib/seahorn/InterMemPreProc.cc
```diff
@@ -0,0 +1,78 @@
...
+
+      ColorMap color_callee, color_caller;
+      NodeSet f_node_safe;
+
+      const Function *f_caller = it->getCallSite().getCaller();
+
+      const Graph &callerG = ga.getGraph(*f_caller);
+      const Graph &calleeG = ga.getSummaryGraph(F);
+
+      std::unique_ptr<SimulationMapper> simMap(new SimulationMapper());
```
agurfinkel: `llvm::make_unique`

lib/seahorn/HornifyModule.cc
igcontreras: The `HornifyModule` class is not available inside of `UfoOpSem`, and the counting takes place there.

Ideally, they should not be accessed in `HornifyModule`. However, the processing done in `UfoOpSem` is done twice for each callsite in `runOnFunction` because of `r.first->second.run();` and `hf->runOnFunction(F);`, the counters are counted twice if we don't do this.

lib/seahorn/HornifyModule.cc
igcontreras: I had to make a shared pointer because the `LegacyOperationalSemantics` has a constructor: 
```
  explicit LegacyOperationalSemantics(const LegacyOperationalSemantics &o)
      : OperationalSemantics(o){};
```

lib/seahorn/HornifyModule.cc
igcontreras: ok!

lib/seahorn/HornifyModule.cc
igcontreras: I'm working on this

include/seahorn/InterMemPreProc.hh
igcontreras: I forgot to change it, both will be passed by reference now.

lib/seahorn/HornifyModule.cc
agurfinkel: I don't understand the answer. If they are only modified by `UfoOpSem`, they should be defined there. If they need to be reset from `HornifyModule`, then `UfoOpSem` can provide API to reset the counters. In any case, the name should change to reflect that they are global, if they are to stay like that for now.

lib/seahorn/HornifyModule.cc
agurfinkel: This does not seem to be a field. At any rate, a comment can be added to address the choice of `shared_ptr` in here.

include/seahorn/UfoOpMemSem.hh
igcontreras: Done

lib/seahorn/HornifyModule.cc
igcontreras: done

caballa:
@igcontreras: after merging the PR in sea-dsa this PR passes all checks. Is it ready to review?

igcontreras:
In `InterMemPreProc` I have to obtain the callsites in the same way as I was doing it seadsa for coloring, I am not usign the `CompleteCallGraph`. The rest is ready.

PR #268, opened 2020-01-14
Fatptr runtime executable

danblitzhou:
- added new transformation `ExtendedPointer.cc` to add fat ptr related instructions
- added both hashmap and pointer bits based runtime implementation for get/set fatptr

### DIFF
--- a/lib/Transforms/Instrumentation/ExtendedPointer.cc
+++ b/lib/Transforms/Instrumentation/ExtendedPointer.cc
@@ -33,11 +33,12 @@
 using namespace llvm;
 
 #define DEBUG_TYPE "sea-bounds-checking"
-#define SEA_SET_FAT_SLOT0 "__sea_set_extptr_slot0"
-#define SEA_GET_FAT_SLOT0 "__sea_get_extptr_slot0"
-#define SEA_SET_FAT_SLOT1 "__sea_set_extptr_slot1"
-#define SEA_GET_FAT_SLOT1 "__sea_get_extptr_slot1"
-#define SEA_COPY_FAT_SLOTS "__sea_copy_extptr_slots"
+#define SEA_SET_FAT_SLOT0 "__sea_set_extptr_slot0_fp"
+#define SEA_GET_FAT_SLOT0 "__sea_get_extptr_slot0_fp"
+#define SEA_SET_FAT_SLOT1 "__sea_set_extptr_slot1_fp"
+#define SEA_GET_FAT_SLOT1 "__sea_get_extptr_slot1_fp"
+#define SEA_COPY_FAT_SLOTS "__sea_copy_extptr_slots_fp"
+#define SEA_RECOVER_FAT_PTR "__sea_recover_pointer_fp"
 
 
 //static cl::opt<bool> SingleErrorBB("bounds-checking-single-trap",
@@ -74,6 +75,8 @@ private:
   Function* m_getFatSlot0; // get ptr base
   Function* m_getFatSlot1; // get ptr size
   Function* m_copyFatSlots; // copy ptr info
+  Function* m_recoverFatPtr; // NEW: for ptr embed method only
+  Value* OrigPtr;
 
 
   BasicBlock *getErrorBB();
@@ -150,8 +153,7 @@ void ExtendedPointer::emitBranchToTrap(Value *Cmp) {
 bool ExtendedPointer::instrument(Value *Ptr, Value *InstVal,
                                 const DataLayout &DL) {
   uint64_t NeededSize = DL.getTypeStoreSize(InstVal->getType());
-  Type *IntTy = DL.getIntPtrType(Ptr->getType());
-  Value *NeededSizeVal = ConstantInt::get(IntTy, NeededSize);
+  Value *NeededSizeVal = ConstantInt::get(IntptrTy, NeededSize);
   DEBUG(dbgs() << "Instrument " << *Ptr << " for " << Twine(NeededSize)
                << " bytes\n");
 
@@ -162,13 +164,15 @@ bool ExtendedPointer::instrument(Value *Ptr, Value *InstVal,
     // get start and end by calling internalized functions
     Value *Start = Builder->CreateCall(m_getFatSlot0, Builder->CreateBitCast(Ptr, Builder->getInt8PtrTy()));
     Value *Size = Builder->CreateCall(m_getFatSlot1, Builder->CreateBitCast(Ptr, Builder->getInt8PtrTy()));
-    Value *PtrInt = Builder->CreatePtrToInt(Ptr, IntTy);
+    Value *recov = Builder->CreateCall(m_recoverFatPtr, Builder->CreateBitCast(Ptr, Builder->getInt8PtrTy()));
+    OrigPtr = Builder->CreateBitCast(recov, Ptr->getType());
+    Value *PtrInt = Builder->CreatePtrToInt(OrigPtr, IntptrTy);
     // Ptr >= Start
-    Value *CmpUnderFlow = Builder->CreateICmpULT(PtrInt, Builder->CreateIntCast(Start, IntTy, true));
+    Value *CmpUnderFlow = Builder->CreateICmpULT(PtrInt, Builder->CreateIntCast(Start, IntptrTy, true));
     // Start + Size >= Ptr + NeededSize
     Value *accessEnd = Builder->CreateAdd(PtrInt, NeededSizeVal);
     Value *ptrEnd = Builder->CreateAdd(Start, Size);
-    Value *CmpOverFlow = Builder->CreateICmpULT(Builder->CreateIntCast(ptrEnd, IntTy, true), accessEnd);
+    Value *CmpOverFlow = Builder->CreateICmpULT(Builder->CreateIntCast(ptrEnd, IntptrTy, true), accessEnd);
     Or = Builder->CreateOr(CmpUnderFlow, CmpOverFlow);
   } else {
     // size and offest statically computed
@@ -189,9 +193,10 @@ bool ExtendedPointer::instrument(Value *Ptr, Value *InstVal,
     Value *Cmp3 = Builder->CreateICmpULT(ObjSize, NeededSizeVal);
     Or = Builder->CreateOr(Cmp2, Cmp3);
     if (!SizeCI || SizeCI->getValue().slt(0)) {
-      Value *Cmp1 = Builder->CreateICmpSLT(Offset, ConstantInt::get(IntTy, 0));
+      Value *Cmp1 = Builder->CreateICmpSLT(Offset, ConstantInt::get(IntptrTy, 0));
       Or = Builder->CreateOr(Cmp1, Or);
     }
+    OrigPtr = Ptr;
   }
   emitBranchToTrap(Or);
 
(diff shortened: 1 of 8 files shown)


lib/Transforms/Instrumentation/ExtendedPointer.cc
```diff
@@ -0,0 +1,409 @@
...
+               << " bytes\n");
+
+  SizeOffsetEvalType SizeOffset = ObjSizeEval->compute(Ptr);
+  Value *Or;
+  if (!ObjSizeEval->bothKnown(SizeOffset)) {
+    WARN << "fatptr instrument " << *Ptr << " for " << Twine(NeededSize) << " bytes\n";
+    // get start and end by calling internalized functions
+    Value *Start = Builder->CreateCall(m_getFatSlot0, Builder->CreateBitCast(Ptr, Builder->getInt8PtrTy()));
+    Value *Size = Builder->CreateCall(m_getFatSlot1, Builder->CreateBitCast(Ptr, Builder->getInt8PtrTy()));
+    Value *recov = Builder->CreateCall(m_recoverFatPtr, Builder->CreateBitCast(Ptr, Builder->getInt8PtrTy()));
```
priyasiddharth: needed for embedded pointer fat pointer impl (not needed for hashmap fat ptr impl)

lib/Transforms/Instrumentation/ExtendedPointer.cc
```diff
@@ -0,0 +1,409 @@
...
+  IRBuilder<>::InsertPointGuard Guard(*Builder);
+  ErrorBB = BasicBlock::Create(Fn->getContext(), "bound_overflow", Fn);
+  Builder->SetInsertPoint(ErrorBB);
+
+//  llvm::Value *F = Intrinsic::getDeclaration(Fn->getParent(), Intrinsic::trap);
+  AttrBuilder AB;
+  AB.addAttribute(Attribute::NoReturn);
+  AttributeList as = AttributeList::get(ctx, AttributeList::FunctionIndex, AB);
+  auto errorFn = dyn_cast<Function>(
+      Md->getOrInsertFunction("__VERIFIER_error", as, Type::getVoidTy(ctx)));
```
priyasiddharth: Add comment: This generates a fn in the BC that is impl in seahorn (found at linked time)

test/fat_ptr/fat_ptr_outofbounds.01.c
```diff
@@ -0,0 +1,28 @@
+// RUN: %sea clang -m64 -S -o /tmp/bitcode.bc %s > /dev/null 2>&1
```
priyasiddharth: run these instructions by hand

lib/Transforms/Instrumentation/ExtendedPointer.cc
```diff
@@ -0,0 +1,222 @@
+//===- ExtendedPointer.cpp - Instrumentation for run-time bounds checking --===//
```
agurfinkel: Rename into `SeaBufferBoundsCheck`

lib/Transforms/Instrumentation/ExtendedPointer.cc
```diff
@@ -0,0 +1,222 @@
...
+    }
+  }
+  return MadeChange;
+}
+namespace seahorn {
+FunctionPass *createSeaExtendedPointerPass() { return new ExtendedPointer(); }
+}
+
+static RegisterPass<ExtendedPointer>
+    X("extended-pointer-pass", "Bounds checking based on extended pointer");
```
agurfinkel: update name to reflect change to class name. Perhaps `sea-bounds-checking`

tools/seapp/seapp.cc
```diff
@@ -209,6 +209,11 @@ static llvm::cl::opt<bool>
             llvm::cl::desc("Wrap memory accesses with special functions"),
             llvm::cl::init(false));
 
+static llvm::cl::opt<bool>
+    ExtendedPointer("ext-ptr",
```
agurfinkel: rename into `SeaBoundsChecking`, same for option

lib/Transforms/Instrumentation/ExtendedPointer.cc
agurfinkel: `RegisterPass` is probably not needed if the pass is wired with a hard-coded option to `seapp`

lib/Transforms/Instrumentation/ExtendedPointer.cc
agurfinkel: I think using `verifier.error` should be better. `__VERIFIER_error()` is a function defined in SV-COMP. We have a pass to promote it to `verifier.error` that we use internally to ensure that there are no name clashes between our special functions and user defined ones. 

There should be a standard utility to access the standard functions. This is something that is currently missing with raw names being used all over the place. 

@danblitzhou did using `verifier.error` cause a problem? Maybe I am missing something.

lib/Transforms/Instrumentation/ExtendedPointer.cc
```diff
@@ -133,44 +150,99 @@ void ExtendedPointer::emitBranchToTrap(Value *Cmp) {
...
   DEBUG(dbgs() << "Instrument " << *Ptr << " for " << Twine(NeededSize)
                << " bytes\n");
 
   SizeOffsetEvalType SizeOffset = ObjSizeEval->compute(Ptr);
-
+  Value *Or;
   if (!ObjSizeEval->bothKnown(SizeOffset)) {
-    ++ChecksUnable;
-    return false;
+    WARN << "fatptr instrument " << *Ptr << " for " << Twine(NeededSize) << " bytes\n";
```
agurfinkel: protect this print by `LOG`. 

This does not look like a warning message to me, more like a regular log

lib/Transforms/Instrumentation/ExtendedPointer.cc
```diff
@@ -133,44 +150,99 @@ void ExtendedPointer::emitBranchToTrap(Value *Cmp) {
...
+    // Ptr >= Start
+    Value *CmpUnderFlow = Builder->CreateICmpULT(PtrInt, Builder->CreateIntCast(Start, IntTy, true));
+    // Start + Size >= Ptr + NeededSize
+    Value *accessEnd = Builder->CreateAdd(PtrInt, NeededSizeVal);
+    Value *ptrEnd = Builder->CreateAdd(Start, Size);
+    Value *CmpOverFlow = Builder->CreateICmpULT(Builder->CreateIntCast(ptrEnd, IntTy, true), accessEnd);
+    Or = Builder->CreateOr(CmpUnderFlow, CmpOverFlow);
+  } else {
+    // size and offest statically computed
+    WARN << "statically instrument " << *Ptr << " for " << Twine(NeededSize) << " bytes\n";
```
agurfinkel: protect by `LOG`. Not a warning

lib/Transforms/Instrumentation/ExtendedPointer.cc
```diff
@@ -133,44 +150,99 @@ void ExtendedPointer::emitBranchToTrap(Value *Cmp) {
...
                << " bytes\n");
 
   SizeOffsetEvalType SizeOffset = ObjSizeEval->compute(Ptr);
-
+  Value *Or;
   if (!ObjSizeEval->bothKnown(SizeOffset)) {
-    ++ChecksUnable;
-    return false;
+    WARN << "fatptr instrument " << *Ptr << " for " << Twine(NeededSize) << " bytes\n";
+    // get start and end by calling internalized functions
```
agurfinkel: add comment with pseudocode of what code the following instructions generate

lib/Transforms/Instrumentation/ExtendedPointer.cc
```diff
@@ -264,13 +264,13 @@ bool ExtendedPointer::runOnFunction(Function &F) {
```
agurfinkel: the change to this file should be split from this commit and merge into the previous one

lib/Transforms/Instrumentation/ExtendedPointer.cc
```diff
@@ -201,41 +206,70 @@ bool ExtendedPointer::instrument(Value *Ptr, Value *InstVal,
...
-    Builder->CreateCall(
-      m_setFatSlot1,
-      {
-        Builder->CreateBitCast(Ptr, Builder->getInt8PtrTy()),
-        sizeVal
-      }
-    );
+    // Type *IntTy = DL.getIntPtrType(Ptr->getType());
+    // set_fat_slot0(Ptr, Base)
+    Builder->SetInsertPoint(withBase);
```
agurfinkel: lost at what this code is trying to do. need a comment

lib/Transforms/Instrumentation/ExtendedPointer.cc
```diff
@@ -201,41 +206,70 @@ bool ExtendedPointer::instrument(Value *Ptr, Value *InstVal,
...
+    withSize->setArgOperand(1, sizeVal);
   } else {
     // copy_fat_slots(Ptr, BasePtr)
-    Builder->CreateCall(
+    if (auto *GEP = dyn_cast<GetElementPtrInst>(Ptr)) {
+       resultType = GEP->getResultElementType();
+    } else {
+      llvm_unreachable("only handling GEP instructions");
+    }
+    CallInst* copied = Builder->CreateCall(
```
agurfinkel: need comment on this as well. Looks like code is first inserted with placeholders and then updated as needed.
Don't remember why we designed it this way.

lib/Transforms/Instrumentation/ExtendedPointer.cc
```diff
@@ -201,41 +206,70 @@ bool ExtendedPointer::instrument(Value *Ptr, Value *InstVal,
...
       m_copyFatSlots,
       {
-        Builder->CreateBitCast(Ptr, Builder->getInt8PtrTy()),
-        Builder->CreateBitCast(BasePtr, Builder->getInt8PtrTy()),
+        Constant::getNullValue(Builder->getInt8PtrTy()),
+        Constant::getNullValue(Builder->getInt8PtrTy())
       }
     );
+    Value* casted = Builder->CreateBitCast(copied, resultType->getPointerTo());
+    WARN << "casting  " << *Ptr << " to " << *casted << " with type " << *resultType->getPointerTo() << " \n";
```
agurfinkel: protect by LOG, remove WARN

sea-rt/seahorn.cpp
```diff
@@ -166,30 +169,81 @@ void __assert_fail (const char * assertion, const char * file,
 }
 
 
-void __sea_set_extptr_slot0(void* ptr, uint32_t base) {
+void* __sea_set_extptr_slot0_fp(void* ptr, size_t base) {
```
agurfinkel: if this implementation is not working, should comment that it is just a draft. 

Probably a good idea to split `seahorn.cpp` into multiple files at this point

lib/Transforms/Instrumentation/ExtendedPointer.cc
danblitzhou: This would copy size and base of `BasePtr` to `Ptr`. If we use fat ptr encoding implementation, then we need to replace all usage of `Ptr` with a new pointer with copied encoding. I am also kind of lost why we needed to set arguments at the end, but I don't think these steps of replacing all usage of `Ptr` with `copied` would be needed if we just use the hashmap runtime implementation. I can revert this part (`instrumentAddress`) to an older / simpler version if we plan to just use the hashmap implementation for now.

lib/Transforms/Instrumentation/ExtendedPointer.cc
danblitzhou: This is also leftover design when I was working on the fat ptr encoding implementation. Needed to replace all usage of `Ptr` with ones with base and size info encoded  (`withSize`); I can revert this pattern if we are just planning to use hashmap implementation.

lib/Transforms/Instrumentation/ExtendedPointer.cc
danblitzhou: I am not sure if using verifier.error would result in different behaviour or not, you can try swapping

agurfinkel: Using `__VERIFIER_error` is currently preferred since there are no API to get a unifrom declaration of the internal intrinsic `verifier.error()`.

lib/Transforms/Instrumentation/ExtendedPointer.cc
agurfinkel: done

lib/Transforms/Instrumentation/ExtendedPointer.cc
agurfinkel: done

tools/seapp/seapp.cc
agurfinkel: done

lib/Transforms/Instrumentation/ExtendedPointer.cc
agurfinkel: fixed

agurfinkel:
Updated and re-submitted as PR #275

PR #273, opened 2020-03-21
Fatptr symbolic execution

priyasiddharth:
Added logic to process fatptr LLVM instructions.


include/seahorn/Expr/Smt/Z3.hh
```diff
@@ -377,6 +377,13 @@ public:
 
   Expr eval(Expr e, bool completion = false) {
     assert(model);
```
agurfinkel: This commit needs to be split into multiple commits based on functionality added. For instance, updates that fix how structs are evaluated in a model should not be part of the semantics.

lib/seahorn/BvOpSem2.cc
```diff
@@ -500,7 +500,9 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemVisitorBase {
                              f->getName().startswith("verifier.nondet") ||
                              f->getName().startswith("__VERIFIER_nondet")))
         visitNondetCall(CS);
-      else
+      else if (f->getName().startswith("__sea_")) {
```
agurfinkel: This is not specific enough prefix. There might be many functions that we introduce with `__sea_` prefix.

lib/seahorn/BvOpSem2.cc
```diff
@@ -513,6 +515,44 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemVisitorBase {
     llvm_unreachable(nullptr);
   }
 
+  void visitFatPointerInstr(CallSite CS) {
+    const Function *f = getCalledFunction(CS);
+    const Instruction &I = *CS.getInstruction();
+
```
agurfinkel: `assert(f);`

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -12,6 +12,8 @@
 namespace seahorn {
 namespace details {
 
+static const int slotBitWidth = 64;
```
agurfinkel: `unsigned`

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -58,15 +63,21 @@ class FatMemManager : public OpSemMemManager {
   }
   /// \brief Creates a fat memory value from raw memory with default value for
   /// fat
-  FatMemValTy mkFatMem(MemValTy rawMem) const { return strct::mk(rawMem); }
+  FatMemValTy mkFatMem(MemValTy rawMem) const {
+    return strct::mk(rawMem, m_ctx.alu().si(0, slotBitWidth),
```
agurfinkel: type error. Slots of fat memory are memories (i.e., arrays). This works now because shadow memory is not supported.

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -403,6 +416,18 @@ class FatMemManager : public OpSemMemManager {
   FatMemValTy zeroedMemory() const override {
     return mkFatMem(m_mem.zeroedMemory());
   }
+
+  Expr getFatData(PtrTy p, unsigned SlotIdx) override {
+    assert(strct::isStructVal(p));
+    assert(SlotIdx < 2);
```
agurfinkel: would be nice to factor out hard coded constant `2` into a constant just like `slotBitWidth`

lib/seahorn/BvOpSem2FatMemMgr.cc
agurfinkel: naming convention: `g_slotBitWidth` so that it is clear that this is a global value

lib/seahorn/BvOpSem2MemRepr.cc
```diff
@@ -52,6 +52,9 @@ Expr OpSemMemArrayRepr::MemCpy(Expr dPtr, Expr sPtr, unsigned len,
       Expr val = op::array::select(srcMem, sIdx);
       res = op::array::store(res, dIdx, val);
     }
+  } else {
+    ERR << "Word size and pointer are not aligned!";
```
agurfinkel: `ERR` usually entails an assertion following it

lib/seahorn/BvOpSem2MemRepr.cc
```diff
@@ -52,6 +52,9 @@ Expr OpSemMemArrayRepr::MemCpy(Expr dPtr, Expr sPtr, unsigned len,
...
+    llvm_unreachable(nullptr);
```
agurfinkel: We had a habit of using `llvm_unreachable` for `assert(false)`. It ended up a bad idea because of how it is optimized. Let's use `assert(false);` instead

test/fatptr/lit.cfg
```diff
@@ -0,0 +1,67 @@
+# -*- Python -*-
```
agurfinkel: We will now have too many tests with similar names. Xiang's tests are under `fat_ptr`. Need to either unify, or rename to be more distinct.

test/fatptr/lit.cfg
```diff
@@ -0,0 +1,67 @@
...
+
```
agurfinkel: check that you actually need a `lit.cfg`. Usually, the one from parent directory is inherited to all descendents. Only create a new configuration when existing ones do not work.

priyasiddharth: AFAICT: '--horn-bv2-ptr-size=8', '--horn-bv2-fatmem' are two flags which are need for fatptr tests to pass but are not default for other tests. I will let this be for now since whether these flags work for every kind of test is orthogonal to this effort.

test/fatptr/lit.cfg
priyasiddharth: Renamed to fat_ptr_symbolic...can unify later as needed.

lib/seahorn/BvOpSem2MemRepr.cc
priyasiddharth: Ack. Added assert.

lib/seahorn/BvOpSem2MemRepr.cc
priyasiddharth: Ack. Added assert

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: done

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: Ack

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: Done

priyasiddharth: Done

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -73,76 +101,92 @@ class FatMemManager : public OpSemMemManager {
   ~FatMemManager() override = default;
 
   FatPtrTy ptrSort() const override {
-    return sort::structTy(m_mem.ptrSort(), m_ctx.alu().intTy(g_slotBitWidth),
+    return sort::structTy(m_main.ptrSort(), m_ctx.alu().intTy(g_slotBitWidth),
```
agurfinkel, line 104: the bitwidth of the slot has to correspond to the word width of the corresponding memory manager.

That is, `g_slotBitWith` and the word width of the slot memory manager must be the same size.

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -73,76 +101,92 @@ class FatMemManager : public OpSemMemManager {
...
                           m_ctx.alu().intTy(g_slotBitWidth));
   }
 
   /// \brief Allocates memory on the stack and returns a pointer to it
   /// \param align is requested alignment. If 0, default alignment is used
   FatPtrTy salloc(unsigned bytes, uint32_t align = 0) override {
-    return mkFatPtr(m_mem.salloc(bytes, align));
+    return mkFatPtr(m_main.salloc(bytes, align), m_slot0.salloc(bytes, align),
```
agurfinkel, line 111: Seems wrong. Probaby should be
```
auto ptrV = m_mem.salloc(bytes, align);
m_slot0.salloc(bytes, align); /* to let slot0manager know about allocation */
m_slot1.salloc(bytes, align);
return mkFatPtr(ptrV);
```

I am assuming that `mkFatPtr()` converts a raw pointer into fat with default values (probably 0) for the slots.

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -73,76 +101,92 @@ class FatMemManager : public OpSemMemManager {
...
+                    m_slot1.salloc(bytes, align));
   }
 
   /// \brief Allocates memory on the stack and returns a pointer to it
   FatPtrTy salloc(Expr elmts, unsigned typeSz, uint32_t align = 0) override {
-    return mkFatPtr(m_mem.salloc(elmts, typeSz, align));
+    return mkFatPtr(m_main.salloc(elmts, typeSz, align),
+                    m_slot0.salloc(elmts, typeSz, align),
```
agurfinkel, line 118: see comment on the other `salloc`

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -73,76 +101,92 @@ class FatMemManager : public OpSemMemManager {
...
+                    m_slot1.salloc(elmts, typeSz, align));
   }
 
   /// \brief Returns a pointer value for a given stack allocation
   FatPtrTy mkStackPtr(unsigned offset) override {
-    return mkFatPtr(m_mem.mkStackPtr(offset));
+    return mkFatPtr(m_main.mkStackPtr(offset), m_slot0.mkStackPtr(offset),
```
agurfinkel, line 124: see comment on `salloc`

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -73,76 +101,92 @@ class FatMemManager : public OpSemMemManager {
...
+                    m_slot1.mkStackPtr(offset));
   }
 
   /// \brief Pointer to start of the heap
-  FatPtrTy brk0Ptr() override { return mkFatPtr(m_mem.brk0Ptr()); }
+  FatPtrTy brk0Ptr() override {
+    return mkFatPtr(m_main.brk0Ptr(), m_slot0.brk0Ptr(), m_slot1.brk0Ptr());
```
agurfinkel, line 130: see comment on `salloc`

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -73,76 +101,92 @@ class FatMemManager : public OpSemMemManager {
...
+  }
 
   /// \brief Allocates memory on the heap and returns a pointer to it
   FatPtrTy halloc(unsigned _bytes, uint32_t align = 0) override {
-    return mkFatPtr(m_mem.halloc(_bytes, align));
+    return mkFatPtr(m_main.halloc(_bytes, align), m_slot0.halloc(_bytes, align),
```
agurfinkel, line 135: see `salloc`

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -73,76 +101,92 @@ class FatMemManager : public OpSemMemManager {
...
+                    m_slot1.halloc(_bytes, align));
   }
 
   /// \brief Allocates memory on the heap and returns pointer to it
   FatPtrTy halloc(Expr bytes, uint32_t align = 0) override {
-    return mkFatPtr(m_mem.halloc(bytes, align));
+    return mkFatPtr(m_main.halloc(bytes, align), m_slot0.halloc(bytes, align),
```
agurfinkel, line 141: see `salloc`

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -73,76 +101,92 @@ class FatMemManager : public OpSemMemManager {
...
+                    m_slot1.halloc(bytes, align));
   }
 
   /// \brief Allocates memory in global (data/bss) segment for given global
   FatPtrTy galloc(const GlobalVariable &gv, uint32_t align = 0) override {
-    return mkFatPtr(m_mem.galloc(gv, align));
+    return mkFatPtr(m_main.galloc(gv, align), m_slot0.galloc(gv, align),
```
agurfinkel, line 147: see `salloc`

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -73,76 +101,92 @@ class FatMemManager : public OpSemMemManager {
...
+                    m_slot1.galloc(gv, align));
   }
 
   /// \brief Allocates memory in code segment for the code of a given
   /// function
   FatPtrTy falloc(const Function &fn) override {
-    return mkFatPtr(m_mem.falloc(fn));
+    return mkFatPtr(m_main.falloc(fn), m_slot0.falloc(fn), m_slot1.falloc(fn));
```
agurfinkel, line 154: see `salloc`

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -73,76 +101,92 @@ class FatMemManager : public OpSemMemManager {
...
   }
 
   /// \brief Returns a function pointer value for a given function
   FatPtrTy getPtrToFunction(const Function &F) override {
-    return mkFatPtr(m_mem.getPtrToFunction(F));
+    return mkFatPtr(m_main.getPtrToFunction(F), m_slot0.getPtrToFunction(F),
```
agurfinkel, line 159: see `salloc`

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -73,76 +101,92 @@ class FatMemManager : public OpSemMemManager {
...
+                    m_slot1.getPtrToFunction(F));
   }
 
   /// \brief Returns a pointer to a global variable
   FatPtrTy getPtrToGlobalVariable(const GlobalVariable &gv) override {
-    return mkFatPtr(getPtrToGlobalVariable(gv));
+    return mkFatPtr(m_main.getPtrToGlobalVariable(gv),
```
agurfinkel, line 165: see `salloc`

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -73,76 +101,92 @@ class FatMemManager : public OpSemMemManager {
...
+    m_slot1.initGlobalVariable(gv);
   }
 
   /// \brief Creates a non-deterministic pointer that is aligned
   ///
   /// Top bits of the pointer are named by \p name and last \c log2(align)
   /// bits are set to zero
   FatPtrTy mkAlignedPtr(Expr name, uint32_t align) const {
-    return mkFatPtr(m_mem.mkAlignedPtr(name, align));
+    return mkFatPtr(m_main.mkAlignedPtr(name, align),
```
agurfinkel, line 182: see comment on `salloc`

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -155,11 +199,13 @@ class FatMemManager : public OpSemMemManager {
...
   Expr mkMemRegisterSort(const Instruction &inst) const {
-    return sort::structTy(m_mem.mkMemRegisterSort(inst));
+    return sort::structTy(m_main.mkMemRegisterSort(inst),
+                          m_slot0.mkMemRegisterSort(inst),
+                          m_slot1.mkMemRegisterSort(inst));
   }
 
   /// \brief Returns a fresh aligned pointer value
-  FatPtrTy freshPtr() override { return mkFatPtr(m_mem.freshPtr()); }
+  FatPtrTy freshPtr() override { return mkFatPtr(m_main.freshPtr()); }
```
agurfinkel, line 208: this is correct and a good example for all other functions that must return a **fat** pointer, but do not have particular value for the **fat**

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -193,20 +239,28 @@ class FatMemManager : public OpSemMemManager {
...
                           uint64_t align) override {
-    return mkFatPtr(
-        m_mem.loadPtrFromMem(mkRawPtr(ptr), mkRawMem(mem), byteSz, align));
+    MemValTy rawVal =
+        m_main.loadPtrFromMem(mkRawPtr(ptr), mkRawMem(mem), byteSz, align);
+    MemValTy slot0Val =
+        m_slot0.loadIntFromMem(mkRawPtr(ptr), mkSlot0Mem(mem), byteSz, align);
+    MemValTy slot1Val =
+        m_slot1.loadIntFromMem(mkRawPtr(ptr), mkSlot1Mem(mem), byteSz, align);
+    LOG("opsem", WARN << "ptr:" << rawVal << " slot0:" << slot0Val
```
agurfinkel, line 255: `WARN` is reserved for warnings. Please do not use it for debug. Use `llvm::errs()` for debug instead.

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -193,20 +239,28 @@ class FatMemManager : public OpSemMemManager {
...
+                      << " slot1:" << slot1Val << "\n");
+    LOG("opsem", WARN << "FatMem:" << mem);
+    return mkFatPtr(rawVal, slot0Val, slot1Val);
   }
 
   /// \brief Pointer addition with numeric offset
   FatPtrTy ptrAdd(FatPtrTy ptr, int32_t _offset) const {
-    PtrTy rawPtr = m_mem.ptrAdd(mkRawPtr(ptr), _offset);
+    PtrTy rawPtr = m_main.ptrAdd(mkRawPtr(ptr), _offset);
```
agurfinkel, line 263: This is an important difference from what Xiang will usually generate. We are changing gep to copy fat while in his case a gep will follow by a copy of fat. Not yet convinced what is the best course of action. But this does not affect current tests. Just something to note in code.

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -224,15 +278,27 @@ class FatMemManager : public OpSemMemManager {
   FatMemValTy storeIntToMem(Expr _val, FatPtrTy ptr, FatMemValTy mem,
                             unsigned byteSz, uint64_t align) override {
     return mkFatMem(
-        m_mem.storeIntToMem(_val, mkRawPtr(ptr), mkRawMem(mem), byteSz, align));
+        m_main.storeIntToMem(_val, mkRawPtr(ptr), mkRawMem(mem), byteSz, align),
+        m_main.storeIntToMem(0, mkRawPtr(ptr), mkSlot0Mem(mem), byteSz, align),
```
agurfinkel, line 282: this is not needed, or is optional, I don't know whether it is cheaper to clear unused slots or let them keep their old value.  At any rate, the number of bytes being written in the slots in definitely not `byteSz` and `align` should not matter (i.e., the store is always aligned and always one word).

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -224,15 +278,27 @@ class FatMemManager : public OpSemMemManager {
...
   /// \brief Stores a pointer into memory
   /// \sa storeIntToMem
   FatMemValTy storePtrToMem(FatPtrTy val, FatPtrTy ptr, FatMemValTy mem,
                             unsigned byteSz, uint64_t align) override {
-    return mkFatMem(m_mem.storePtrToMem(mkRawPtr(val), mkRawPtr(ptr),
-                                        mkRawMem(mem), byteSz, align));
+    // assert that ptr is FatPtrTy
+    MemValTy main = m_main.storePtrToMem(mkRawPtr(val), mkRawPtr(ptr),
+                                         mkRawMem(mem), byteSz, align);
+    MemValTy slot0 = m_slot0.storeIntToMem(getFatData(val, 0), mkRawPtr(ptr),
```
agurfinkel, line 293: not storing `byteSz`. See comment in `storeIntToMem`

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -224,15 +278,27 @@ class FatMemManager : public OpSemMemManager {
...
+                                           mkSlot0Mem(mem), byteSz, align);
+    MemValTy slot1 = m_slot1.storeIntToMem(getFatData(val, 1), mkRawPtr(ptr),
+                                           mkSlot1Mem(mem), byteSz, align);
+    LOG("opsem", WARN << "ptr:" << main << " slot0:" << slot0
```
agurfinkel, line 297: Reserve `WARN` for warnings

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -323,96 +389,109 @@ class FatMemManager : public OpSemMemManager {
   FatMemValTy MemSet(PtrTy ptr, Expr _val, unsigned len, FatMemValTy mem,
                      uint32_t align) override {
     return mkFatMem(
-        m_mem.MemSet(mkRawPtr(ptr), _val, len, mkRawMem(mem), align));
+        m_main.MemSet(mkRawPtr(ptr), _val, len, mkRawMem(mem), align),
+        m_main.MemSet(mkRawPtr(ptr), _val, len, mkSlot0Mem(mem), align),
```
agurfinkel, line 393: Multiple problems:
1. `m_main` is not the right manager for slot0 and slot1
2. `_val` is not the right value to be written to the slot

Same as in `storeIntToMem`. Might be easier to leave unused slots uninitialized.

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -323,96 +389,109 @@ class FatMemManager : public OpSemMemManager {
...
+        m_main.MemSet(mkRawPtr(ptr), _val, len, mkSlot1Mem(mem), align));
   }
 
   /// \brief Executes symbolic memcpy with concrete length
   FatMemValTy MemCpy(PtrTy dPtr, PtrTy sPtr, unsigned len, Expr memTrsfrRead,
                      uint32_t align) override {
-    return mkFatMem(m_mem.MemCpy(mkRawPtr(dPtr), mkRawPtr(sPtr), len,
-                                 mkRawMem(memTrsfrRead), align));
+    return mkFatMem(m_main.MemCpy(mkRawPtr(dPtr), mkRawPtr(sPtr), len,
```
agurfinkel, line 400: I think is is ok. Would be good to document the intended semantics. Some examples with `MemCpy` will help narrow this down.

I think we now assume that every byte of raw memory has slot0 and slot1, which themselves are usually wider. `memcpy` copies bytes and underlying slots. 

Seems like good semantics.

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -323,96 +389,109 @@ class FatMemManager : public OpSemMemManager {
...
   }
 
   /// \brief Executes symbolic memcpy from physical memory with concrete
   /// length
   FatMemValTy MemFill(PtrTy dPtr, char *sPtr, unsigned len, FatMemValTy mem,
                       uint32_t align = 0) override {
     return mkFatMem(
-        m_mem.MemFill(mkRawPtr(dPtr), sPtr, len, mkRawMem(mem), align));
+        m_main.MemFill(mkRawPtr(dPtr), sPtr, len, mkRawMem(mem), align),
+        m_main.MemFill(mkRawPtr(dPtr), sPtr, len, mkSlot0Mem(mem), align),
```
agurfinkel, line 414: same comments as `memset`. Don't need to initialize slots, but if we do, we need to initialize them properly, not with the same values as raw memory

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -323,96 +389,109 @@ class FatMemManager : public OpSemMemManager {
...
+        m_main.MemFill(mkRawPtr(dPtr), sPtr, len, mkSlot1Mem(mem), align));
   }
 
   /// \brief Executes inttoptr conversion
   FatPtrTy inttoptr(Expr intVal, const Type &intTy, const Type &ptrTy) const {
-    return mkFatPtr(m_mem.inttoptr(intVal, intTy, ptrTy));
+    return mkFatPtr(m_main.inttoptr(intVal, intTy, ptrTy),
```
agurfinkel, line 420: I assume that `inttoptr` sets slots to default values. Now they are set to the value of the integer. This will typecheck now, but might not in the future when slots are of different width than pointer size.

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -431,9 +510,11 @@ class FatMemManager : public OpSemMemManager {
 FatMemManager::FatMemManager(Bv2OpSem &sem, Bv2OpSemContext &ctx,
                              unsigned ptrSz, unsigned wordSz, bool useLambdas)
     : OpSemMemManager(sem, ctx, ptrSz, wordSz),
-      m_mem(sem, ctx, ptrSz, wordSz, useLambdas) {
+      m_main(sem, ctx, ptrSz, wordSz, useLambdas),
+      m_slot0(sem, ctx, ptrSz, wordSz, useLambdas),
```
agurfinkel, line 514: word size of `slot0` and `slot1` are almost definitely not the same as word size of `m_main`

lib/seahorn/BvOpSem2RawMemMgr.cc
```diff
@@ -351,10 +352,14 @@ PtrTy RawMemManager::ptrAdd(PtrTy ptr, Expr offset) const {
...
-///
+/// \param[in] ptr is the address at which _val will be stored
+/// \param[in] _val is the value being written
+/// \param[in] mem is the memory bank/register being written to
+/// \param[in] ty is the type of value being stored
+/// \param[in] align is the known alignment of the ptr
 /// Returns an expression describing the state of memory in \c memReadReg
 /// after the store
-/// \sa loadIntFromMem
+/// \sa storeIntFromMem
```
agurfinkel, line 362: this is wrong. This is already `storeIntToMem`. The additional function to **see also** is `loadIntFromMem`

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: Ack

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: Acl

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: ack

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: ack

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: ack

lib/seahorn/BvOpSem2RawMemMgr.cc
priyasiddharth: fixed typo

priyasiddharth:
Fixed implementation accounting for comments so far. Xiang's tests pass as expected.

include/seahorn/Expr/Smt/Z3.hh
agurfinkel: This comment still remains. Git history needs to be reworked. I can help if needed.

lib/seahorn/BvOpSem2.cc
agurfinkel: looks like this is still a todo

lib/seahorn/BvOpSem2.cc
agurfinkel: still todo

lib/seahorn/BvOpSem2FatMemMgr.cc
agurfinkel: still todo

lib/seahorn/BvOpSem2FatMemMgr.cc
agurfinkel: todo

agurfinkel:
I still see many old todo items, but also getting lost in too many comments. 
Would be good to address them, fix history, and submit as new PR that can be reviewed again

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -279,8 +283,10 @@ class FatMemManager : public OpSemMemManager {
                             unsigned byteSz, uint64_t align) override {
     return mkFatMem(
         m_main.storeIntToMem(_val, mkRawPtr(ptr), mkRawMem(mem), byteSz, align),
-        m_main.storeIntToMem(0, mkRawPtr(ptr), mkSlot0Mem(mem), byteSz, align),
-        m_main.storeIntToMem(0, mkRawPtr(ptr), mkSlot1Mem(mem), byteSz, align));
+        m_slot0.storeIntToMem(m_ctx.alu().si(0, g_slotBitWidth), mkRawPtr(ptr),
+                              mkSlot0Mem(mem), g_slotByteWidth, align),
```
agurfinkel, line 287: I suggest to replace `m_slot0.storeIntToMem...` with `mkSlot0Mem(mem)`.

When writing a thin integer into fat memory we have a choice of what to do with the shadow. 
Zero it out, initialize non-deterministically, or leave it unchanged. I think leaving it unchanged
is best from performance perspective. So that would be my first choice. We can turn it into a
command line option if needed.

include/seahorn/Expr/Smt/Z3.hh
priyasiddharth: I will break it off before final review (after LLVM 10 merge).

lib/seahorn/BvOpSem2.cc
priyasiddharth: I don't think this logic precludes adding new __sea_functions. We will need to add a sibling to visitFatPtrInstr(...) for each function. Overall, I remember we wanted to rename functions anyway. My preference is to make it
`__sea_extptr_[get|set|copy|recover] `
once merged with Xiang's implementation. This will also make the if condition tighter.

lib/seahorn/BvOpSem2.cc
priyasiddharth: I don't recall the context here. But I assume we just want to make sure f is not null.

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: For 1, you may be looking at an outdated version. I have already fixed it as follows

 ```
   return mkFatMem(
        m_main.MemSet(mkRawPtr(ptr), _val, len, mkRawMem(mem), align),
        m_slot0.MemSet(mkRawPtr(ptr), m_ctx.alu().si(0, g_slotBitWidth), len,
                       mkSlot0Mem(mem), align),
        m_slot1.MemSet(mkRawPtr(ptr), m_ctx.alu().si(0, g_slotBitWidth), len,
                       mkSlot1Mem(mem), align));
```
2. I don't think`_val` is wrong since slot values are don't care?
3. Ack on not initializing.

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: Fixed

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: Ack.

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: done

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: done

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: ack

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: ack

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: noted in code.

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: ack. let the old value of slot0 and slot1 persist.

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: Ack

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: Ack. Fixed

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: done

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: done

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: I am not seeing why `byteSz` is wrong.

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: memfill funcdoc says it's some kind of memcpy so I copy slot0 and slot1 data also. I don't see why we can let them be uninitialized.

priyasiddharth:
I have responded to all comments AFAICT. 2-3 comments are unclear so request additional clarification - left them unresolved for now.

lib/seahorn/BvOpSem2.cc
agurfinkel: I mean that you currently treat all functions with prefix `__sea_` as fat ptr related. But they don't have to be. `__sea_exptr` is a better, more specific prefix.

lib/seahorn/BvOpSem2FatMemMgr.cc
agurfinkel: `MemFill` takes a concrete memory region and copies it into a symbolic region. The concrete memory region does not have fat part.

lib/seahorn/BvOpSem2.cc
agurfinkel: Yes. `getCalledFunction` returns null if the call is indirect and cannot be resolved to a known function.

lib/seahorn/BvOpSem2FatMemMgr.cc
agurfinkel: yes, old code. the new is good. But would forgo initialization right now.

lib/seahorn/BvOpSem2FatMemMgr.cc
priyasiddharth: Done

lib/seahorn/BvOpSem2.cc
priyasiddharth: Ok will make the change after rebase with HEAD.

agurfinkel:
is this branch dead? can I close it?

priyasiddharth:
This has been merged already in https://github.com/seahorn/seahorn/pull/280. Closing this PR.

PR #281, opened 2020-05-12
feat: Client side pre commit checks

priyasiddharth:
1. Please look at `.conform.yaml` for checks added. 
2. The checks are benign until the pre-commit tool is enabled for a local repo so merging this and beta testing for a couple of days can help.
3. The same checks will be used in a github action (PR to follow once this is submitted and tested in daily workflows).


.pre-commit-config.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v2.4.0
+  hooks:
+  # This hook trims trailing whitespace (active check).
+  - id: trailing-whitespace
+  # Ensures that a file is either empty, or ends with one newline (active check).
+  - id: end-of-file-fixer
+
+- repo: https://gitlab.com/daverona/pre-commit-cpp
```
agurfinkel, line 25: this does not look right. The goal is to hook commit messages, this also runs clang-format, and ignoring our custom configuration in `.clang-format`.

We already have an action for clang-format on PRs. This should be enough.

CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+sudo apt-get install clang-format
+```
+
+Enable precommit inside root of local git repo
+
+```
+>pre-commit install
+```
+
+If you don't have it installed, you can install it using the [getgo tool](https://github.com/golang/tools/tree/master/cmd/getgo).
```
agurfinkel, line 34: what is **it** ? What is `getgo`. Do I need to install it as well?

agurfinkel: running `pre-commit` fails on my mac

agurfinkel:
https://magit.vc/manual/magit/My-Git-hooks-work-on-the-command_002dline-but-not-inside-Magit.html

Is this going to be a problem?

Instead of installing a lot of things and running on my local configurations, I would rather have the check ran as action on PRs. Often, when working any check like that would be in the way. However, when cleaning the work up for merging to master, it would be very welcome.

agurfinkel:
For some reason, this also includes a number of irrelevant commits, so I'm closing it until it is rebased

PR #282, opened 2020-05-12
feat: add conventional commit check github action and client config

priyasiddharth:
1. Add github action and conform configuration
2. Add pre-commit config file (for local validation, if desired)
3. Add documentation for enabling local validation if needed

Comments from closed pr#281

...
https://magit.vc/manual/magit/My-Git-hooks-work-on-the-command_002dline-but-not-inside-Magit.html

Is this going to be a problem?

Instead of installing a lot of things and running on my local configurations, I would rather have the check ran as action on PRs. Often, when working any check like that would be in the way. However, when cleaning the work up for merging to master, it would be very welcome.
...
>The local check is only done for people who want it and install conform using pre-commit hooks. The github action is only executed on push and pull_request and does not depend on client configuration. I don't have a solution for the custom workflow using magit but then you may choose not to use local check at all. I imagine some people may find the local check useful than figuring that a push failed because of a commit message at the last minute. For these users, instructions in the guide and distributing the `.pre-commit` config file is all that is needed.

...
this does not look right. The goal is to hook commit messages, this also runs clang-format, and ignoring our custom configuration in .clang-format._
...
>I have removed clang-format check for now to keep the change simple. I would like the option of automatic clang format from cmdline on commit for similar reason as previous comment.

Thanks

### DIFF
--- a/.conform.yaml
+++ b/.conform.yaml
@@ -1,0 +1,31 @@
+# Enabling a subset of options
+# Full list at
+#  https://github.com/talos-systems/conform/blob/master/README.md
+
+policies:
+  - type: commit
+    spec:
+      header:
+        length: 80
+        imperative: false
+        case: lower
+        invalidLastCharacters: .
+      body:
+        required: false
+      # Allow maximum of one commit ahead of master
+      # TODO: figure out intent, false for now
+      maximumOfOneCommit: false
+      conventional:
+        types:
+          - fix
+          - feat
+          - build
+          - ci
+          - docs
+          - style
+          - refactor
+          - perf
+          - test
+        scopes:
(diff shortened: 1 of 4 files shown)


agurfinkel:
I don't like clang-format on commit because sometimes formatting does affect compilation. Sometimes it affects readability. You should clang format in the IDE and place directive to selectively disable formatting as needed. On the other hand, a check that committed code is formatted (i.e., failure if it is not) is a good thing to have.

Since this does not affect people that do not install the hooks (like me :) ) there is no issues in including it. 

Will you add actions to check messages on PRs? Just like what we have now to check for formatting?

PR #287, opened 2020-05-13
bug(z3_api)

igcontreras:


units/units_z3_api.cpp
```diff
@@ -0,0 +1,100 @@
...
+  query = mk<AND>(mk<NEQ>(x, solution), fapp);
+
+  // Register new relation to query without variables
+  ExprVector qtype;
+  qtype.push_back(bTy);
+  Expr query_name = mkTerm<string>("query1", efac);
+  Expr qdecl = bind::fdecl(query_name, qtype);
+
+  // this line makes it crash, try uncommenting it
+  // qdecl = bind::boolConst(query_name);
```
igcontreras, line 73: Uncommenting this line causes it to crash

PR #291, opened 2020-05-13
Basic type checker

colinxm:
Extended the type checker skeleton to booleans, as stated in the Type Checker project


agurfinkel:
`clang-format`, rebase against `dev10`, there should be no merge commits in your PR.
Follow https://www.conventionalcommits.org/en/v1.0.0/

include/seahorn/Expr/ExprOpBind.hh
```diff
@@ -184,6 +184,11 @@ inline Expr typeOf(Expr v) {
   //     return array_ty;
   //   }
   // }
+  
+  if (isOpX <AND> (v) || isOpX <OR> (v) || isOpX <XOR> (v) ||  isOpX <IMPL> (v)
```
agurfinkel, line 188: You should not change this file

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -12,10 +12,64 @@ namespace {
...
+    LOG("tc", llvm::errs() << "post visiting expression: " << *exp << "\n";); 
+
+    if (!checkChildren(exp)) {
+      LOG("tc", llvm::errs() << "improper types at:" << *exp  << "\n";);
+
+      expTypeStack.push_back(Expr());
+      return exp;
+    }
+
+      expTypeStack.push_back(bind::typeOf(exp));
```
agurfinkel, line 71: The stack will not work. Expressions are a DAG not a tree.
I'd want to see a more complete and documented design of the traversal your are planning to do.
This is kind of a mess right now.

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -12,64 +12,72 @@ namespace {
 
 //==-- main implementation goes here --==//
 class TCVR {
-  ExprVector expTypeStack;
-
-  bool isConst (Expr exp) {
-   return bind::isBoolConst(exp) || bind::isIntConst(exp);
+  ExprMap exprMap;
```
agurfinkel, line 15: filed names should start with `m_`. A good name would be `m_cache`.
https://github.com/seahorn/seahorn/wiki/Developer-Corner

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -12,64 +12,72 @@ namespace {
...
+
+  bool isBool(Expr exp) {
```
agurfinkel, line 17: this is good in principle, but you want the code to be parametric in types. So you want it to return the type, rather than assume that you know the type names

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -12,64 +12,72 @@ namespace {
...
+
+    return false;
   }
 
-  bool isVar (Expr exp) {
-   return bind::isBoolVar(exp) || bind::isIntVar(exp);
-  }
+  Expr inferType(Expr exp /*,TypeChecker & ty*/) {
+    if (isBool(exp))
+      return sort::boolTy(exp->efac());
```
agurfinkel, line 32: you also want to check that the expression is well formed -- i.e., children are of the right type and the right number of children. For example, is `NOT (A, B)` allowed? is `AND A` allowed? `AND` with no operands?

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -12,64 +12,72 @@ namespace {
...
-      Expr a = expTypeStack.back(); 
-      expTypeStack.pop_back();
+  // returns true if children are of correct type
+  bool checkChildren(Expr exp) {
 
-      bool properTypes = (a == sort::boolTy (exp->efac())) && (b == c);
-      return properTypes;
+    if (isOpX<ITE>(exp)) {
+      // ite (a,b,c) : a is bool, b and c same type
+      bool aIsBool = exprMap[exp->arg(0)] == sort::boolTy(exp->efac());
```
agurfinkel, line 52: wrap calls to `exprMap` with a getter function. Cache `boolTy`, don't use `operator[]`, use `.at()` instead.

agurfinkel: use C++ algorithms to check that all arguments are of the same type: https://en.cppreference.com/w/cpp/algorithm/all_any_none_of

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -12,64 +12,72 @@ namespace {
...
+      bool bCSame = exprMap[exp->arg(1)] == exprMap[exp->arg(2)];
+      return aIsBool && bCSame;
     }
-    //default: check that chilren expressions are of the same types as the parent
 
-    int numChildren = exp->arity();
-    Expr parentType = bind::typeOf(exp);
+    // default: check that chilren expressions are of the same types as the
```
agurfinkel, line 57: **children**

lib/seahorn/Smt/TypeChecker.cc
colinxm: Is it fine to return Expr() when it is not well formed or would you recommend introducing a new error type?

agurfinkel: I would not abuse `Expr()` too much. It is basically the same as `nullptr`. Having it all over the place will cause all sort of issues if null is not expected. For example, you would not be able to easily print type of all expressions to discover at which point type error was introduced.

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -12,10 +12,92 @@ namespace {
 
 //==-- main implementation goes here --==//
 class TCVR {
+  ExprMap m_cache;
+
+  bool isConst(Expr exp) {
+    return bind::isBoolConst(exp) || bind::isIntConst(exp);
+  }
+
+  bool isVar(Expr exp) { return bind::isBoolVar(exp) || bind::isIntVar(exp); }
```
agurfinkel, line 21: I think vars are deprecated. Don't worry about them. We have constants and bound variables, but no variables.

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -12,10 +12,92 @@ namespace {
...
+      return boolSort;
+    else
+      return Expr();
+  }
+
+  // ensures the expression has the correct number of children and all children
+  // are bool types
+  Expr boolCheckChildren(Expr exp, bool (*checkNumChildren)(int)) {
+    auto isBool = [this](Expr exp) {
+      return this->m_cache.at(exp) == sort::boolTy(exp->efac());
```
agurfinkel, line 40: don't create so many expressions. use `isOp<BOOL_TY>()` instead. Cache types in type checker if needed.

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -12,10 +12,92 @@ namespace {
...
+    if (exp->arity() == 3 && m_cache.at(exp->arg(0)) == boolSort &&
+        (m_cache.at(exp->arg(1)) == m_cache.at(exp->arg(2))))
+      return boolSort;
+    else
+      return Expr();
+  }
+
+  // ensures the expression has the correct number of children and all children
+  // are bool types
+  Expr boolCheckChildren(Expr exp, bool (*checkNumChildren)(int)) {
```
agurfinkel, line 38: use `std::function` to pass function objects

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -12,10 +12,92 @@ namespace {
...
+
+    if (checkNumChildren(exp->arity()) &&
+        std::all_of(exp->args_begin(), exp->args_end(), isBool))
+      return sort::boolTy(exp->efac());
+    else
+      return Expr();
+  }
+
+  Expr inferTypeBoolBinary(Expr exp) {
+    auto checkNumChildren = [](int numChildren) -> bool {
```
agurfinkel, line 51: use `unsigned` instead of `int`. Number of children cannot be negative

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -12,10 +12,92 @@ namespace {
...
+      return numChildren == 2;
+    };
+    return boolCheckChildren(exp, checkNumChildren);
+  }
+
+  Expr inferTypeBoolUnary(Expr exp) {
+    auto checkNumChildren = [](int numChildren) -> bool {
```
agurfinkel, line 58: For small numbers, probably more fun to use templates.
```
template<N> bool num_kids (Expr e) { return e->arity() == N; }
```
then use `num_kids<2>(e)`

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -12,10 +12,92 @@ namespace {
...
+  }
 
   /// Called after children have been visited
   Expr postVisit(Expr exp) {
     LOG("tc", llvm::errs() << "post visiting expression: " << *exp << "\n";);
+
+    Expr type = inferType(exp);
+
+    m_cache.insert({exp, type});
+    if (!type)
```
agurfinkel, line 97: need a good way to bailout. You can't just terminate the visitor, but you can ensure that pre-visit returns false from this point on. 

Also need a strategy to remember what was the failure and a way to report this to clients.

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -62,8 +146,8 @@ class TypeChecker::Impl {
 public:
   Expr typeOf(Expr e) {
     TCV _visitor;
-    visit(_visitor, e);
-    return _visitor.knownTypeOf(e);
+    Expr v = dagVisit(_visitor, e);
```
agurfinkel, line 149: probably `visit` is just as good since your type-checker maintains the cache already. 
This way everything is cached twice.

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -12,27 +12,181 @@ namespace {
...
+
+  Expr inferTypeBoolUnary(Expr exp) {
+    std::function<bool(Expr)> checkNumChildren =
+        std::bind(&TCVR::numOfChildrenEq<1>, this, std::placeholders::_1);
+    return boolCheckChildren(exp, checkNumChildren);
+  }
+
+  Expr inferTypeBoolNary(Expr exp) {
+    std::function<bool(Expr)> checkNumChildren = std::bind(
+        &TCVR::numOfChildrenGreaterEq<2>, this, std::placeholders::_1);
```
colinxm, line 112: @agurfinkel is this what you meant when you recommended to use templates instead of lambdas?

agurfinkel: Something got lost in translation. Don't use `std::bind` if you can. Use lambdas.
You had function pointers before (I think), and I was suggesting to use `std::function` for them instead.

colinxm: oh ok I must have misread something

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -12,27 +12,181 @@ namespace {
...
+
+  // ensures the expression has the correct number of children and all children
+  // are the same and correct type
+  Expr numTypeCheckChildren(Expr exp,
+                            std::function<bool(Expr exp)> checkNumChildren) {
+    if (!checkNumChildren(exp))
+      return Expr();
+
+    Expr type = m_cache.at(exp->first());
+    auto isSameType = [this, type](Expr exp) {
```
agurfinkel, line 49: should this by `&this`?

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -12,27 +12,181 @@ namespace {
...
+      return sort::boolTy(exp->efac());
+    else
+      return Expr();
+  }
+
+  // ensures the expression has the correct number of children and all children
+  // are bool types
+  Expr boolCheckChildren(Expr exp,
+                         std::function<bool(Expr exp)> checkNumChildren) {
+    auto isBool = [this](Expr exp) {
```
agurfinkel, line 85: `&this`?

lib/seahorn/Smt/TypeChecker.cc
colinxm: It seems like you can't capture this by reference https://stackoverflow.com/questions/16323032/why-cant-i-capture-this-by-reference-this-in-lambda

agurfinkel: Of course. `this` is a pointer , my bad :), but `type` can be captured by reference

include/seahorn/Expr/ExprOpCore.hh
```diff
@@ -411,8 +418,13 @@ struct DefOp : public B {
   };
 
 /// macro for defining a class for a single operator
-#define NOP(NAME, TEXT, STYLE, BASE)                                           \
+#define NOP_TYPECHECK(NAME, TEXT, STYLE, BASE, TYPECHECK)                      \
```
colinxm, line 421: So just to make it clear, at the moment I'm planning to use NOP_TYPECHECK temporarily until I can change every NOP just so nothing breaks

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -50,25 +95,37 @@ class TCV : public std::unary_function<Expr, VisitAction> {
...
 };
 
-TypeChecker::TypeChecker() : m_impl(new TypeChecker::Impl()) {}
+TypeChecker::TypeChecker() : m_impl(new TypeChecker::Impl(this)) {}
 TypeChecker::~TypeChecker() { delete m_impl; }
 Expr TypeChecker::typeOf(Expr e) { return m_impl->typeOf(e); }
+Expr TypeChecker::getErrorExp() { // to be called after typeOf() or sortOf()
+  return m_impl->getErrorExp();
+}
+void TypeChecker::reset() {
```
colinxm, line 126: Is this an alright way to go about this? The problem that I had was that there isn't a way to know when the user is done with the expression so stuff either has to be reset by the user (like im doing here) or the user has to use a new instance of TypeChecker for a new expression

colinxm: never mind. I'll just store the topmost expression and reset when post-visiting that expression

include/seahorn/Expr/ExprOpBv.hh
```diff
@@ -165,59 +179,188 @@ enum class BvOpKind {
...
+NOP_TYPECHECK(BSUB, "bvsub", FUNCTIONAL, BvOp, typeCheck::bvType::Nary)
+NOP_TYPECHECK(BMUL, "bvmul", FUNCTIONAL, BvOp, typeCheck::bvType::Nary)
+NOP_TYPECHECK(BUDIV, "bvudiv", FUNCTIONAL, BvOp, typeCheck::bvType::Nary)
+NOP_TYPECHECK(BSDIV, "bvsdiv", FUNCTIONAL, BvOp, typeCheck::bvType::Nary)
+NOP_TYPECHECK(BUREM, "bvurem", FUNCTIONAL, BvOp, typeCheck::bvType::Nary)
+NOP_TYPECHECK(BSREM, "bvsrem", FUNCTIONAL, BvOp, typeCheck::bvType::Nary)
+NOP_TYPECHECK(BSMOD, "bvsmod", FUNCTIONAL, BvOp, typeCheck::bvType::Nary)
+NOP_TYPECHECK(BULT, "bvult", FUNCTIONAL, BvOp, typeCheck::bvType::Nary)
+NOP_TYPECHECK(BSLT, "bvslt", FUNCTIONAL, BvOp, typeCheck::bvType::Nary)
+NOP_TYPECHECK(BULE, "bvule", FUNCTIONAL, BvOp, typeCheck::bvType::Nary)
```
agurfinkel, line 319: picking this one at random, but most of these functions are binary. For example, `bvule` is bit-vector unsigned less than or equal, so it's `x <=u y`

include/seahorn/Expr/ExprOpBv.hh
```diff
@@ -165,59 +179,188 @@ enum class BvOpKind {
...
+NOP_TYPECHECK(BSLE, "bvsle", FUNCTIONAL, BvOp, typeCheck::bvType::Nary)
+NOP_TYPECHECK(BUGE, "bvuge", FUNCTIONAL, BvOp, typeCheck::bvType::Nary)
+NOP_TYPECHECK(BSGE, "bvsge", FUNCTIONAL, BvOp, typeCheck::bvType::Nary)
+NOP_TYPECHECK(BUGT, "bvugt", FUNCTIONAL, BvOp, typeCheck::bvType::Nary)
+NOP_TYPECHECK(BSGT, "bvsgt", FUNCTIONAL, BvOp, typeCheck::bvType::Nary)
+NOP_TYPECHECK(BCONCAT, "concat", FUNCTIONAL, BvOp, typeCheck::bvType::Concat)
```
agurfinkel, line 325: concat can be nary -- did not check whether this is supported, but ran into this problem before so mentioning it here

include/seahorn/Expr/ExprOpBv.hh
```diff
@@ -165,59 +179,188 @@ enum class BvOpKind {
...
+NOP_TYPECHECK(BEXTRACT, "extract", FUNCTIONAL, BvOp, typeCheck::bvType::Extract)
+NOP_TYPECHECK(BSEXT, "bvsext", FUNCTIONAL, BvOp, typeCheck::bvType::Extend)
+NOP_TYPECHECK(BZEXT, "bvzext", FUNCTIONAL, BvOp, typeCheck::bvType::Extend)
+NOP_TYPECHECK(BREPEAT, "bvrepeat", FUNCTIONAL, BvOp, typeCheck::bvType::Rotate)
```
agurfinkel, line 329: is that true that type of `bvrepeat` is same as `Rotate`?

include/seahorn/Expr/ExprOpBv.hh
```diff
@@ -165,59 +179,188 @@ enum class BvOpKind {
...
+NOP_TYPECHECK(BEXT_ROTATE_RIGHT, "bvextrotright", FUNCTIONAL, BvOp,
+              typeCheck::bvType::Rotate)
+NOP_TYPECHECK(INT2BV, "int2bv", FUNCTIONAL, BvOp, typeCheck::bvType::Int2Bv)
+NOP_TYPECHECK(BV2INT, "bv2int", FUNCTIONAL, BvOp, typeCheck::bvType::Bv2Int)
 // Add w Overflow
-NOP(SADD_NO_OVERFLOW, "bvsadd_no_overflow", FUNCTIONAL, BvOp)
-NOP(UADD_NO_OVERFLOW, "bvuadd_no_overflow", FUNCTIONAL, BvOp)
-NOP(SADD_NO_UNDERFLOW, "bvbadd_no_underflow", FUNCTIONAL, BvOp)
+NOP_TYPECHECK(SADD_NO_OVERFLOW, "bvsadd_no_overflow", FUNCTIONAL, BvOp,
+              typeCheck::bvType::NaryBool)
```
agurfinkel, line 345: this and other overflow / underflow functions are binary. Return a bool indicating whether corresponding arithmetic operation would overflow

include/seahorn/Expr/ExprOpCompare.hh
```diff
@@ -4,19 +4,36 @@
...
 namespace expr {
 
 namespace op {
 enum class CompareOpKind { EQ, NEQ, LEQ, GEQ, LT, GT };
+namespace typeCheck {
+namespace compareType {
+
+struct Equality {
+  static inline Expr inferType(Expr exp, TypeChecker &tc) {
+    return typeCheck::binary<BOOL_TY, ANY_TY>(exp, tc);
```
agurfinkel, line 18: would appreciate a comment

include/seahorn/Expr/ExprOpCore.hh
```diff
@@ -115,6 +123,9 @@ public:
     return llvm::isa<TerminalBase>(op) &&
            llvm::cast<TerminalBase>(op)->m_kind == terminal_type::getKind();
   }
+
+ // TODO
```
agurfinkel, line 127: why is this marked as a TODO?

include/seahorn/Expr/ExprOpCore.hh
```diff
@@ -46,8 +53,8 @@ struct TerminalBase : public expr::Operator {
 
 /// \brief Terminal Operator
 /// \p T is the type wrapped by the terminal, and \p P is the implementation of
-/// terminal traits for \p T
-template <typename T, typename P = TerminalTrait<T>>
+/// terminal traits for \p T , C is the tyepChecker
+template <typename T, typename P = TerminalTrait<T>, typename C = typeCheck::Any>
```
agurfinkel, line 57: I think type checking should be part of `TerminalTrait`, not sometehing specified on the side. Just like printing and ordering.

include/seahorn/Expr/ExprOpCore.hh
```diff
@@ -349,13 +358,14 @@ using namespace ps;
...
-template <typename T, typename B, typename P, typename K, K kind>
+template <typename T, typename B, typename P, typename K, K kind, typename C>
 struct DefOp : public B {
-  using this_type = DefOp<T, B, P, K, kind>;
+  using this_type = DefOp<T, B, P, K, kind, C>;
   using kind_type = K;
   using base_type = B;
   using op_type = T;
   using ps_type = P;
+  using typeChecker_type = C;
```
agurfinkel, line 368: probably `checker_type` fits better with the rest of the context

include/seahorn/Expr/ExprOpGate.hh
```diff
@@ -20,9 +20,9 @@ struct GateOp : public expr::Operator {
 
 /// an output gate
 NOP(OUT_G, "OUT_G", PREFIX, GateOp)
-NOP(AND_G, "/\\", INFIX, GateOp);
-NOP(OR_G, "\\/", INFIX, GateOp);
-NOP(NEG_G, "~", PREFIX, GateOp);
+NOP_TYPECHECK(AND_G, "/\\", INFIX, GateOp, typeCheck::boolType::Nary);
```
agurfinkel, line 23: Shouldn't `NOP_TYPECHECK` be renamed into `NOP` and replace all uses of NOP.

include/seahorn/Expr/ExprOpSort.hh
```diff
@@ -18,36 +17,52 @@ enum class SimpleTypeOpKind {
...
-  ANY_TY
+  ANY_TY,
+  ERROR_TY
 };
+
+namespace typeCheck {
+  namespace simpleType {
+struct Simple {
+  static inline Expr inferType(Expr exp, TypeChecker &tc) {
+    return exp;
```
agurfinkel, line 28: type of a type is `ANY`, not type itself. For example, the type of `int` is not `int`, it is `type int`. But we don't need the hierarchy that deep, so we can conflate it into `any`

include/seahorn/Expr/ExprOpStruct.hh
```diff
@@ -6,15 +6,82 @@
...
 enum class StructOpKind { MK_STRUCT, EXTRACT_VALUE, INSERT_VALUE };
+
+namespace typeCheck {
+namespace structType {
+struct Struct {
+  static inline Expr inferType(Expr exp, TypeChecker &tc) {
+    ExprVector args;
+    bool wellFormed = true;
+
+    for (auto b = exp->args_begin(), e = exp->args_end(); wellFormed && b != e;
```
agurfinkel, line 21: comment would be welcome

include/seahorn/Expr/ExprOpVariant.hh
```diff
@@ -33,9 +33,36 @@ struct PS_TAG {
 } // namespace variant
 
 enum class VariantOpKind { VARIANT, TAG };
+
+namespace typeCheck {
+namespace variantType {
```
agurfinkel, line 38: these expressions are just for creating structured names. Their types is `any`.

include/seahorn/Expr/ExprOpGate.hh
```diff
@@ -20,9 +20,9 @@ struct GateOp : public expr::Operator {
 
 /// an output gate
 NOP(OUT_G, "OUT_G", PREFIX, GateOp)
```
agurfinkel, line 22: Gates are just mutable boolean operators. Their inputs are boolean and their outputs are boolean.

an output gate `OUT_G` marks an output of a function. Its input is a Boolean and output a Boolean

include/seahorn/Expr/ExprOpNum.hh
```diff
@@ -53,4 +75,4 @@ struct ITV_PS {
 } // namespace numeric
 NOP(ITV, "itv", numeric::ITV_PS, NumericOp)
```
agurfinkel, line 76: interval. Type is a pair of numbers.

include/seahorn/Expr/ExprOpNum.hh
```diff
@@ -22,18 +23,39 @@ enum class NumericOpKind {
...
+NOP_TYPECHECK(MINUS, "-", INFIX, NumericOp, typeCheck::numType::Nary)
+NOP_TYPECHECK(MULT, "*", INFIX, NumericOp, typeCheck::numType::Nary)
+NOP_TYPECHECK(DIV, "/", INFIX, NumericOp, typeCheck::numType::Nary)
+NOP_TYPECHECK(IDIV, "/", INFIX, NumericOp, typeCheck::numType::Nary)
+NOP_TYPECHECK(MOD, "mod", INFIX, NumericOp, typeCheck::numType::Nary)
+NOP_TYPECHECK(REM, "%", INFIX, NumericOp, typeCheck::numType::Nary)
+NOP_TYPECHECK(UN_MINUS, "-", PREFIX, NumericOp, typeCheck::numType::Nary)
+NOP_TYPECHECK(ABS, "abs", FUNCTIONAL, NumericOp, typeCheck::numType::Unary)
 
 NOP(PINFTY, "oo", PREFIX, NumericOp)
```
agurfinkel, line 60: Infinitys are used in intervals to mark lack of a bound, as in `[0, oo]` to represent all positive numbers. I don't think these expressions are used now. It is ok to leave them untyped (i.e., type as `any`)

include/seahorn/Expr/TypeChecker.hh
```diff
@@ -1,17 +1,168 @@
...
   class Impl;
-  Impl* m_impl;
+  Impl *m_impl;
 
 public:
   TypeChecker();
   ~TypeChecker();
   Expr sortOf(Expr e) { return this->typeOf(e); }
-  Expr typeOf(Expr e); 
+  Expr typeOf(Expr e);
```
agurfinkel, line 14: comment each method. These are user facing and need substantive documentation.
substantive == 3-4 lines

include/seahorn/Expr/TypeChecker.hh
```diff
@@ -1,17 +1,168 @@
...
+  Expr getErrorExp();         // to be called after typeOf() or sortOf()
+  ExprSet getEndExps(Expr e); // to be called after typeOf() or sortOf()
 };
 
+namespace op {
```
agurfinkel, line 19: These don't belong to the same header file. The code up to hear is likley to be included by every client of type checker. The code after here is only needed to implement type checking functionality. Maybe move it into `TypeCheckerDefs.hh` or `TypeCheckerUtils.hh`

include/seahorn/Expr/TypeChecker.hh
```diff
@@ -1,17 +1,168 @@
...
+namespace typeCheck {
+enum Comparison { Equal, GreaterEqual };
+
+template <Comparison compareType, unsigned int numChildren>
+static inline bool checkNumChildren(Expr exp) {
```
agurfinkel, line 24: the way you use this functions, it seems unnatural to pass `numChildren` as a template parameter. This increases compilation time without much gain in performance. Might as well pass it as an argument.

Probably this is because I've suggested it. But I had a different -- more template based -- design in mind where this would make more sense.

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -6,41 +6,134 @@
...
+  void reset(Expr exp) {
+    if (m_isWellFormed)
+      m_errorExp = Expr();
+
+    m_topMost = Expr();
+    m_isWellFormed = true;
+    m_cache.clear();
+  }
+
+  bool isConst(Expr exp) {
```
agurfinkel, line 32: probably no longer needed

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -6,41 +6,134 @@
...
+    return bind::isBoolConst(exp) || bind::isIntConst(exp) ||
+           bind::isRealConst(exp) || bind::isUnintConst(exp) ||
+           bind::isArrayConst(exp) || bind::isStructConst(exp) ||
+           bv::isBvConst(exp);
+  }
+
+  bool isValue(Expr exp) { return isOpX<TRUE>(exp) || isOpX<FALSE>(exp); }
```
agurfinkel, line 39: looks like a bunch of this is old code that needs to be cleaned up. I will ignore anything that I believe to be stale

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -6,41 +6,134 @@
...
+  }
+
+  void mapEndTypes(Expr exp) {
+    ExprSet set;
+
+    if (exp->arity() == 0) {
+      set.insert(exp);
+    }
+
+    std::for_each(exp->args_begin(), exp->args_end(), [this, &set](Expr child) {
```
agurfinkel, line 52: can use `llvm::make_range` and a normal `for` loop instead.

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -6,41 +6,134 @@
...
   Expr postVisit(Expr exp) {
     LOG("tc", llvm::errs() << "post visiting expression: " << *exp << "\n";);
+
+    Expr type = exp->inferType(exp, *m_tc);
+
+    m_cache.insert({exp, type});
+    mapEndTypes(exp);
+
+    if (isOp<ERROR_TY>(type)) {
+      LOG("tc", llvm::errs()
```
agurfinkel, line 72: use `ERR` instead of `llvm::errs()`

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -6,41 +6,134 @@
...
-    // if e is non null, return ANY type, ow return nil expression
-    return e ? sort::anyTy(e->efac()) : Expr();
+    Expr knownType;
+
+    if (m_isWellFormed) 
+      knownType = e ? m_cache.at(e) : Expr();
+    else
+      knownType = sort::errorTy(e->efac());
+
+    if (e == m_topMost) // done traversing entire expression
```
agurfinkel, line 120: Not sure why **top most** expression is important. Some comments on that will be useful.

In particular, we would like to run type checker over multiple expressions (think vector of expressions), while re-using the type cache. In this case, top-most expression changes all the time.

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -6,41 +6,134 @@
...
+      reset(e);
+
+    return knownType;
   }
+
+  Expr getErrorExp() { return m_errorExp; }
+
+  ExprSet getEndExps(Expr e) { return m_endExpr.at(e); }
```
agurfinkel, line 128: comment what `EndExps` are

include/seahorn/Expr/ExprOpBinder.hh
```diff
@@ -126,14 +125,62 @@ struct BINDER {
...
+  ExprSet boundVars(exp->args_begin(), exp->args_end() - 1);
+
+  auto contains = [&boundVars](Expr exp) -> bool {
+    return boundVars.count(exp);
+  };
+
+  return std::all_of(bodyEndExps.begin(), bodyEndExps.end(), contains);
+}
+
+struct Lambda {
```
agurfinkel, line 147: lambda's are functional type. For example, `lambda x :: x + 1` has type: `typeOf(x) -> typeOf(x + 1)`
where `typeOf` is the type of appropriate expression. Internally, lambda arguments are represented with bound variables. So the expression above looks like: `lambda :: B0 + 1`, where `B0` is the only bound variable and is therefore the argument. You can infer the type for arguments from the lambda operator itself (it keep track of how many variables is bounds and their sorts), you can infer the type of the body from type checking the body recursively.

lib/seahorn/Smt/TypeChecker.cc
colinxm: The problem that I had was that it skips all children and returns an error if it finds something wrong. Without the topmost expression, if you find an error every expression after that will also say that its an error, even if its not.

colinxm: I could keep the top most expression and just not clear the cache when I reset everything

colinxm: or maybe just remove it all together with the assumption that when you find an error it will say error for everything after that?

lib/seahorn/Smt/TypeChecker.cc
colinxm: So I think I’d like your input on this if possible. I’m not sure if the way I’m going about this is good.

I used this as a part of finding the end of an expression (consts, vars, etc.) so that I can check if the bounded variables in quantifiers and functions match the variables in the body. One I reach a simple type (const, vars …) I put it in a map. This seems hard coded to me but I’m not really sure how else to approach this.

lib/seahorn/Smt/TypeChecker.cc
agurfinkel: If an expression is typed as `errorTy`, then all of its parents will be typed as `errorTy`. This seems good and also stable (so can be cached). The first expression that was typed as `errorTy` can be logged in a field to be returned to the user. User can reset last error during the next call (or the reset can happen automatically).

lib/seahorn/Smt/TypeChecker.cc
agurfinkel: I do not understand the problem that you are solving. Can you be more explicit?

Perhaps you need to know whether an expression is a value or an operator. If so, this is a property of the operator and should be moved there.

colinxm: So I guess I could just make a generic function to check if its , for example, a const but I'm not sure if there's other types that could be used as bounded variables in quantifiers that I'm forgetting about

colinxm: sorry I made that comment without refreshing the page.  I didn't realize that you had responded. if for example you had for all (a b (a && b &&c)), c is not bounded so that would be an error

colinxm: Ya I think checking whether it is a value or operator would work, I'll add that in

lib/seahorn/Smt/TypeChecker.cc
colinxm: So with that being said, maybe I'll make the reset happen automatically (using the topmost expression)

lib/seahorn/Smt/TypeChecker.cc
agurfinkel: I still don't understand what you are doing. The only bound variables that can appear under a binder are `BVar`, all other terminals are constants and cannot be bound.

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -6,41 +6,134 @@
...
+  void mapEndTypes(Expr exp) {
+    ExprSet set;
+
+    if (exp->arity() == 0) {
+      set.insert(exp);
+    }
+
+    std::for_each(exp->args_begin(), exp->args_end(), [this, &set](Expr child) {
+      if (this->m_endExpr.count(child)) {
+        ExprSet childSet = this->m_endExpr.at(child);
```
agurfinkel, line 54: this creates an expensive copy of a set

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -6,41 +6,134 @@
...
+           bv::isBvConst(exp);
+  }
+
+  bool isValue(Expr exp) { return isOpX<TRUE>(exp) || isOpX<FALSE>(exp); }
+
+  bool isSimpleType(Expr exp) {
+    return isConst(exp) || isValue(exp) || bv::is_bvnum(exp);
+  }
+
+  void mapEndTypes(Expr exp) {
```
agurfinkel, line 45: I don't understand this function. Creating a map from expressions to sets sounds quite expensive. I need to know more about what it is for.

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -6,41 +6,134 @@
...
 using namespace expr;
-
 namespace {
 // local code in this namespace
-
 //==-- main implementation goes here --==//
 class TCVR {
 
+  ExprMap m_cache;
+  std::map<Expr, ExprSet> m_endExpr;
```
agurfinkel, line 15: document what this map is for

lib/seahorn/Smt/TypeChecker.cc
agurfinkel: the format for a binder is `(binder (a:Int b:Int c:Int) body))`, the decls `a`, `b`, and `c` are just arbitrary names. They do not appear in `body`. Technically they can, but their meaning would be different. The only important part of them is the sort. Perhaps that is the source of confusion.

colinxm: Oh I see, I guess I didn't understand the structure of it. Ill fix that which should also fix the other comments about the expensive set copy and mapping

agurfinkel: Well, the code is not super documented (please help in your code!), so it is not easy for you to figure out on your own. Ask questions when in doubt.

include/seahorn/Expr/ExprOpBind.hh
```diff
@@ -26,15 +26,102 @@ struct SCOPE_PS {
...
+  // Return type: FUNCTIONAL_TY
+  static inline Expr inferType(Expr exp, TypeCheckerHelper &helper) {
+    if (exp->arity() < 2)
+      return sort::errorTy(exp->efac());
+
+    auto isType = [&helper](Expr exp) -> bool {
+      return isOp<TYPE_TY>(helper.typeOf(exp));
+    };
+
+    auto begin = exp->args_begin() + 1; // note: name is the first child
```
agurfinkel, line 78: `++exp->args_begin()` is more robust, or even `std::advance`

include/seahorn/Expr/ExprOpBind.hh
```diff
@@ -26,15 +26,102 @@ struct SCOPE_PS {
...
+
+    Expr functionalType =
+        helper.typeOf(exp->first()); // functional types should be of the form :
+                                     // arg1Type, arg2Type ... -> returnType
+
+    if (!(isOp<FUNCTIONAL_TY>(functionalType) &&
+          exp->arity() == functionalType->arity()))
+      return sort::errorTy(exp->efac());
+
+    auto fappArgs = exp->args_begin() + 1; // note: the first child is the fdecl
```
agurfinkel, line 104: use `++` or `std::advance`

include/seahorn/Expr/ExprOpBind.hh
```diff
@@ -26,15 +26,102 @@ struct SCOPE_PS {
...
+    // Check that each fapp argument type matches its corresponding functional
+    // arguments
+    while (fappArgs != exp->args_end()) {
+      Expr fappArgType = helper.typeOf(*fappArgs);
+      Expr functionalArgType = *functionalArgs;
+
+      if (fappArgType != functionalArgType)
+        return sort::errorTy(exp->efac());
+
+      fappArgs++;
```
agurfinkel, line 116: use `++fappArgs`, same for next line

include/seahorn/Expr/ExprOpBinder.hh
```diff
@@ -126,14 +128,86 @@ struct BINDER {
...
+namespace bindType {
+
+static inline bool binderCheck(Expr exp, TypeCheckerHelper &helper,
+                               ExprVector &boundTypes) {
+  if (exp->arity() == 0)
+    return false;
+
+  // makes sure that all of the binders arguments are constants and store their
+  // types
+  for (auto b = exp->args_begin(), e = exp->args_end() - 1; b != e; b++) {
```
agurfinkel, line 141: I have not looked at all the supporting function, but this might be more complex than you realize. I think at this point, I'd rather skip this check completely.

I am not sure how your `getBoundVars` is implemented. It should take nested quantifiers into account. Not sure if it does.

Similarly, the top-down visitor should take nested quantifiers into account and maintain a stack with what quantifier is being processed. Not all type checking results can be cached across different quantifiers (this is a conjecture right now, I need to think of an example to be sure). 

Essentially, in a top-down traverse we must maintain the list of all variables that are currently bound and their types, then whenever a bound variable is processed, it can look up that its type is correct.

A yet another alternative is to type a BVar variable as a free function, so that a body of the quantifier has a functional type. Then the binder checks that it properly binds the right variables. 

But again, this is getting complicated, but exactly in the cases that we do not currently use.

So I propose to type-check that the body is well typed and ignore checking the binders.

include/seahorn/Expr/TypeChecker.hh
```diff
@@ -1,17 +1,66 @@
...
-#include "seahorn/Expr/ExprApi.hh"
+#include "seahorn/Expr/ExprCore.hh"
+#include "seahorn/Expr/ExprOpSort.hh"
 
 namespace expr {
-class TypeChecker {
+
+class TypeChecker;
+
+class TypeCheckerHelper {
```
agurfinkel, line 9: Lots of comments are missing. Especially in this file. What is a `TypeCheckerHelper`, why there is a need both for `TypeChecker` and helper. How to extend type checker when new operators are added. What is special about terminals, etc.

include/seahorn/Expr/TypeChecker.hh
```diff
@@ -1,17 +1,66 @@
...
-  ~TypeChecker();
+
+  /*
+  See typeOf(Expr e);
+  */
   Expr sortOf(Expr e) { return this->typeOf(e); }
-  Expr typeOf(Expr e); 
+
+  /*
+  Returns the type of the passed expression. If an error is found (ie. the
```
agurfinkel, line 51: Use doxygen style in these user facing comments. We don't really run doxygen, but it helps in structuing the comments and referring to argument names and returns. Light doxygen is ok

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -3,44 +3,158 @@
...
+
+  TypeCheckerHelper *const m_helper;
+
+  // Keeps track of the expression that the typechecker is called with.
+  // The expression is done traversing when it reaches the top most expression
+  // so it can reset
+  Expr m_topMost;
+
+  // Keeps track of all the bound variables that an expression uses
+  std::map<Expr, ExprSet> m_boundVarMap;
```
agurfinkel, line 32: as in the comment elsewhere, bound variables are more complex than they appear, let's ignore them for now

agurfinkel:
Merged manually. @colinxm don't forget to update your repo. After my commit, your `dev10` branch is out of sync. You should rename it and start from the current `dev10` instead.

PR #300, opened 2020-05-20
Use llvm10 utilities to promote indirect calls

caballa:


include/seahorn/Transforms/Utils/DevirtFunctions.hh
```diff
@@ -22,6 +22,8 @@ class CompleteCallGraphAnalysis;
 
 namespace seahorn {
 
+//#define USE_BOUNCE_FUNCTIONS
```
agurfinkel, line 25: you can go all the way and add a cmake option to turn this on during compilation

caballa: Yes

caballa:
@agurfinkel : note that we could use as an alternative method to devirtualize the metadata that adds `CalledValuePropagation` which is the LLVM analysis that resolves indirect calls based on the sparse propagator solver. I didn't bother because it shouldn't be better than seadsa.

agurfinkel:
ready to merge

caballa:
Yes, let me move the pragma flag to cmake

PR #303, opened 2020-05-28
feat(opsem): add wide memory mgr class

priyasiddharth:
Wide memory manager (similar to a fat memory manager) carries extra information about pointers in the memory model. Namely, the size of the allocated region starting from the ptr address.

TBD:  to add a isdeferencable() function so that programs can call it to find invalid ptrs.

### DIFF
--- a/lib/seahorn/BvOpsem2WideMemMgr.cc
+++ b/lib/seahorn/BvOpsem2WideMemMgr.cc
@@ -1,0 +1,562 @@
+#include "BvOpSem2Context.hh"
+#include "BvOpSem2RawMemMgr.hh"
+
+#include "BvOpSem2MemManagerMixin.hh"
+
+#include "seahorn/Expr/ExprOpStruct.hh"
+#include "seahorn/Support/SeaDebug.h"
+#include "seahorn/Support/SeaLog.hh"
+
+static const unsigned int g_slotBitWidth = 64;
+static const unsigned int g_slotByteWidth = g_slotBitWidth / 8;
+
+static const unsigned int g_uninit = 0xDEADBEEF;
+
+namespace seahorn {
+namespace details {
+class WideMemManager : public OpSemMemManagerBase {
+
+  /// \brief Knows the memory representation and how to access it
+  std::unique_ptr<OpSemMemRepr> m_memRepr;
+
+  /// \brief Base name for non-deterministic pointer
+  Expr m_freshPtrName;
+
+  /// \brief Register that contains the value of the stack pointer on
+  /// function entry
+  Expr m_sp0;
+
+  /// \brief Source of unique identifiers
+  mutable unsigned m_id;
+
+  const Expr m_uninit_size;
+
+  /// \brief Memory manager for raw pointers
+  RawMemManager m_main;
+  RawMemManager m_size;
+
+public:
+  using RawPtrTy = OpSemMemManager::PtrTy;
+  using RawMemValTy = OpSemMemManager::MemValTy;
+  using RawPtrSortTy = OpSemMemManager::PtrSortTy;
+  using RawMemSortTy = OpSemMemManager::MemSortTy;
+  using MemRegTy = OpSemMemManager::MemRegTy;
+
+  // size = size in bits
+  struct PtrTyImpl {
+    Expr m_v;
+
+    PtrTyImpl(RawPtrTy &&ptr, Expr &&size) {
+      m_v = strct::mk(std::move(ptr), std::move(size));
+    }
+
+    PtrTyImpl(const RawPtrTy &ptr, const Expr &size) {
+      m_v = strct::mk(ptr, size);
+    }
+
+    explicit PtrTyImpl(const Expr &e) {
+      // Our ptr is a struct of two exprs
+      assert(strct::isStructVal(e));
+      assert(e->arity() == 2);
+      m_v = e;
+    }
+
+    Expr v() const { return m_v; }
+    Expr toExpr() const { return v(); }
+    explicit operator Expr() const { return toExpr(); }
+
+    RawPtrTy getRaw() { return strct::extractVal(m_v, 0); }
+
+    RawPtrTy getSize() { return strct::extractVal(m_v, 1); }
+  };
+
+  struct MemValTyImpl {
+    Expr m_v;
+
+    MemValTyImpl(RawMemValTy &&raw_val, Expr &&size_val) {
+      m_v = strct::mk(std::move(raw_val), std::move(size_val));
+    }
+
+    MemValTyImpl(const RawPtrTy &raw_val, const RawMemValTy &size_val) {
+      m_v = strct::mk(raw_val, size_val);
+    }
+
+    explicit MemValTyImpl(const Expr &e) {
+      // Our ptr is a struct of two exprs
+      assert(strct::isStructVal(e));
+      assert(e->arity() == 2);
+      m_v = e;
+    }
+
+    Expr v() const { return m_v; }
+    Expr toExpr() const { return v(); }
+    explicit operator Expr() const { return toExpr(); }
+
+    RawMemValTy getRaw() { return strct::extractVal(m_v, 0); }
+
+    RawMemValTy getSize() { return strct::extractVal(m_v, 1); }
+  };
+
+  struct PtrSortTyImpl {
+    Expr m_ptr_sort;
+
+    PtrSortTyImpl(RawPtrSortTy &&ptr_sort, Expr &&size_sort) {
+      m_ptr_sort = sort::structTy(std::move(ptr_sort), std::move(size_sort));
+    }
+
+    PtrSortTyImpl(const RawPtrSortTy &ptr_sort, const Expr &size) {
+      m_ptr_sort = sort::structTy(ptr_sort, size);
+    }
+
+    Expr v() const { return m_ptr_sort; }
+    Expr toExpr() const { return v(); }
+    explicit operator Expr() const { return toExpr(); }
+
+    RawPtrSortTy getRawSort() { return m_ptr_sort->arg(0); }
+  };
+
+  struct MemSortTyImpl {
+    Expr m_mem_sort;
+
+    MemSortTyImpl(RawMemSortTy &&mem_sort, Expr &&size_sort) {
+      m_mem_sort = sort::structTy(std::move(mem_sort), std::move(size_sort));
+    }
+
+    MemSortTyImpl(const RawMemSortTy &mem_sort, const Expr &size_sort) {
+      m_mem_sort = sort::structTy(mem_sort, size_sort);
+    }
+
+    Expr v() const { return m_mem_sort; }
+    Expr toExpr() const { return v(); }
+    explicit operator Expr() const { return toExpr(); }
+  };
+
+  using PtrTy = PtrTyImpl;
+  using PtrSortTy = PtrSortTyImpl;
+  using MemValTy = MemValTyImpl;
+  using MemSortTy = MemSortTyImpl;
+
+  /// \brief A null pointer expression (cache)
+  PtrTy m_nullPtr;
+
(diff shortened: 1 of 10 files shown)


lib/seahorn/CMakeLists.txt
```diff
@@ -42,7 +42,8 @@ add_llvm_library (seahorn.LIB DISABLE_LLVM_LINK_LLVM_DYLIB
   BvOpSem2ConstEval.cc
   BvOpSem2RawMemMgr.cc
   BvOpSem2FatMemMgr.cc
-  VCGen.cc
+        BvOpsem2WideMemMgr.cc
```
agurfinkel, line 45: indentation

test/opsem/lit.cfg
```diff
@@ -54,7 +54,8 @@ seabmc = [seahorn_cmd, '--horn-bmc-engine=mono',
 # no --horn-solve
 seasmt = [seahorn_cmd, '--horn-bmc-engine=mono',
           '--horn-bmc', '--horn-bv2=true', '--log=opsem',
-          '--keep-shadows=true' ]
+          '--horn-bv2-widemem',
```
agurfinkel, line 57: this will override our existing tests

lib/seahorn/BvOpsem2WideMemMgr.cc
```diff
@@ -0,0 +1,543 @@
...
+  Expr m_freshPtrName;
+
+  /// \brief Register that contains the value of the stack pointer on
+  /// function entry
+  Expr m_sp0;
+
+  /// \brief Source of unique identifiers
+  mutable unsigned m_id;
+
+  const Expr m_uninit_size = m_ctx.alu().si(g_uninit, g_slotBitWidth);
```
agurfinkel, line 35: I am never clear about the rules when such initialization are ok. They were not ok before in C++ and now they are sometimes ok. I would prefer for this to be moved to the constructor.

lib/seahorn/BvOpsem2WideMemMgr.cc
```diff
@@ -0,0 +1,543 @@
...
+                     m_ctx.alu().intTy(g_slotBitWidth));
+  }
+
+  PtrTy salloc(unsigned int bytes, uint32_t align) {
+    assert(isa<AllocaInst>(m_ctx.getCurrentInst()));
+    align = std::max(align, m_alignment);
+    auto region = m_allocator->salloc(bytes, align);
+    assert(region.first - region.second > 0);
+    return PtrTy(mkStackPtr(region.second).getRaw(),
+                 bytesToSlotExpr(region.first - region.second));
```
agurfinkel, line 165: You can add a constructor to `PtrTy` that takes an `Expr` and an `unsigned` and converts the size to `Expr` in the constructor.

lib/seahorn/BvOpsem2WideMemMgr.cc
```diff
@@ -0,0 +1,543 @@
...
+      LOG("opsem", WARN << "imprecise handling of dynamically "
+                        << "sized stack allocation of " << *elmts
+                        << " elements of size" << typeSz << " bytes\n";);
+      return freshPtr();
+    }
+
+    // -- have a good region, return pointer to it
+    return PtrTy(
+        mkStackPtr(region.second).getRaw(),
+        m_ctx.alu().si(8 * (region.first - region.second), g_slotBitWidth));
```
agurfinkel, line 193: why `8 * ...`. I was expecting size to be in bytes.

lib/seahorn/BvOpsem2WideMemMgr.cc
```diff
@@ -0,0 +1,543 @@
...
+    PtrTy res = freshPtr();
+    LOG("opsem", WARN << "halloc() not implemented!\n");
+    return res;
+  }
+
+  PtrTy galloc(const GlobalVariable &gv, uint32_t align) {
+    uint64_t gvSz = m_sem.getTD().getTypeAllocSize(gv.getValueType());
+    auto range = m_allocator->galloc(gv, gvSz, std::max(align, m_alignment));
+    return PtrTy(m_ctx.alu().si(range.first, ptrSzInBits()),
+                 bytesToSlotExpr(range.second - range.first));
```
agurfinkel, line 218: maybe `range` should have `size()` method so that we don't have to do this computation each time

lib/seahorn/BvOpsem2WideMemMgr.cc
```diff
@@ -0,0 +1,543 @@
...
+  PtrTy falloc(const Function &fn) {
+    auto range = m_allocator->falloc(fn, m_alignment);
+    return PtrTy(m_ctx.alu().si(range.first, ptrSzInBits()),
+                 bytesToSlotExpr(range.second - range.first));
+  }
+
+  // TODO: What is the right size to return here?
+  PtrTy getPtrToFunction(const Function &F) {
+    m_size.getPtrToFunction(F);
+    return PtrTy(m_main.getPtrToFunction(F), m_uninit_size);
```
agurfinkel, line 230: Technically, the size of the allocation is controlled by `falloc()` above. `m_main.getPtrToFunction` will get a range from the allocator and return `range.first`. You can add a method in the allocator to return size of any base pointer that it knows about.

lib/seahorn/BvOpsem2WideMemMgr.cc
```diff
@@ -0,0 +1,543 @@
...
+  }
+
+  PtrTy getPtrToGlobalVariable(const GlobalVariable &gv) {
+    uint64_t gvSz = m_sem.getTD().getTypeAllocSize(gv.getValueType());
+    return PtrTy(m_ctx.alu().si(
+                     m_allocator->getGlobalVariableAddr(gv, gvSz, m_alignment),
```
agurfinkel, line 236: a similar thing can be done for `getPtrToFunction` above.

lib/seahorn/BvOpsem2WideMemMgr.cc
```diff
@@ -0,0 +1,543 @@
...
+                     ptrSzInBits()),
+                 bytesToSlotExpr(gvSz));
+  }
+
+  void initGlobalVariable(const GlobalVariable &gv) const {
```
agurfinkel, line 241: why is this not simply forwarded to `m_main`?

lib/seahorn/BvOpsem2WideMemMgr.cc
```diff
@@ -0,0 +1,543 @@
...
+
+  Expr coerce(Expr sort, Expr val) { return m_memRepr->coerce(sort, val); }
+
+  PtrTy ptrAdd(PtrTy ptr, int32_t _offset) const {
+    if (_offset == 0)
+      return ptr;
+    expr::mpz_class offset((signed long)_offset);
+    auto address = m_ctx.alu().doAdd(
+        ptr.getRaw(), m_ctx.alu().si(offset, ptrSzInBits()), ptrSzInBits());
+    auto new_size = m_ctx.alu().doSub(
```
agurfinkel, line 294: optimization opportunity: `size` might be a number at this point, in which case sub can be done concretely and not symbolically

lib/seahorn/CMakeLists.txt
priyasiddharth: done

lib/seahorn/BvOpsem2WideMemMgr.cc
priyasiddharth: ack.

lib/seahorn/BvOpsem2WideMemMgr.cc
priyasiddharth: Let me know if there is another way. m_ctx is not accessible from PtrTy. I will need to pass a reference of WideMemMgr to PtrTy which makes it complicated compared to current `bytesToSlotExpr` impl

lib/seahorn/BvOpsem2WideMemMgr.cc
priyasiddharth: good catch!

lib/seahorn/BvOpsem2WideMemMgr.cc
agurfinkel: why can't `m_unit_size` be initialized in the initializer list right before `m_nullptr`?

lib/seahorn/BvOpsem2WideMemMgr.cc
agurfinkel: I thought that `region.first - region.second` is a number and you convert it here to an expr.
Maybe I missed something.

lib/seahorn/BvOpsem2WideMemMgr.cc
priyasiddharth: Good point. done.

lib/seahorn/BvOpsem2WideMemMgr.cc
priyasiddharth: Currently I have

```
Expr bytesToSlotExpr(unsigned bytes) {
    return m_ctx.alu().si(bytes, g_slotBitWidth);
  }
```

lib/seahorn/BvOpsem2WideMemMgr.cc
priyasiddharth: AddrInterval is a std::pair... i can
1. add a size() fn in bvopsem2context OR
2. change AddrInterval to be a class with a size() operator OR
3. probably add a template method -- will have to figure out how...

agurfinkel: make it into a proper class. It is a pair because it did not need additional methods before

lib/seahorn/BvOpsem2WideMemMgr.cc
priyasiddharth: Added for functions for now.

lib/seahorn/BvOpsem2WideMemMgr.cc
priyasiddharth: Let this be for now... added todo since this is not on critical path imo.

lib/seahorn/BvOpsem2WideMemMgr.cc
priyasiddharth: ack. refactoring artifact.

lib/seahorn/BvOpsem2WideMemMgr.cc
priyasiddharth: good point. done

test/opsem/lit.cfg
priyasiddharth: removed.

priyasiddharth:
I am tracking overall changes in https://github.com/priyasiddharth/seahorn/projects/1. widememmgr currently works with opsem tests except nham.01.ll where for nham.01.ll without -horn-solve, the test does not output any sat/unsat results so it just sits there. With horn solve it works as intended. 

In any case, my preference is to have this submitted and submit enhancements, fixes incrementallt as per the project tracker.

agurfinkel:
I'm ready to accept the PR as soon as you think it is ready (i.e., before all functionality is there but when the initial functionality is sufficiently stable).

The **just sitting there** on `nham.01.ll` example is concerning since it usually indicates an infinite loop. You should be able to probe it with the debugger (i.e., by running in a debugger and stopping with Ctrl-C to see where you are in the code).

The fact that the example works with `--solve` is even stranger since this means that the infinite loop is not in the encoding.

One thing that comes to mind are lambda functions. Printing them to disk is problematic -- there is a certain inlining happening that causes an exponential explosion.

lib/seahorn/BvOpsem2WideMemMgr.cc
agurfinkel: I see. Too much context is then needed to move this into the constructor -- which is what you said. Maybe inlining the function is more readable. I think I got confused by the name.

PR #304, opened 2020-05-29
feat(opsem): use memmssa api to define memin, memout reg for bvopsem2

priyasiddharth:
Use seamemoryaccess objects to point to the right memory aaray to use instead of relying on shadowmem calls. Currently we use the shadowmem calss as keyto get to seamemoryaccess objects. Also added seamemoryaccess as a terminal


lib/seahorn/BvOpSem2.cc
```diff
@@ -595,119 +601,234 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemVisitorBase {
...
+                dyn_cast<seadsa::SeaMemoryDef>(ma)) {
+          Expr memOut = m_ctx.mkRegister(*ma);
+          auto *pDefAccess = pMemDef->getDefiningAccess();
+          Expr memIn = m_ctx.mkRegister(*pDefAccess);
+          m_ctx.read(memIn);
+          m_ctx.setMemReadRegister(memIn);
+          m_ctx.setMemWriteRegister(memOut);
+          m_ctx.setMemScalar(false /* TODO: Make this work */);
+        } else {
+          LOG("opsem", errs()
```
agurfinkel, line 685: assert, same as before

lib/seahorn/BvOpSem2.cc
```diff
@@ -1938,8 +2059,24 @@ Expr Bv2OpSemContext::mkPtrRegisterSort(const GlobalVariable &gv) const {
...
+  return mem().mkMemRegisterSort(value);
+}
+
+// Load/store instructions may point to a global
+// or be a global. These are called scalars.
+Expr Bv2OpSemContext::mkRegister(const seadsa::SeaMemoryAccess &access) {
+  if (Expr r = getRegister(access)) {
+    return r;
+  }
+  Expr reg;
```
agurfinkel, line 2072: add a comment about handling unique scalars

lib/seahorn/BvOpSem2Context.hh
```diff
@@ -553,7 +554,7 @@ public:
   virtual PtrSortTy mkPtrRegisterSort(const GlobalVariable &gv) const = 0;
 
   /// \brief Returns sort of memory-holding register for an instruction
-  virtual MemSortTy mkMemRegisterSort(const Instruction &inst) const = 0;
+  virtual Expr mkMemRegisterSort(const Value &value) const = 0;
```
agurfinkel, line 557: return type should be `MemSortTy`. Comment needs to be updated because it talks about instruction.
I would prefer to to not have `mkMemRegisterSort` for `Value`. It is easy to abuse and get wrong

agurfinkel: Ok, I see why you are using `Value`. Add assertions to make sure that only `MemSSA` or `Instruciton` are allowed as arguments and nothing else.

PR #305, opened 2020-06-06
feat(opsem): wire up is_deref(..) and add option to ignore alignment

priyasiddharth:
Also add tests and integrate them into CI and coverage.

### DIFF
--- a/lib/seahorn/BvOpSem2MemRepr.cc
+++ b/lib/seahorn/BvOpSem2MemRepr.cc
@@ -55,9 +55,9 @@ Expr OpSemMemArrayRepr::MemCpy(Expr dPtr, Expr sPtr, unsigned len,
       res = op::array::store(res, dIdx, val);
     }
   } else {
-    LOG("opsem.array", errs() << "Word size and pointer are not aligned and "
-                                 "alignment is not ignored!"
-                              << "\n");
+    LOG("opsem.array", ERR << "Word size and pointer are not aligned and "
+                              "alignment is not ignored!"
+                           << "\n");
     assert(false);
   }
   return res;
--- a/lib/seahorn/BvOpSem2MemRepr.cc
+++ b/lib/seahorn/BvOpSem2MemRepr.cc
@@ -43,7 +43,8 @@ Expr OpSemMemArrayRepr::MemCpy(Expr dPtr, Expr sPtr, unsigned len,
   Expr res;
 
   if (wordSzInBytes == 1 || (wordSzInBytes == 4 && align == 4) ||
-      (wordSzInBytes == 8 && (align == 4 || align == 8))) {
+      (wordSzInBytes == 8 && (align == 4 || align == 8)) ||
+      m_memManager.isIgnoreAlignment()) {
     Expr srcMem = memTrsfrRead;
     res = srcMem;
     for (unsigned i = 0; i < len; i += wordSzInBytes) {
@@ -54,7 +55,8 @@ Expr OpSemMemArrayRepr::MemCpy(Expr dPtr, Expr sPtr, unsigned len,
       res = op::array::store(res, dIdx, val);
     }
   } else {
-    LOG("opsem.array", errs() << "Word size and pointer are not aligned!"
+    LOG("opsem.array", errs() << "Word size and pointer are not aligned and "
+                                 "alignment is not ignored!"
                               << "\n");
     assert(false);
   }
@@ -132,7 +134,8 @@ Expr OpSemMemLambdaRepr::MemCpy(Expr dPtr, Expr sPtr, unsigned len,
   Expr res;
 
   if (wordSzInBytes == 1 || (wordSzInBytes == 4 && align == 4) ||
-      (wordSzInBytes == 8 && (align == 4 || align == 8))) {
+      (wordSzInBytes == 8 && (align == 4 || align == 8)) ||
+      m_memManager.isIgnoreAlignment()) {
     Expr srcMem = memTrsfrRead;
 
     if (len > 0) {
@@ -158,7 +161,8 @@ Expr OpSemMemLambdaRepr::MemCpy(Expr dPtr, Expr sPtr, unsigned len,
       res = memTrsfrRead;
     }
   } else {
-    LOG("opsem.lambda", errs() << "Word size and pointer are not aligned!"
+    LOG("opsem.lambda", errs() << "Word size and pointer are not aligned and "
+                                  "alignment is not ignored!"
                                << "\n");
     assert(false);
   }
--- a/lib/seahorn/BvOpSem2RawMemMgr.cc
+++ b/lib/seahorn/BvOpSem2RawMemMgr.cc
@@ -636,6 +636,11 @@ Expr RawMemManager::setFatData(PtrTy p, unsigned SlotIdx, Expr data) {
   return Expr();
 }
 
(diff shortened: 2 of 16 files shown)


lib/seahorn/BvOpSem2MemRepr.cc
```diff
@@ -152,7 +156,15 @@ Expr OpSemMemLambdaRepr::MemCpy(Expr dPtr, Expr sPtr, unsigned len,
       Expr decl = bind::fname(addr);
       res = mk<LAMBDA>(decl, ite);
       LOG("opsem.lambda", errs() << "MemCpy " << *res << "\n");
+    } else {
+      // result is the same as source
+      res = memTrsfrRead;
     }
+  } else {
+    LOG("opsem.lambda", errs() << "Word size and pointer are not aligned and "
```
agurfinkel, line 164: Good place to use `ERR`

lib/seahorn/BvOpSem2MemRepr.cc
```diff
@@ -152,7 +156,15 @@ Expr OpSemMemLambdaRepr::MemCpy(Expr dPtr, Expr sPtr, unsigned len,
...
+                                  "alignment is not ignored!"
+                               << "\n");
+    assert(false);
```
agurfinkel, line 167: what happens in release mode when `assert(false)` is ignored?

lib/seahorn/BvOpSem2RawMemMgr.cc
```diff
@@ -621,6 +636,11 @@ Expr RawMemManager::setFatData(PtrTy p, unsigned SlotIdx, Expr data) {
   return Expr();
 }
 
+Expr RawMemManager::isDereferenceable(PtrTy p, Expr byteSz) {
+  LOG("opsem", WARN << "isDeferenceable() not implemented");
```
agurfinkel, line 640: Maybe `ERR` ?

lib/seahorn/BvOpSem2RawMemMgr.cc
```diff
@@ -674,6 +694,9 @@ void RawMemManager::onFunctionEntry(const Function &fn) {
 Expr RawMemManager::zeroedMemory() const {
   return m_memRepr->FilledMemory(ptrSort(), m_ctx.alu().si(0, wordSzInBits()));
 }
-
+const std::unique_ptr<OpSemAllocator> &RawMemManager::getMAllocator() const {
```
agurfinkel, line 697: probably wnat to return `OpSemAllocator&` rather than ref to unique pointer

agurfinkel:
looks good, but some checks are failing. The PR is a draft so I can't merge it

agurfinkel:
appears that install target is broken. Either missing file or a typo. Can you please fix it so that CI can restart

lib/seahorn/BvOpSem2MemRepr.cc
priyasiddharth: done

lib/seahorn/BvOpSem2MemRepr.cc
priyasiddharth: We return expr() in release mode.

lib/seahorn/BvOpSem2RawMemMgr.cc
priyasiddharth: done

lib/seahorn/BvOpSem2RawMemMgr.cc
priyasiddharth: done

priyasiddharth:
Done.

agurfinkel:
ready to rebase?

priyasiddharth:
Yes.

lib/seahorn/BvOpsem2WideMemMgr.cc
```diff
@@ -314,10 +316,7 @@ class WideMemManager : public OpSemMemManagerBase {
       return ptrAdd(ptr, _offset.get_si());
     }
     auto address = m_ctx.alu().doAdd(ptr.getRaw(), offset, ptrSzInBits());
-    auto new_size = m_ctx.alu().doSub(
-        ptr.getSize(),
-        m_ctx.alu().doZext(offset, g_slotBitWidth, m_main.wordSzInBits()),
-        g_slotBitWidth);
+    auto new_size = m_ctx.alu().doSub(ptr.getSize(), offset, g_slotBitWidth);
```
agurfinkel, line 319: if `ptrSzInBits()  != g_slotBitWidth` then this will fail because `offset` must be of one width or the other.
Should use `sext` to sign extended offset (offset can be negative) or trunc offset to fit the other bit-width.

I believe offset must be same bitwidth as `ptrSzInBits()` so it might require shrinking of `g_slotBitWidth` is smaller.

lib/seahorn/BvOpsem2WideMemMgr.cc
```diff
@@ -504,10 +503,7 @@ class WideMemManager : public OpSemMemManagerBase {
   PtrTy gep(PtrTy ptr, gep_type_iterator it, gep_type_iterator end) const {
     RawPtrTy rawPtr = m_main.gep(ptr.getRaw(), it, end);
     auto offset = m_main.ptrSub(rawPtr, ptr.getRaw());
-    auto new_size = m_ctx.alu().doSub(
-        ptr.getSize(),
-        m_ctx.alu().doZext(offset, g_slotBitWidth, m_main.wordSzInBits()),
-        g_slotBitWidth);
+    auto new_size = m_ctx.alu().doSub(ptr.getSize(), offset, g_slotBitWidth);
```
agurfinkel, line 506: flagged for efficiency. Symbolically, things will be easier if offset is computed first and then added to both raw part and subtracted from size. Currently, the expressions that are generated will be something like:
```
rawPtr = ptr + offset
newSize = size - (ptr - rawPtr)
```
There is no automatic simplification. So even if size was numeric and offset is numeric, the new size is not numeric.

lib/seahorn/BvOpsem2WideMemMgr.cc
```diff
@@ -542,13 +538,29 @@ class WideMemManager : public OpSemMemManagerBase {
     LOG("opsem", WARN << "setFatData() not implemented!\n");
     return nullPtr();
   }
+
+  Expr isDereferenceable(PtrTy p, Expr byteSz) {
```
agurfinkel, line 542: add also a version with the signature `Expr isDereferenceable(PtrTy p, unsigned bytes)` and re-direct to it. We might be able to support numeric ranges differently than symbolic.

lib/seahorn/BvOpsem2WideMemMgr.cc
```diff
@@ -542,13 +538,29 @@ class WideMemManager : public OpSemMemManagerBase {
...
+    // size should be >= byteSz
+    if (m_ctx.alu().isNum(byteSz) && m_ctx.alu().isNum(p.getSize())) {
+      signed numBytes = m_ctx.alu().toNum(byteSz).get_si();
```
agurfinkel, line 545: these should be unsigned. Perhaps we need `get_ui()` in ALU for that.
Probably does not matter in practice since we expected these sizes to be fairly small

agurfinkel:
Looks great. 

I've left a few comments. Most indicate potential problems that are likely not to materialize. 
One is an important performance issue (handling of `gep`). 
I will rebase this into `dev10` and let you submit a new PR addressing those changes.

PR #307, opened 2020-06-08
Support --devirt-functions-allow-incomplete with sea-dsa

caballa:


agurfinkel:
looks good. should I rebase?

PR #309, opened 2020-06-16
feat(opsem): add an extra-wide memory manager which tracks (base, off…

priyasiddharth:
…set, size)

details:
1. Moved some common functions between {,extra} wide memory manager to a wide mem mgr mixin class by introducing primitives like getAddressable(). The refactor is not complete since I wanted to get something out soon. This is a prototype of where I want to go eventually.
2. Fixed a flag in opsem/widemem tests and an actual bug in wide memory manager. The fact that no-one has noticed may mean that tests are not running in CI. That problem is out of scope for this commit.
3. Replaced test-widemem flag with test-opsem-extra-widemem since tests in subdirectory opsem/widemem are automatically run as part of test-opsem. Now we have three targets 1) test-opsem 2) test-opsem-widemem and 3) test-opsem-extra-widemem.
4. Fix widememmgr name


agurfinkel:
fantastic

PR #310, opened 2020-06-19
Speedup gate SSA transformation.

parvmor:
The commit speeds up the gate ssa transformation. Instead of iterating over the `flowValues` and checking the dominance relation we can iterate over the nodes dominators and check if `flowValues` has a corresponding value.

As an example, for the SVCOMP-19 benchmark instance `c/*/linux-3.14__complex_emg__linux-alloc-spinlock__drivers-media-usb-usbvision-usbvision_true-unreach-call.cil.i`, the change reduces the transformation time from ~250 seconds to ~12 seconds.

Question: What I do not understand is why the `flowValues` has an entry for a post-dominator that isn't immediate but not for the immediate post-dominator. We are walking over `cdInfo` is reverse topological order!

### DIFF
--- a/lib/Analysis/GateAnalysis.cc
+++ b/lib/Analysis/GateAnalysis.cc
@@ -20,6 +20,7 @@
 #include "seahorn/Analysis/ControlDependenceAnalysis.hh"
 
 #include "seahorn/Support/SeaDebug.h"
+#include "seahorn/Support/Stats.hh"
 
 #define GSA_LOG(...) LOG("gsa", __VA_ARGS__)
 
@@ -149,9 +150,7 @@ GateAnalysisImpl::processIncomingValues(PHINode *PN, Instruction *insertionPt) {
     BasicBlock *falseDest = BI->getSuccessor(1);
     assert(trueDest == currentBB || falseDest == currentBB);
 
-    if (ThinnedGsa) {
-      incomingBlockToValue[incomingBlock] = incomingValue;
-    } else {
+    if (!ThinnedGsa) {
       Value *SI = m_IRB.CreateSelect(
           cond, trueDest == currentBB ? incomingValue : Undef,
           falseDest == currentBB ? incomingValue : Undef,
@@ -171,23 +170,21 @@ void GateAnalysisImpl::processPhi(PHINode *PN, Instruction *insertionPt) {
   DenseMap<BasicBlock *, Value *> incomingBlockToValue =
       processIncomingValues(PN, insertionPt);
 
+  // Make sure CD blocks are sorted in reverse-topological order. We need this
+  // because we want to process them in order opposite to execution order.
+  auto GreaterThanTopo = [this](BasicBlock *first, BasicBlock *second) {
+    return m_CDA.getBBTopoIdx(first) > m_CDA.getBBTopoIdx(second);
+  };
+  // TODO: Look into the performance effects of using a set wrt a vector.
+  // For now we use a set since, the number of redundant blocks are huge. Using
+  // a set results in a smaller time complexity.
+  std::set<BasicBlock *, decltype(GreaterThanTopo)> cdInfo(GreaterThanTopo);
   // Collect all blocks the incoming blocks are control dependent on.
-  std::vector<BasicBlock *> cdInfo;
   for (unsigned i = 0, e = PN->getNumIncomingValues(); i != e; ++i) {
     auto *BB = PN->getIncomingBlock(i);
     for (BasicBlock *CDBlock : m_CDA.getCDBlocks(BB))
-      cdInfo.push_back(CDBlock);
+      cdInfo.insert(CDBlock);
   }
-
-  // Make sure CD blocks are sorted in reverse-topological order. We need this
-  // because we want to process them in order opposite to execution order.
-  std::sort(cdInfo.begin(), cdInfo.end(),
-            [this](BasicBlock *first, BasicBlock *second) {
-              return m_CDA.getBBTopoIdx(first) > m_CDA.getBBTopoIdx(second);
-            });
-  // We can have repeated blocks if multiple incoming blocks have non-empty
-  // intersections of blocks they are control dependent on.
-  cdInfo.erase(std::unique(cdInfo.begin(), cdInfo.end()), cdInfo.end());
 
   // Mapping from blocks in cdInfo to values potentially guarded by gammas.
   DenseMap<BasicBlock *, Value *> flowingValues(incomingBlockToValue.begin(),
@@ -199,7 +196,6 @@ void GateAnalysisImpl::processPhi(PHINode *PN, Instruction *insertionPt) {
 
   // For all blocks in cdInfo and inspect their successors to construct gamma
   // nodes where needed.
-  unsigned cdNum = 0;
   GSA_LOG(errs() << "CDInfo size: " << cdInfo.size() << "\n");
   for (BasicBlock *BB : cdInfo) {
     GSA_LOG(errs() << "CDBlock: " << BB->getName() << "\n");
@@ -230,15 +226,23 @@ void GateAnalysisImpl::processPhi(PHINode *PN, Instruction *insertionPt) {
 
       // Or the successors unconditionally flows to an already processed block.
       // Note that there can be at most one such block.
-      for (auto &BlockValuePair : flowingValues) {
-        if (m_PDT.dominates(BlockValuePair.first, S)) {
-          SuccToVal[S] = BlockValuePair.second;
+
+      // Used to iterate over all the post-dominators of `S`.
+      BasicBlock *postDomBlock = S;
+      while (postDomBlock) {
+        auto it = flowingValues.find(postDomBlock);
+        if (it != flowingValues.end()) {
+          SuccToVal[S] = it->second;
           GSA_LOG(errs() << "2) SuccToVal[" << S->getName()
-                         << "] = postdom for cd "
-                         << BlockValuePair.first->getName() << ": "
-                         << SuccToVal[S]->getName() << "\n");
+                         << "] = postdom for cd " << postDomBlock->getName()
+                         << ": " << it->second->getName() << "\n");
           break;
         }
+        auto *treeNode = m_PDT.getNode(postDomBlock)->getIDom();
+        if (treeNode == nullptr) {
+          break;
+        }
+        postDomBlock = treeNode->getBlock();
       }
     }
 
@@ -267,8 +271,6 @@ void GateAnalysisImpl::processPhi(PHINode *PN, Instruction *insertionPt) {
         flowingValues[BB] = Ite;
       }
     }
-
-    ++cdNum;
   }
 
   BasicBlock *IDomBlock = m_DT.getNode(currentBB)->getIDom()->getBlock();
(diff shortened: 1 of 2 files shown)


lib/Analysis/GateAnalysis.cc
```diff
@@ -171,24 +170,19 @@ void GateAnalysisImpl::processPhi(PHINode *PN, Instruction *insertionPt) {
...
       processIncomingValues(PN, insertionPt);
 
+  // Make sure CD blocks are sorted in reverse-topological order. We need this
+  // because we want to process them in order opposite to execution order.
+  auto bb_to_compare = [this](BasicBlock *first, BasicBlock *second) {
+    return m_CDA.getBBTopoIdx(first) > m_CDA.getBBTopoIdx(second);
+  };
   // Collect all blocks the incoming blocks are control dependent on.
-  std::vector<BasicBlock *> cdInfo;
+  std::set<BasicBlock *, decltype(bb_to_compare)> cdInfo(bb_to_compare);
```
agurfinkel, line 179: If I interpret this code correctly, `std::set` is used to ensure that collected basic blocks are sorted.
It is more efficient and simpler to collect basic blocks first into a vector and then sort them. As the previous code was doing.

lib/Analysis/GateAnalysis.cc
```diff
@@ -230,15 +223,32 @@ void GateAnalysisImpl::processPhi(PHINode *PN, Instruction *insertionPt) {
...
       // Or the successors unconditionally flows to an already processed block.
       // Note that there can be at most one such block.
-      for (auto &BlockValuePair : flowingValues) {
-        if (m_PDT.dominates(BlockValuePair.first, S)) {
-          SuccToVal[S] = BlockValuePair.second;
-          GSA_LOG(errs() << "2) SuccToVal[" << S->getName()
-                         << "] = postdom for cd "
-                         << BlockValuePair.first->getName() << ": "
-                         << SuccToVal[S]->getName() << "\n");
+      bool found = false;
```
agurfinkel, line 226: @kuhar can you review this part?

lib/Analysis/GateAnalysis.cc
caballa: Yes, this is what it's doing the code

parvmor: Not really. The examples I ran were slower with using vector, sorting and then deleting redundant entries.

agurfinkel:
@parvmor thank you! This looks good. I've asked @kuhar to review the logic change. I will do the same soon.

Can you please comment inside the code on the algorithm and the difference between the current and previous implementation. A small example would be much appreciated.

parvmor:
Sure! While looking at the logic can you please also look into the question I mentioned earlier?

lib/Analysis/GateAnalysis.cc
agurfinkel: Hmmm... this does not make sense. From complexity perspective, both algorithms are identical. Using a properly sized vector and sorting requires fewer memory allocations. I would suggest profiling the code to see where the time is going.

http://lafstern.org/matt/col1.pdf
see also `boost::flat_set`

agurfinkel:
@kuhar maybe you can answer the question as well: 

> Question: What I do not understand is why the `flowValues` has an entry for a post-dominator that isn't immediate but not for the immediate post-dominator. We are walking over `cdInfo` in reverse topological order!

lib/Analysis/GateAnalysis.cc
parvmor: So for the examples I ran, the vector initially holds around 700,000 entries (`m`) and de-duplication reduces the number of entries to around 4000 (`n`). With a vector the expected complexity is now `m log(m)` and with a set it is `m log(n)`.

agurfinkel: `flat_set` would be a terrible choice here since the insertions are not in order.

parvmor: Ahh, just saw the insertion time complexity.

agurfinkel: I am surprised at the number of blocks. If it is that high, you should mark the blocks that are entered in the set and skip adding them a second time. In which case using `set` is more justified. At least a comment to that effect would be welcome so that we can evaluate this in the future.

parvmor: > you should mark the blocks that are entered in the set and skip adding them a second time

I can do this but can you explain how does this helps? Checking if the element is already present should have the same complexity as inserting an already existing element. Maybe the confusion is between a `set` and `multiset`.

parvmor: Also, maybe you would prefer this more: Maintain a hash set to avoid redundant entries and then convert it into a vector and sort it in reverse topological order.

agurfinkel: since basic blocks have identifiers, I would have used a bit vector. But at that point, your use of `std::set` is justified.

lib/Analysis/GateAnalysis.cc
```diff
@@ -171,24 +170,19 @@ void GateAnalysisImpl::processPhi(PHINode *PN, Instruction *insertionPt) {
   DenseMap<BasicBlock *, Value *> incomingBlockToValue =
       processIncomingValues(PN, insertionPt);
 
+  // Make sure CD blocks are sorted in reverse-topological order. We need this
+  // because we want to process them in order opposite to execution order.
+  auto bb_to_compare = [this](BasicBlock *first, BasicBlock *second) {
```
kuhar, line 175: Nit: use ThisCase for lambdas and give it a more descriptive name, e.g., `s/bb_to_compare/GreaterThanTopo`

lib/Analysis/GateAnalysis.cc
kuhar: Since both pointers and topo indices are unique, yo can deduplicate the values first with a `DenseSet` and then copy it over to a vector and sort. But this shouldn't matter much.

lib/Analysis/GateAnalysis.cc
```diff
@@ -230,15 +223,32 @@ void GateAnalysisImpl::processPhi(PHINode *PN, Instruction *insertionPt) {
...
       // Note that there can be at most one such block.
-      for (auto &BlockValuePair : flowingValues) {
-        if (m_PDT.dominates(BlockValuePair.first, S)) {
-          SuccToVal[S] = BlockValuePair.second;
-          GSA_LOG(errs() << "2) SuccToVal[" << S->getName()
-                         << "] = postdom for cd "
-                         << BlockValuePair.first->getName() << ": "
-                         << SuccToVal[S]->getName() << "\n");
+      bool found = false;
+      BasicBlock *bb =  S;
```
kuhar, line 227: nit: Can you give more descriptive names to the variables here and format the code with clang-format?

lib/Analysis/GateAnalysis.cc
```diff
@@ -230,15 +223,32 @@ void GateAnalysisImpl::processPhi(PHINode *PN, Instruction *insertionPt) {
...
+      auto *node = m_PDT.getNode(bb)->getIDom();
+      while (node != nullptr) {
```
kuhar, line 229: nit: `while (node)`

lib/Analysis/GateAnalysis.cc
```diff
@@ -230,15 +223,32 @@ void GateAnalysisImpl::processPhi(PHINode *PN, Instruction *insertionPt) {
...
+        auto it = flowingValues.find(bb);
+        if (it != flowingValues.end()) {
```
kuhar, line 231: nit: `if (flowingValues.count(bb) == 0)`

lib/Analysis/GateAnalysis.cc
```diff
@@ -230,15 +223,32 @@ void GateAnalysisImpl::processPhi(PHINode *PN, Instruction *insertionPt) {
...
+          SuccToVal[S] = it->second;
+          found = true;
           break;
         }
+        bb = node->getBlock();
+        node = m_PDT.getNode(bb)->getIDom();
+      }
+      if (node == nullptr) {
```
kuhar, line 239: Can this loop be restructured so that we don't have to manually do one more iteration at the end? It seems you only really need to collect all the basic blocks along the postdom chain.

lib/Analysis/GateAnalysis.cc
kuhar: and also in the code below

lib/Analysis/GateAnalysis.cc
kuhar: Can you comment what this case means? `node == nullptr` should happen when you traverse the postdominance chain all the way up to the virtual root, right?

kuhar:
Nice. I think the gist of this change is that you made an observation that there are usually fewer postdominator nodes than the size of flowing values (O(|F|)) ands it's faster to interchange the checks to go over postdominators first.

kuhar:
> @kuhar maybe you can answer the question as well:
> 
> > Question: What I do not understand is why the `flowValues` has an entry for a post-dominator that isn't immediate but not for the immediate post-dominator. We are walking over `cdInfo` in reverse topological order!

IIRC the intention was to propagate flowing values downwards, so that the values are final when you reach the processed phi node. I don't remember the details, but I think that if an immediate postdominator of a block is the final value, then this implies `case 1)` in the code.

agurfinkel:
@kuhar thank you!

lib/Analysis/GateAnalysis.cc
parvmor: I don't understand the benefit of this. This will require two accesses to `flowingValues` in case the value is already present: first to check if the value is present and second to get the value.

lib/Analysis/GateAnalysis.cc
parvmor: Changed the loop condition and removed the condition. Shouldn't matter anymore.

lib/Analysis/GateAnalysis.cc
parvmor: Added a comment to reflect the use case of set.

lib/Analysis/GateAnalysis.cc
kuhar: right, I missed the use below

lib/Analysis/GateAnalysis.cc
```diff
@@ -230,15 +226,23 @@ void GateAnalysisImpl::processPhi(PHINode *PN, Instruction *insertionPt) {
 
       // Or the successors unconditionally flows to an already processed block.
       // Note that there can be at most one such block.
-      for (auto &BlockValuePair : flowingValues) {
-        if (m_PDT.dominates(BlockValuePair.first, S)) {
-          SuccToVal[S] = BlockValuePair.second;
+
+      // Used to iterate over all the post-dominators of `S`.
+      BasicBlock *post_dominator = S;
```
kuhar, line 231: nit: seahorn used thisCase for local variable names.
Consider a less generic name, e.g., `postDomBlock`.

lib/Analysis/GateAnalysis.cc
```diff
@@ -230,15 +226,23 @@ void GateAnalysisImpl::processPhi(PHINode *PN, Instruction *insertionPt) {
...
+          SuccToVal[S] = it->second;
           GSA_LOG(errs() << "2) SuccToVal[" << S->getName()
-                         << "] = postdom for cd "
-                         << BlockValuePair.first->getName() << ": "
-                         << SuccToVal[S]->getName() << "\n");
+                         << "] = postdom for cd " << post_dominator->getName()
+                         << ": " << it->second->getName() << "\n");
+          break;
+        }
+        auto *tree_node = m_PDT.getNode(post_dominator)->getIDom();
```
kuhar, line 241: nit: treeNode

parvmor:
I am not sure which one is the current developmental branch. @agurfinkel should be able to change the base branch for the commit.

caballa:
I'll merge it into master and then I can port it to `dev10`

caballa:
@parvmor: thank you very much!

PR #315, opened 2020-06-26
Widemem - fixes to run trusty

priyasiddharth:


include/seahorn/Expr/Smt/Z3.hh
```diff
@@ -482,6 +494,20 @@ public:
     Expr res = z3.toExpr(z3::ast(ctx, Z3_simplify_ex(ctx, ast, m_params)),
                          m_ast_to_expr);
     m_cache.insert({e, res});
+    LOG(
```
agurfinkel, line 497: the LOG messages do not belong here. They are strictly for bvopsem2

lib/seahorn/BvOpSem2.cc
```diff
@@ -530,7 +530,9 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemVisitorBase {
       auto *val = CS.getArgument(2);
       Expr symVal = lookup(*val);
       if (symVal && arg0 && arg1) {
-        res = m_ctx.alu().Extract({symVal, val->getType()->getScalarSizeInBits()}, arg0->getZExtValue(), arg1->getZExtValue());
+        res =
```
agurfinkel, line 533: space issue

lib/seahorn/BvOpSem2.cc
```diff
@@ -1765,6 +1767,13 @@ Bv2OpSemContext::Bv2OpSemContext(Bv2OpSem &sem, SymStore &values,
   zeroE = mkTerm<expr::mpz_class>(0UL, efac());
   oneE = mkTerm<expr::mpz_class>(1UL, efac());
 
+  m_z3.reset(new EZ3(efac()));
+  m_z3_simplifier.reset(new ZSimplifier<EZ3>(*m_z3));
+  auto &params = m_z3_simplifier->params();
+  params.set("ctrl_c", true);
+  if (SimplifyExpr) {
+    m_shouldSimplify = true;
```
agurfinkel, line 1775: `m_shouldSimplify = SimplifyExpr;`

lib/seahorn/BvOpSem2.cc
```diff
@@ -1794,48 +1803,12 @@ Bv2OpSemContext::Bv2OpSemContext(SymStore &values, ExprVector &side,
...
-    if (strct::isStructVal(u)) {
-      llvm::SmallVector<Expr, 8> kids;
-      for (unsigned i = 0, sz = u->arity(); i < sz; ++i)
-        kids.push_back(m_z3_simplifier->simplify(u->arg(i)));
-      _u = strct::mk(kids);
-    } else {
-      _u = m_z3_simplifier->simplify(u);
-    }
-
-    LOG(
```
agurfinkel, line 1823: these log messages have to be moved back from Z3.hh to here

lib/seahorn/BvOpSem2Context.hh
```diff
@@ -110,6 +112,11 @@ public:
   ~Bv2OpSemContext() override = default;
 
   EZ3 *getZ3() const { return m_z3.get(); }
+  std::shared_ptr<ZSimplifier<EZ3>> getSimplifier() {
```
agurfinkel, line 115: Most of the time you would want to return `ZSimplifier<EZ3> &` 

What is the reason for sharing ownership with the caller?

lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
```diff
@@ -544,6 +547,11 @@ class ExtraWideMemManager : public OpSemMemManagerBase {
 
   Expr isDereferenceable(PtrTy p, Expr byteSz) {
     // size should be >= byteSz + offset
+    if (m_ctx.shouldSimplify()) {
+      ScopedStats _st_("opsem.simplify");
+      p = PtrTy(m_ctx.getSimplifier()->simplify(p.toExpr()));
```
agurfinkel, line 552: does this simplification has a big effect? 

I'd imagine bigger effect on size and offset portion

agurfinkel:
Needs a few tweaks before I can accept it.

include/seahorn/Expr/Smt/Z3.hh
priyasiddharth: I have moved them back... can you please let me know why though?

lib/seahorn/BvOpSem2.cc
priyasiddharth: This has been auto reformatted by the editor(vscode) using clang-format. If this is a issue with not following clang-format, I can fix it.

lib/seahorn/BvOpSem2.cc
priyasiddharth: done!

lib/seahorn/BvOpSem2Context.hh
priyasiddharth: Fixed... may i know the reason for making it a shared_ptr in the first place... who is an entity we want to share the simplifier with?

lib/seahorn/BvOpSem2.cc
agurfinkel: ok. maybe it just looks strange in the web interface

lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
priyasiddharth: Good point! I fixed to only simplify offset, size sampling we don't seem to be spending much time in `simplify` but rather `solver::check_sat()`

lib/seahorn/BvOpSem2Context.hh
agurfinkel: Not really sure. Maybe there was some issue with copy constructor or something like that. We can try moving it to `std::unique_ptr` instead.

priyasiddharth:
Ok to review... once if passes the presubmit checks.

include/seahorn/Expr/Smt/Z3.hh
agurfinkel: the logs are specific for debugging bvopsem. They are of no use for other uses of the simplifier

PR #316, opened 2020-07-03
Fixes in seainspect and vcgen

priyasiddharth:
VSCode seems to be autofixing things without being asked... This looks reasonable in the present case but I will see if I can get it to fix only modified code.

Let me know if you want me to revert this and send another PR with auto format disabled

### DIFF
--- a/include/seahorn/Expr/ExprOpStruct.hh
+++ b/include/seahorn/Expr/ExprOpStruct.hh
@@ -2,6 +2,7 @@
 #pragma once
 #include "seahorn/Expr/ExprApi.hh"
 #include "seahorn/Expr/ExprCore.hh"
+#include "seahorn/Expr/ExprOpBinder.hh"
 #include "seahorn/Expr/ExprOpBool.hh"
 #include "seahorn/Expr/ExprOpCore.hh"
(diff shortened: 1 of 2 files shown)


include/seahorn/Expr/ExprOpStruct.hh
```diff
@@ -138,7 +138,7 @@ inline Expr push_ite_struct(Expr c, Expr lhs, Expr rhs) {
 
   llvm::SmallVector<Expr, 8> vals;
   for (unsigned i = 0, sz = lhs->arity(); i < sz; ++i) {
-    vals.push_back(boolop::lite(c, lhs->arg(i), rhs->arg(i)));
+    vals.push_back(bind::lite(c, lhs->arg(i), rhs->arg(i)));
```
agurfinkel, line 141: this is an ok fix for current problem, but I am afraid that we might not be handling all combinations of structs and lambdas. For instance, `bind::lite` does not call `push_ite_struct`, so creating an `ite` over a `lambda` that returns a `struct` will not be handled correctly.

agurfinkel:
I am surprised that commit linter passed these commit messages. They don't strictly follow the format -- capital letters and length.

PR #318, opened 2020-07-10
fix(opsem): gep of extra wide mem mgr did not account for nested gep

priyasiddharth:


agurfinkel:
I will merge after PR completes
Would be good to add a test for this as well.

PR #322, opened 2020-07-27
feat(opsem): add fpf command which adds fat-bnds-check instrumentatio…

priyasiddharth:
…n to bpf workflow

fpf=FrontEnd.cmds + [FatBoundsCheck(), Unroll(), CutLoops(), Seaopt(), Seahorn(solve=True)])

### DIFF
--- a/py/sea/commands.py
+++ b/py/sea/commands.py
@@ -894,6 +894,48 @@ class Seaopt(sea.LimitedCmd):
         if args.llvm_asm: argv.append ('-S')
         return self.seaoptCmd.run (args, argv)
 
+
+class FatBoundsCheck(sea.LimitedCmd):
+    def __init__(self, quiet=False):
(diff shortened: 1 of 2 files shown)


py/sea/commands.py
```diff
@@ -1632,3 +1674,5 @@ def run(self, args, extra):
 Smc = sea.SeqCmd ('smc', 'alias for fe|opt|smc',
                    [Clang(), Seapp(), SimpleMemoryChecks(), MixedSem(),
                     Seaopt(), Seahorn(solve=True)])
+fpf = sea.SeqCmd('fpf', 'fe|fat-bnd-check|unroll|cut-loops|opt|horn --solve',
```
agurfinkel, line 1677: please rename into `Fpf` to be consistent with other variable names

priyasiddharth: done

PR #332, opened 2020-08-04
Expression evaluator

colinxm:


include/seahorn/Expr/EvalModel.hh
```diff
@@ -0,0 +1,198 @@
...
+#include "seahorn/Expr/ExprOpBind.hh"
+#include "seahorn/Expr/ExprOpBinder.hh"
+
+#include <math.h>
+#include <stdlib.h>
+
+namespace expr {
+namespace eval {
+
+template <typename T> struct BvNum {
```
agurfinkel, line 15: add comment with intended meaning

agurfinkel:
This fantastic. The only request I have is to add comments to indicate the intention of each class and method

include/seahorn/Expr/EvalModel.hh
agurfinkel: shouldn't width be not always stored as part of `BvNum`? Some representation of `T` might have width already stored in them

include/seahorn/Expr/EvalModel.hh
```diff
@@ -0,0 +1,302 @@
...
+ * Holds the data for an array.
+ *
+ * In EvalModel, the array expressions are mapped to this struct
+ *
+ */
+template <typename T> struct Arr {
+  bool m_hasDefaultVal = false;
+  BvNum<T> m_defaultVal;
+
+  unsigned m_valWidth = 0;
```
agurfinkel, line 49: what is `m_valWidth`? isn't width already in `BvNum`?

include/seahorn/Expr/EvalModel.hh
```diff
@@ -0,0 +1,302 @@
...
+
+  bool operator==(const BvNum<T> &bnum) const {
+    return (num == bnum.num) && (width == bnum.width);
+  }
+};
+
+/**
+ * Holds the data for an array.
+ *
+ * In EvalModel, the array expressions are mapped to this struct
```
agurfinkel, line 42: a bit more about how arrays are represented.

include/seahorn/Expr/EvalModel.hh
```diff
@@ -0,0 +1,302 @@
...
+        m_valWidth(defaultVal.width) {}
+  Arr(unsigned valWidth) : m_hasDefaultVal(false), m_valWidth(valWidth) {}
+};
+
+/**
+ * Holds the data for a lambda. Maps bound vars to their value
+ *
+ * The purpose of this is to keep bound vars within the scope of the lambda. The
+ * same bound var expression could have different values in different lambdas
+ *
```
agurfinkel, line 64: In general, lambda is a function. What does that mean for `bound var` to have different values?
Say I have lambda expressions of this form : `[](bvnum x) { return x + 1; }`, where `x` is the bound variable.
I then get stuck trying to parse the comment and apply it to the lambda above.

include/seahorn/Expr/EvalModel.hh
```diff
@@ -0,0 +1,302 @@
...
+ * \note the get methods generate a value if one does not already exist
+ *
+ *  To create a new model, extend EvalModel, and override generateNum(). See
+ *  EvalModel_Rand for an example
+ */
+template <typename Type> class EvalModel {
+
+protected:
+  /// maps the constants/bvars to their values
+  std::map<Expr, BvNum<Type>> m_cache;
```
agurfinkel, line 87: is this `cache` or `data`? Cache is usually something that can be re-computed and might become stale.
I think this is the actual data that underpins the model

include/seahorn/Expr/EvalModel.hh
```diff
@@ -0,0 +1,302 @@
...
+    assert(isOp<LAMDBA>(lambda));
+    Lambda<Type> l;
+    m_lambdaCache.insert({lambda, l});
+  }
+};
+
+/**
+ * \brief Eval model to generate numbers randomly
+ */
+template <typename T> class EvalModel_Rand : public EvalModel<T> {
```
agurfinkel, line 279: I dislike mixing camel and snake case. Let's remove `_`.

include/seahorn/Expr/EvalUtils.hh
```diff
@@ -0,0 +1,43 @@
...
+#include "seahorn/Expr/ExprOpBinder.hh"
+
+#include <math.h>
+#include <stdlib.h>
+
+namespace expr {
+namespace eval {
+namespace evalUtils {
+
+/// removes upper bits
```
agurfinkel, line 16: a few more words. how the bits are removed? zeroed or what?

include/seahorn/Expr/EvalUtils.hh
```diff
@@ -0,0 +1,43 @@
...
+
+  for (int i = 0; i < numBits; i++) {
+    mask = mask | ((((T)1) << i));
+  }
+
+  return num & mask;
+}
+
+/// removes lower bits
+template <typename T> T maskLowerBits(const T &num, unsigned numBits) {
```
agurfinkel, line 28: why is one function called `masked...` another is `overflow` but comment makes seem like they are related.
Maybe use related names as well?

include/seahorn/Expr/EvalModel.hh
colinxm: The purpose of m_valWidth is for generating new BvNums. If you try to generate a new number when there isn't any BvNums stored yet then you won't know how wide to make it

colinxm: Maybe I could just store the width in m_valWidth and store the values only as the number and not bvnum to reduce some overhead?

include/seahorn/Expr/EvalModel.hh
colinxm: I might need to change the wording, I meant the expressions itself. Is it possible that you might have a bound variable expression that is used in more than one lambda? Obviously the value of it would be different depending on where its used but the Expr is the same

agurfinkel: This makes sense. Bound variables are arguments and each lambda can use them in their own way. 

Seems like you only consider lambdas that can be written as a table. This is ok for now

include/seahorn/Expr/EvalModel.hh
```diff
@@ -0,0 +1,367 @@
...
+ * number and provides methods to get/set the numbers width
+ *
+ * To Add a new type of number (a new NumContainer), template specialize
+ * chooseContainer() with the NumContainer
+ *
+ * \note bools are stored as 1/0 with a width of 1
+ */
+template <typename T> struct BvNum {
+private:
+  std::shared_ptr<NumContainerBase<T>> num;
```
agurfinkel, line 71: this is way too complicated and inefficient. I was proposing not keeping `width` in `BvNum<T>`, and then add it in the specialization `BvNum<mpz_class>`.

The current solution is even heavier than the previous one since now every number contains a pointer on top of the width!

colinxm: Ok it should be fixed now. I remove the BvNum<T> struct all together and stored the number directly. I figured if its not storing the width then its no different than just storing the number directly

agurfinkel:
the PR is a draft so I cannot merge it

agurfinkel:
ready to merge. please mark as non draft

colinxm:
> ready to merge. please mark as non draft

sorry about that. Its changed now

PR #338, opened 2020-08-28
refactor(opsem): encapsulate solver in bv2 ctx

priyasiddharth:
1. keep track of assertions during single vacuity and incremental check
2. add printing of debug location if available

### DIFF
--- a/lib/seahorn/BvOpSem2.cc
+++ b/lib/seahorn/BvOpSem2.cc
@@ -2,6 +2,8 @@
 #include "BvOpSem2RawMemMgr.hh"
 
 #include "llvm/CodeGen/IntrinsicLowering.h"
+#include "llvm/IR/DebugInfo.h"
+#include "llvm/IR/DebugLoc.h"
 #include "llvm/IR/GetElementPtrTypeIterator.h"
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/Format.h"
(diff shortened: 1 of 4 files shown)


lib/seahorn/BvOpSem2.cc
```diff
@@ -157,7 +159,6 @@ struct OpSemVisitorBase {
       : m_ctx(ctx), m_efac(m_ctx.getExprFactory()), m_sem(sem) {
 
     trueE = m_ctx.m_trueE;
-    falseE = m_ctx.m_falseE;
```
agurfinkel, line 160: why?

lib/seahorn/BvOpSem2.cc
```diff
@@ -185,6 +186,7 @@ struct OpSemVisitorBase {
 
     if (const Constant *cv = dyn_cast<Constant>(&v)) {
       return m_ctx.getConstantValue(*cv);
+      falseE = m_ctx.m_falseE;
```
agurfinkel, line 189: why? won't execute after return

lib/seahorn/BvOpSem2.cc
```diff
@@ -579,59 +581,85 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemVisitorBase {
...
+                 << "]";
+          } else { INFO << "Assertion passed"; });
     }
   }
+  void visitAssertIf(CallSite CS) {
+    // NOTE: sea.assert.if.not is not supported
+    auto I = CS.getInstruction();
+    Expr ante = lookup(*CS.getArgument(0));
+    Expr conseq = lookup(*CS.getArgument(1));
+    const llvm::DebugLoc &dloc = I->getDebugLoc();
```
agurfinkel, line 645: what happens if I had no debug location?

lib/seahorn/BvOpSem2.cc
```diff
@@ -579,59 +581,85 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemVisitorBase {
...
+    Expr e = boolop::lneg(conseq);
     Stats::resume("opsem.incbmc");
     tribool conseqRes = solveWithConstraints(e);
     Stats::stop("opsem.incbmc");
     if (conseqRes) {
-      LOG("opsem", ERR << "Consequent is sat/unknown: " << *I << "\n");
+      LOG(
+          "opsem",
+          if (dloc) {
+            ERR << "Consequent is sat/unknown: "
```
agurfinkel, line 627: this is only reachable when Sat but prints that might be unknown

lib/seahorn/BvOpSem2.cc
```diff
@@ -2254,6 +2282,22 @@ std::pair<char *, unsigned>
...
   return m_memManager->getGlobalVariableInitValue(gv);
 }
+
+void Bv2OpSemContext::resetSolver() {
+  m_z3_solver->reset();
+  m_addedToSolver.clear();
+}
+
+void Bv2OpSemContext::addToSolver(const Expr e) {
+  if (m_addedToSolver.find(e) == m_addedToSolver.end()) {
```
agurfinkel, line 2292: use count() instead of find()

lib/seahorn/BvOpSem2Context.hh
```diff
@@ -102,6 +102,7 @@ private:
   std::shared_ptr<ZSolver<EZ3>> m_z3_solver;
 
   bool m_shouldSimplify = false;
+  std::set<Expr> m_addedToSolver;
```
agurfinkel, line 105: probably unordered set is better here

agurfinkel: or use ExprSet if don't want to make decision

lib/seahorn/BvOpSem2.cc
priyasiddharth: It just prints "Antecedent/Consequest is ..unsat/sat" without debugLoc/instruction info.

lib/seahorn/BvOpSem2.cc
priyasiddharth: done

lib/seahorn/BvOpSem2.cc
agurfinkel: ok, thought it might crash. looks good now.

lib/seahorn/BvOpSem2Context.hh
priyasiddharth: good point

lib/seahorn/BvOpSem2.cc
agurfinkel: I think this was deleted by mistake, and the place where the line is moved is dead code.

lib/seahorn/BvOpSem2.cc
agurfinkel: this is dead code. I think it is a typo. See above.

lib/seahorn/BvOpSem2.cc
priyasiddharth: good point. fixed.

lib/seahorn/BvOpSem2.cc
priyasiddharth: yes.. this is a mix of editor clang-format/misbehaving and my eyeballs not catching it.

lib/seahorn/BvOpSem2.cc
priyasiddharth: same as above.

priyasiddharth:
Should be ready for another review. thanks

priyasiddharth:
Ok I moved all references to a `solver` from `visitAssert` to `solveWithConstraints`.  This makes `solveWithConstraints` reusable. It can be called in fresh/incremental modes.

PR #342, opened 2020-09-08
Add __seahorn_get_value_*() implementation for missing ctype i1

danblitzhou:
`uint1_t` is not defined in C, so `get_value_int(1)` will not be properly expanded like with other integer types; need to force define `__seahorn_get_value_i1` so that nondet boolean functions will behave as expected in executable counter examples.


agurfinkel:
@danblitzhou commit message is not passing our commit linter. Use conventional commits (https://www.conventionalcommits.org/en/v1.0.0/). See commit history for examples.

PR #343, opened 2020-09-09
adding logic to detect conditional branches which become unconditional

priyasiddharth:
The reasoning is that an unconditional branch will remove one branch and we want to know when this happens to understands if important `sassert` stmts are removed.


lib/Transforms/Instrumentation/BranchSentinelPass.cc
```diff
@@ -0,0 +1,92 @@
...
+  static char ID;
+  seahorn::SeaBuiltinsInfo *m_SBI;
+
+  AddBranchSentinelPass() : FunctionPass(ID) {}
+  void augmentBranchInstWithSentinel(BranchInst &BI, IRBuilder<> &B);
+  bool runOnFunction(Function &F) override;
+
+  void getAnalysisUsage(AnalysisUsage &AU) const override {
+    AU.addRequired<seahorn::SeaBuiltinsInfoWrapperPass>();
+    // preserve nothing
```
agurfinkel, line 37: should probably preserve everything. Otherwise hard to combine with other passes

agurfinkel:
looks good. will merge after CI finishes

agurfinkel (approved):

PR #345, opened 2020-09-10
fix(sentinel): branch sentinal pass is now its own subcommand

priyasiddharth:
run by sea spf

### DIFF
--- a/py/sea/commands.py
+++ b/py/sea/commands.py
@@ -977,9 +977,6 @@ class AddBranchSentinel(sea.LimitedCmd):
         ap = super (AddBranchSentinel, self).mk_arg_parser (ap)
         ap.add_argument('--log', dest='log', default=None,
                         metavar='STR', help='Log level')
-        add_bool_argument(ap, 'add-branch-sentinel', dest='add_sentinel', 
-                          help='add branch sentinel for every conditional branch')
-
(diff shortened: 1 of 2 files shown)


agurfinkel:
but why?

before it could be enabled by a switch and ran with vacuity. Now, there is no easy way to run it in verify-c-common

Maybe I am missing something

priyasiddharth:
why?
`name_out_file` takes in arguments from the parent cmd which does not know about the args inside child cmd. The proper way to fix it will be to create a hierarchy of parsers but AFAICT it's not easy to refactor it in the current design... i am adding a new command `spf` which will be sentinal pass + fpf.. this will sidestep the current issue.

how easy is to run it?
`./verify --command=spf --vac seahorn/jobs/byte_buf_reset/
`or
`cmake -DSEA_LINK=llvm-link -DSEAHORN_ROOT=~/seahorn/seahorn/build-dbg/run ../ -GNinja -DVERIFY_FLAGS='--command=spf --vac' -DCMAKE_CXX_COMPILER=clang++-10   -DCMAKE_C_COMPILER=clang-10`

agurfinkel:
I see the problem. Agree with the work around. Sad that it has to be that way. Thx!

PR #349, opened 2020-09-19
fix(yices2:eval): remove redundent recursive handling of non-atomics

danblitzhou:
Removed redundent recursive handling of `NEG` and `AND`; `yices_get_value` should be able to handle leaf nodes without extra handling, see: https://github.com/SRI-CSL/yices2/blob/master/src/include/yices.h#L3822

### DIFF
--- a/lib/seahorn/Smt/MarshalYices.cc
+++ b/lib/seahorn/Smt/MarshalYices.cc
@@ -634,10 +634,8 @@ Expr marshal_yices::eval(Expr expr, ExprFactory &efac, ycache_t &cache,
   Expr domain = nullptr;
   Expr range = nullptr;
 
-  if (bind::isBoolConst(expr) || bind::isIntConst(expr) ||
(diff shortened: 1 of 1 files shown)


lib/seahorn/Smt/MarshalYices.cc
```diff
@@ -634,10 +634,8 @@ Expr marshal_yices::eval(Expr expr, ExprFactory &efac, ycache_t &cache,
   Expr domain = nullptr;
   Expr range = nullptr;
 
-  if (bind::isBoolConst(expr) || bind::isIntConst(expr) ||
-      bind::isRealConst(expr) || op::bv::isBvConst(expr)) {
-    // do nothing
-  } else if (bind::isConst<ARRAY_TY>(expr)) {
+  /* special handling for const array*/
```
agurfinkel, line 637: technically, this is just array, not const-array. That is, constant of type array, as opposed to an array that has one constant value for all entries

PR #351, opened 2020-10-22
feat(opsem): tracking raw mem mgr to track use before def

priyasiddharth:
Tracking and incremental solving enabled by  --horn-bv2-tracking-mem (off by default)

Implemented a new 'raw' memory manager which adds a metadata memory alongside raw memory.
This metadata memory is used to tracking memory properties.
Currently the only property being tracked is 'written-to' or not.
This enable use before def warnings.

to follow in subsequent PRs:
1. tests
2. wiring to run it in sea fpf mode

Limitations:
1. Currently raw mem mgr can be used only with ExtraWideMemMgr.

### DIFF
--- a/lib/seahorn/BvOpSem2TrackingRawMemMgr.hh
+++ b/lib/seahorn/BvOpSem2TrackingRawMemMgr.hh
@@ -1,0 +1,443 @@
+#pragma once
+
+#include "BvOpSem2Context.hh"
+#include "BvOpSem2RawMemMgr.hh"
+
+#include "seahorn/Expr/ExprOpStruct.hh"
+#include "llvm/IR/DebugInfo.h"
+#include "llvm/IR/DebugLoc.h"
+#include "llvm/Support/MathExtras.h"
+#include <bits/stdint-uintn.h>
+
+namespace seahorn {
+namespace details {
+
+// This memory manager adds a metadata memory(backed by raw memory) sitting side
+// by side to conventional memory(backed by raw memory). The word size for
+// conventional memory can be greater than metadata memory.
+//
+// Currently this implementation has a metadata memory word size of 1 byte.
+// For every byte written to conventional memory, we set the corresponding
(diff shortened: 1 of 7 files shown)


lib/seahorn/BvOpSem2TrackingRawMemMgr.hh
```diff
@@ -0,0 +1,443 @@
...
+    // make sure we can fit the supplied value in metadata memory slot
+    assert(llvm::Log2_64(val) + 1 <= g_MetadataBitWidth);
+    return m_metadata.MemSet(ptr, m_ctx.alu().si(val, g_MetadataBitWidth), len,
+                             memIn.getMetadata(), align);
+  }
+
+  // This function reports a warning if there is a possibily for the
+  // passed value to not be equal to 1.
+  // This implies we tried to load memory that was never stored into.
+  void CheckDefBeforeUse(Expr val) {
```
agurfinkel, line 209: eventually this should be factored out

lib/seahorn/BvOpSem2TrackingRawMemMgr.hh
```diff
@@ -0,0 +1,443 @@
...
+  }
+
+  MemValTy storePtrToMem(PtrTy val, PtrTy ptr, MemValTy mem,
+                         unsigned int byteSz, uint64_t align) {
+    RawMemValTy rawVal =
+        m_main.storePtrToMem(val, ptr, mem.getRaw(), byteSz, align);
+    // Number of slots to fill with '1' depends on slot size i.e
+    // g_MetadataByteWidth
+    unsigned int len = byteSz / g_MetadataByteWidth;
+    RawMemValTy metadataVal = memsetMetaData(ptr, len, mem, align, 1U);
```
agurfinkel, line 273: eventually we want to set only one byte per object. not clear whether it will clash with other clients

agurfinkel (approved):
LGTM

PR #352, opened 2020-10-26
refactor(opsem): split mem mgr templates into hdr and src

priyasiddharth:


lib/seahorn/BvOpSem2ExtraWideMemMgr.hh
```diff
@@ -3,7 +3,7 @@
 #include "BvOpSem2Context.hh"
 #include "BvOpSem2RawMemMgr.hh"
 #include "BvOpSem2TrackingRawMemMgr.hh"
-#include "BvOpSem2WideMemManagerMixin.hh"
+#include "BvOpSem2WideMemManagerMixin.cc"
```
agurfinkel, line 6: looks like a bug. You should never include a `cc` file in a header file!

lib/seahorn/BvOpSem2WideMemMgr.cc
```diff
@@ -1,7 +1,7 @@
 #include "BvOpSem2Context.hh"
 #include "BvOpSem2RawMemMgr.hh"
 
-#include "BvOpSem2WideMemManagerMixin.hh"
+#include "BvOpSem2WideMemManagerMixin.cc"
```
agurfinkel, line 4: `cc` file inclusion

agurfinkel:
I am guessing this is in reaction to my request to not have so much code in the header files. But this does not follow the methodology I suggested, instead moves code into `cc` files and includes `cc` files, thus making them header files.

The end result is now worse than before -- code is still in header files, yet now the header files are hidden as `cc` files. Maybe we should talk before another big code change.

lib/seahorn/BvOpSem2ExtraWideMemMgr.hh
priyasiddharth: fixed

lib/seahorn/BvOpSem2WideMemMgr.cc
priyasiddharth: fixed

priyasiddharth:
Fixed following https://isocpp.org/wiki/faq/templates#separate-template-class-defn-from-decl

PR #353, opened 2020-11-04
refactor(opsem): detach tracking memory write from tracking raw mem mgr

priyasiddharth:
This is in preparation to provide intrinsics to set metadata memory and get metadata memory

### DIFF
--- a/lib/seahorn/BvOpSem2Context.hh
+++ b/lib/seahorn/BvOpSem2Context.hh
@@ -478,10 +478,10 @@ public:
 };
 
 /// \brief Creates an instance of OpSemAllocator
-std::unique_ptr<OpSemAllocator> mkNormalOpSemAllocator(OpSemMemManager &mem,
-                                                       unsigned maxSymbAllocSz = 4096);
-std::unique_ptr<OpSemAllocator> mkStaticOpSemAllocator(OpSemMemManager &mem,
-                                                       unsigned maxSymbAllocSz = 4096);
+std::unique_ptr<OpSemAllocator>
+mkNormalOpSemAllocator(OpSemMemManager &mem, unsigned maxSymbAllocSz = 4096);
+std::unique_ptr<OpSemAllocator>
+mkStaticOpSemAllocator(OpSemMemManager &mem, unsigned maxSymbAllocSz = 4096);
 
 class OpSemMemManagerBase {
 protected:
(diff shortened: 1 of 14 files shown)


lib/seahorn/BvOpSem2Context.hh
```diff
@@ -733,6 +733,29 @@ public:
   /// \brief return True expr if number of bytes(byteSz) is within allocated
   /// bounds, False expr otherwise.
   virtual Expr isDereferenceable(PtrTy p, Expr byteSz) = 0;
+
+  /// \brief memset metadata memory associated with a Tracking Memory
+  /// manager and return resulting memory. The 'raw' portion of memory is
+  /// untouched.
+  virtual MemValTy memsetMetaData(PtrTy p, unsigned int len, MemValTy memIn,
+                                  uint32_t align, unsigned int val) = 0;
```
agurfinkel, line 741: why is alignment needed for meta-data?

lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
```diff
@@ -19,6 +21,27 @@ ExtraWideMemManager<T>::ExtraWideMemManager(Bv2OpSem &sem, Bv2OpSemContext &ctx,
   static_assert(std::is_base_of<OpSemMemManagerBase, T>::value,
                 "T not derived from OpSemMemManagerBase");
 }
+
+template <class T>
+typename ExtraWideMemManager<T>::RawMemValTy
+ExtraWideMemManager<T>::setModified(ExtraWideMemManager::PtrTy ptr,
+                                    ExtraWideMemManager::MemValTy mem,
+                                    uint64_t align) {
```
agurfinkel, line 29: why is alignment needed?

lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
```diff
@@ -542,6 +573,49 @@ template <class T>
...
   return m_ctx.alu().si(bytes, g_slotBitWidth);
 }
+template <class T>
+typename ExtraWideMemManager<T>::MemValTy
+ExtraWideMemManager<T>::memsetMetaData(ExtraWideMemManager::PtrTy ptr,
+                                       unsigned int len,
+                                       ExtraWideMemManager::MemValTy memIn,
+                                       uint32_t align, unsigned int val) {
+  RawMemValTy rawOut = memIn.getRaw();
+  if (HasTracking<T>::value) {
```
agurfinkel, line 583: this can be moved into method template parameter so that you have
`memsetMetaData<true>()` and `memsetMetaData<false>()`

lib/seahorn/BvOpSem2MemRepr.cc
```diff
@@ -13,6 +13,12 @@ auto as_std_array(const T &t, const Rest &... rest) ->
 namespace seahorn {
 namespace details {
 
+// MemSet operates at word level.
+// _val must fit within a byte
+// _val is converted to a byte.
```
agurfinkel, line 18: I think this comment is more about implementation than use, so it should be inside the method

lib/seahorn/BvOpSem2WideMemManagerMixin.hh
```diff
@@ -169,8 +172,31 @@ public:
...
+                          unsigned int val) override;
+
+  Expr getMetaData(PtrTy ptr, PtrTy memIn, unsigned int byteSz,
+                   uint32_t align) override;
+  unsigned int getMetaDataMemWordSzInBits() override;
 };
 
+// 'HasTracking' is a solution for conditionally compiling in memory tracking
+// code only when needed. It utilizes C++ metaprogramming and LLVM optimization.
+// FIXME: It suffers from the shortcoming that unneeded calls cannot be
```
agurfinkel, line 190: I think if you move the if-then-else statement based on this into template parameter of the function, you will no longer need to define these functions in all memory managers.

agurfinkel (approved):

priyasiddharth:
Thank you. I will work on these comments in a subsequent PR.

PR #354, opened 2020-11-04
Supporting memhavoc in Executable cex 

danblitzhou:
Recall definition of memhavoc: 
`void memhavoc(void *ptr, size_t sz)`
this tells seahorn to treat memory content starting at `ptr` with size in bytes of `sz` as non-deterministic value.

With change in https://github.com/seahorn/sea-dsa/pull/105, we can track shadow mem symbol of `ptr` and thus track SMT-solver value of non-deterministic memory content.

In CexHarness.cc:
- find and store shadow.mem Expr of `ptr` and value of `sz` for each memhavoc call
- `exprToMemSegment(ptr, sz, ctx, datalayout)`: use HexDump to convert `ptr` Expr to key value pairs of <address id, memory content block> where each block length is word size of current data layout; If Key value pairs only contains default value, put default content block in `n` size array where `n=ceil(size / block size)`, otherwise convert key value pairs into content block arrays (XXX: this is not working at the moment)
- store pointers to all generated memory content block array for memhavoc in a global variable, 
- synthesis memhavoc(ptr, sz):
  - get counter, increment counter like other nd functions
  - get pointer `src` to memory content block array from global array at counter index (`__seahorn_get_value_ptr`)
  - call `memcpy(ptr, src, sz)` to fill memory


lib/seahorn/Smt/HexDump.cc
```diff
@@ -261,7 +263,7 @@ class Pairs {
     // value into the set ( with a blank index)
     if (m_set.size() == 0)
       m_set.emplace(mkTerm<std::string>("*", m_defaultValue->efac()),
-                    m_defaultValue);
+                    m_defaultValue, true);
```
danblitzhou, line 266: reverting since `isRepeated()` should only be relevant between "gaps"

lib/seahorn/CexHarness.cc
```diff
@@ -199,6 +303,60 @@ std::unique_ptr<Module> createCexHarness(BmcTraceWrapper &trace,
...
+            if (auto *hPtrAllocInst = dyn_cast<Instruction>(hPtr)) {
+              hPtrE = trace.eval(loc, *hPtrAllocInst, true);
+            } else {
+              LOG("cex", errs() << "unhandled Value of memhavoc ptr: " << *hPtr
+                                << "\n");
+              continue;
+            }
+
+            FuncValueMap[CF].push_back(shadowMemV);
+            HavocPtrs.push_back(std::make_pair(hPtrE->arg(0), sizeE));
```
agurfinkel, line 356: I think this silently assumes that memory is a struct. This is not always the case.
At any rate, comment what `->arg(0)` is for.

agurfinkel: has this been addressed or is it obsolete?

agurfinkel:
please clang-format so that formatting check stops failing

include/seahorn/Bmc.hh
```diff
@@ -14,20 +21,66 @@
 
 namespace seahorn {
 using namespace expr;
+using ZModel_ref = std::shared_ptr<ZModel<EZ3>>;
```
agurfinkel, line 24: I think our naming convention is not to mix camel and snake case. That is, either caps or underscore, but not both.

include/seahorn/Bmc.hh
```diff
@@ -14,20 +21,66 @@
...
 
 namespace bmc_impl {
 /// true if I is a call to a void function
 bool isCallToVoidFn(const llvm::Instruction &I);
 /// computes an implicant of f (interpreted as a conjunction) that
 /// contains the given model
-void get_model_implicant(const ExprVector &f, ZModel<EZ3> &model,
-                         ExprVector &out, ExprMap &active_bool_map);
+template <typename model_ref>
```
agurfinkel, line 31: minimize code that lives in the header file used by clients. 
`Bmc.hh` is included by clients of Bmc engines. Clients should not import implementation. 

Put implementation into another header file that is only included by `Bmc.cc`. Usual name is `BmcImpl.hh` or some similar convention

include/seahorn/Bmc.hh
```diff
@@ -79,7 +132,8 @@ public:
   }
 
   /// Returns the BMC trace (if available)
-  virtual BmcTrace getTrace();
+  template <class Engine, class Model>
+  BmcTrace<Engine, Model> getTrace();
```
agurfinkel, line 136: define local type for `BmcTrace`. For example
```
using BmcTraceTy = BmcTrace<Engine, Model>
```

include/seahorn/Bmc.hh
```diff
@@ -117,10 +171,10 @@ public:
   std::vector<SymStore> &getStates() { return m_states; }
 };
 
+template <class Engine, class Model>
```
agurfinkel, line 174: Move into `BmcImpl.hh` header. Eventually, we will make it a subclass of some general `BmcTrace` class. For now, moving into a header file will suffice.

include/seahorn/Expr/ExprMemMap.hh
```diff
@@ -0,0 +1,174 @@
+/// Pointer Expr -> Symbolic Memory Map
```
agurfinkel, line 1: Is this new code or factored out from `HexDump`?

include/seahorn/MemSimulator.hh
```diff
@@ -31,20 +31,20 @@ class MemSimulator {
   // -- start of internally allocated memory
   unsigned m_intMemStart;
```
agurfinkel, line 32: `MemSimulator` is dead code. If it gets too hard to update it, we can remove it.

lib/seahorn/Bmc.cc
```diff
@@ -130,205 +129,11 @@ raw_ostream &BmcEngine::toSmtLib(raw_ostream &out) {
   return m_smt_solver.toSmtLib(out);
 }
 
-BmcTrace BmcEngine::getTrace() {
+template <> BmcTrace<BmcEngine, ZModel_ref> BmcEngine::getTrace() {
```
agurfinkel, line 132: this template specialization is unexpected here. Probably issue somewhere else in the design. 
I expect not to have `template<>` in front of this method.

lib/seahorn/Bmc.cc
```diff
@@ -130,205 +129,11 @@ raw_ostream &BmcEngine::toSmtLib(raw_ostream &out) {
...
-        out << " [" << (*dloc).getFilename() << ":" << dloc.getLine() << "]";
-      }
-      out << "\n";
-      out.resetColor();
-    }
-  }
-  out << "End trace\n";
-  return out;
+  return BmcTrace<BmcEngine, ZModel_ref>(
+      *this, std::make_shared<ZModel<EZ3>>(model));
```
agurfinkel, line 136: Worried about this line. What does it actually do? 
Create a model on the heap using a copy constructor and and making it shared?

This is quite different from the previous design that does not allocate on the heap. Why the change?

lib/seahorn/Bmc.cc
```diff
@@ -439,6 +198,35 @@ void unsat_core(ZSolver<EZ3> &solver, const ExprVector &f, bool simplify,
     out.push_back(bind::fname(bind::fname(c))->arg(0));
 }
 
+template <>
```
agurfinkel, line 201: why `template<>`?

lib/seahorn/HornCex.cc
```diff
@@ -95,7 +95,8 @@ using namespace llvm;
 namespace seahorn {
 
 template <typename O> class SvCompCex;
-static void dumpSvCompCex(BmcTrace &trace, std::string CexFile);
+static void dumpSvCompCex(BmcTrace<BmcEngine, ZModel_ref> &trace,
```
agurfinkel, line 98: Add a type for `BmcTrace<BmcEngine, ZModel_ref>` and use throughout. For example, call it `ZBmcTrace`

lib/seahorn/HornCex.cc
```diff
@@ -302,7 +303,7 @@ bool HornCex::runOnFunction(Module &M, Function &F) {
   LOG("cex", errs() << "Validated CEX by BMC engine.\n";);
 
   // get bmc trace
-  BmcTrace trace(bmc.getTrace());
+  BmcTrace<BmcEngine, ZModel_ref> trace(bmc.getTrace<BmcEngine, ZModel_ref>());
```
agurfinkel, line 306: This looks wrong. `getTrace` should not need to have template parameters specified. It gets them from its class `bmc`

lib/seahorn/HornCex.cc
```diff
@@ -483,7 +484,8 @@ static void debugLocToSvComp(const Instruction &inst, SvCompCex<O> &svcomp) {
   svcomp.edge(file, (int)dloc.getLine(), "");
 }
 
-static void dumpSvCompCex(BmcTrace &trace, std::string CexFile) {
+static void dumpSvCompCex(BmcTrace<BmcEngine, ZModel_ref> &trace,
```
agurfinkel, line 487: type for `BmcTrace`. see comment above.

lib/seahorn/SolverBmc.cc
```diff
@@ -120,274 +118,11 @@ raw_ostream &SolverBmcEngine::toSmtLib(raw_ostream &out) {
   return out;
 }
 
-SolverBmcTrace SolverBmcEngine::getTrace() {
+template <>
```
agurfinkel, line 121: should not need `template<>`

agurfinkel (changes requested):
Looking good. A few confusions on my part around templates.

let's fix it and merge as soon as possible

include/seahorn/Expr/ExprMemMap.hh
danblitzhou: factored out from `HexDump`; cut short much of the layers of old design;
now its just:
`ExprMemMap` has `MemCellSet` and `MemExprVisitor`;
`MemCellSet` contains a set of `MemCell`s and functions for access

include/seahorn/MemSimulator.hh
danblitzhou: `HornCex.cc` is the only place that still uses it, what is our plan for that file?

lib/seahorn/Bmc.cc
danblitzhou: I changed `m_model` from a reference to a shared pointer to match `Solver::model_ref`, which is a shared pointer of model implementation.

lib/seahorn/Bmc.cc
danblitzhou: this is a specialization of template `Out` using llvm::raw_ostream

danblitzhou:
Regarding `CexHarness`, right now I just copied some functions into a `impl` header file to make it work with template `BmcTrace` so that `horn-bmc-solver=smt-y2` produces cex harness. 
I have a more comprehensive refactor of cex harness generating system almost ready on top of changes in this PR, maybe it's better to put those changes in a separate PR.

agurfinkel:
CI in azure times out after 1.5 hrs. Do the tests pass for you locally, or is that a genuine problem?

include/seahorn/MemSimulator.hh
agurfinkel: We can refactor anything we want. There are no other dependencies other than our code.

lib/seahorn/Bmc.cc
```diff
@@ -130,205 +129,10 @@ raw_ostream &BmcEngine::toSmtLib(raw_ostream &out) {
...
-          out.changeColor(raw_ostream::CYAN);
-        out << " [" << (*dloc).getFilename() << ":" << dloc.getLine() << "]";
-      }
-      out << "\n";
-      out.resetColor();
-    }
-  }
-  out << "End trace\n";
-  return out;
+  return ZBmcTraceTy(*this, std::make_shared<ZModel<EZ3>>(model));
```
agurfinkel, line 135: This is potentially problematic. Double check ownership.

`model` from the `m_smt_solver` is already a smart pointer, wrapping it in another smart pointer is at the very least annoying. Plus the allocation is shifted to the heap...

lib/seahorn/Bmc.cc
agurfinkel: add a comment that this is a specialization, just like the comment above

lib/seahorn/BmcPass.cc
```diff
@@ -348,24 +348,24 @@ class BmcPass : public llvm::ModulePass {
...
-          errs() << "Analyzed Function:\n" << F << "\n";
-          errs() << "Trace \n";
-          BmcTrace trace(bmc.getTrace());
-          trace.print(errs());
-        });
-
     if (res) {
+      ZBmcTraceTy trace(bmc.getTrace());
+      LOG(
+          "cex", if (res) {
```
agurfinkel, line 354: formatting. I usually do 
```
LOG("cex", { if... } );
```
Then clang-format works properly.

lib/seahorn/BmcPass.cc
```diff
@@ -348,24 +348,24 @@ class BmcPass : public llvm::ModulePass {
...
       if (CexFileRef != "") {
         if (CexFileRef.endswith(".ll") || CexFileRef.endswith(".bc")) {
           auto &tli = getAnalysis<TargetLibraryInfoWrapperPass>();
           auto const &dl = F.getParent()->getDataLayout();
-          BmcTrace trace(bmc.getTrace());
-          BmcTraceWrapper trace_wrapper(trace);
-          dumpLLVMCex(trace_wrapper, CexFileRef, dl, tli.getTLI(F),
-                      F.getContext());
+          BmcTraceWrapper<ZBmcTraceTy> trace_wrapper(trace);
+          dumpLLVMCex<ZBmcTraceTy>(trace_wrapper, CexFileRef, dl, tli.getTLI(F),
```
agurfinkel, line 367: `ZBmcTraceTy` should not be needed. If it is needed, then something is wrong with how it is declared.

lib/seahorn/BmcPass.cc
```diff
@@ -397,27 +397,40 @@ class BmcPass : public llvm::ModulePass {
...
-    LOG(
-        "cex", if (res == solver::SolverResult::SAT) {
-          errs() << "Analyzed Function:\n" << F << "\n";
-          errs() << "Trace \n";
-          SolverBmcTrace trace(bmc.getTrace());
-          trace.print(errs());
-        });
-    // todo: support --cex mode (generate harness)
+    if (res == SolverResult::SAT) {
+      SolverBmcTraceTy trace(bmc.getTrace());
```
agurfinkel, line 414: this whole `if` should be inside the `LOG` macro. If `LOG` is disabled, none of this code should be executed.

lib/seahorn/BmcPass.cc
```diff
@@ -397,27 +397,40 @@ class BmcPass : public llvm::ModulePass {
...
+        errs() << "Trace \n";
+        trace.print(errs());
+      });
+      StringRef CexFileRef(HornCexFile);
+      if (CexFileRef != "") {
+        if (CexFileRef.endswith(".ll") || CexFileRef.endswith(".bc")) {
+          auto &tli = getAnalysis<TargetLibraryInfoWrapperPass>();
+          auto const &dl = F.getParent()->getDataLayout();
+          BmcTraceWrapper<SolverBmcTraceTy> trace_wrapper(trace);
+          dumpLLVMCex<SolverBmcTraceTy>(trace_wrapper, CexFileRef, dl,
```
agurfinkel, line 426: `SolverBmcTraceTy` is not needed as specialization for `dumpLLVMVCex`

lib/seahorn/HornCex.cc
```diff
@@ -327,11 +327,12 @@ bool HornCex::runOnFunction(Module &M, Function &F) {
...
   if (HornCexFileRef.endswith(".ll") || HornCexFileRef.endswith(".bc")) {
     const DataLayout &dl = M.getDataLayout();
-    std::unique_ptr<BmcTraceWrapper> traceW(new BmcTraceWrapper(trace));
+    std::unique_ptr<BmcTraceWrapper<ZBmcTraceTy>> traceW(
+        new BmcTraceWrapper<ZBmcTraceTy>(trace));
     if (memSim) {
       traceW.reset(new BmcTraceMemSim(*memSim));
     }
-    dumpLLVMCex(*traceW, HornCexFileRef, dl, tli, F.getContext());
+    dumpLLVMCex<ZBmcTraceTy>(*traceW, HornCexFileRef, dl, tli, F.getContext());
```
agurfinkel, line 335: no need for `ZBmcTraceTy`

agurfinkel (changes requested):

lib/seahorn/Bmc.cc
danblitzhou: `m_smt_solver.getModel()` returns a `ZModel<EZ3>` object/reference, which is why I wanted to wrap it around `ZModelRef` in the first place; perhaps I should initialize `ZBmcTraceTy::m_model` with the pointer of `m_smt_solver.getModel()` directly?

agurfinkel: ZModel is a smart pointer. You can copy it by value.

agurfinkel (approved):

PR #355, opened 2020-11-09
introduce is_modified intrinsic

priyasiddharth:
Some left over items will be added subsequently once core functionality is out since i don't want to block on refactoring.


agurfinkel (approved):
looks good. no comments.

PR #357, opened 2020-11-11
refactor(opsem): use compile time compute to reduce duplicate code

priyasiddharth:
This uses 'tags' to enable custom functionality to be added to only a subset of memory managers without adding default(NOP) impl to the rest of memory managers.

It has been implemented using boost::hana which I chose in favour of std::enable_if since it is easier to construct a solution using hana

Pros of boost::hana vs raw SFINAE (C++ 14)
1. Not easy to check if component class of mixin has a feature in raw SFINAE. This is an advertised idiom in Hana.
2. Easier to construct - Raw SFINAE does not support partial specialization of class methods. Thus we cannot use a class template param in specializing a method. This leads to more verbosity at call site.
3. while the sprinkling of eval_if + lambdas takes getting used to - imo it is more readable/understandable than the equivalent function specialization which will require two method declarations + definitions for true/false for each method we want to specialized.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -247,8 +247,9 @@ if(NOT LLVM_FOUND)
   return()
 endif()
 
-set (Boost_USE_STATIC_LIBS ${SEAHORN_STATIC_EXE})
-find_package (Boost 1.65 REQUIRED)
+# boost version >= 1.71 because of https://github.com/boostorg/hana/issues/446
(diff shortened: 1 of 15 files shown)


priyasiddharth:
i'm looking into why the build is failing...

priyasiddharth:
@agurfinkel Please take a look. Thanks.

PR #360, opened 2020-11-17
feat(opsem): add semantics for reset_modified

priyasiddharth:


lib/seahorn/BvOpSem2.cc
```diff
@@ -643,6 +643,7 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemVisitorBase {
     if (!m_ctx.getMemReadRegister()) {
       LOG("opsem", ERR << "No read register found - check if corresponding"
                           "shadow instruction is present.");
+      m_ctx.setMemReadRegister(Expr());
```
agurfinkel, line 646: seems dangerous. This is unexpected. The program should crash or exit at this point.

lib/seahorn/BvOpSem2.cc
```diff
@@ -653,6 +654,25 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemVisitorBase {
     m_ctx.setMemReadRegister(Expr());
   }
 
+  void visitResetModified(CallSite CS) {
+    if (!m_ctx.getMemReadRegister() || !m_ctx.getMemWriteRegister()) {
+      LOG("opsem",
+          ERR << "No read/write register found - check if corresponding"
+                 "shadow instruction is present.");
+      m_ctx.setMemReadRegister(Expr());
```
agurfinkel, line 662: same comment as before. If this is needed, it is masking a deeper problem.

agurfinkel: or maybe this is just resetting the variables to null. In this case, add a comment. I though they are set to nondet.

lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
```diff
@@ -583,7 +587,7 @@ ExtraWideMemManager<T>::memsetMetaData(ExtraWideMemManager::PtrTy ptr,
       MemoryFeatures::has_tracking(hana::type<T>{}),
       [&](auto _) {
         return _(m_main).memsetMetaData(ptr.getBase(), len, memIn.getRaw(),
-                                        val);
+                                        val = val);
```
agurfinkel, line 590: this assignment `val = val` looks off

lib/seahorn/BvOpSem2WideMemManagerMixin.cc
```diff
@@ -35,8 +35,9 @@ OpSemWideMemManagerMixin<BaseT>::memsetMetaData(PtrTy p, unsigned int len,
   return hana::eval_if(
       MemoryFeatures::has_tracking(hana::type<BaseT>{}),
       [&](auto _) {
-        auto r = _(base()).memsetMetaData(BasePtrTy(std::move(p)), len,
-                                          BaseMemValTy(std::move(memIn)), 0);
+        auto r =
+            _(base()).memsetMetaData(BasePtrTy(std::move(p)), len,
+                                     BaseMemValTy(std::move(memIn)), val = val);
```
agurfinkel, line 40: `val = val` assignment again

lib/seahorn/BvOpSem2WideMemManagerMixin.cc
```diff
@@ -49,8 +50,9 @@ OpSemWideMemManagerMixin<BaseT>::memsetMetaData(PtrTy p, Expr len,
...
```
agurfinkel, line 55: `val = val`

agurfinkel (approved):

lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
priyasiddharth: using named parameters just to make i know what i am passing since `len` and `val` have same datatype. Earlier due to automated refactoring, this was set to 0 and I had to visually squint to see that val was turned to 0

lib/seahorn/BvOpSem2WideMemManagerMixin.cc
priyasiddharth: same comment as above.

lib/seahorn/BvOpSem2WideMemManagerMixin.cc
priyasiddharth: same as above

lib/seahorn/BvOpSem2.cc
priyasiddharth: I am copying the pattern here: https://sourcegraph.com/github.com/seahorn/seahorn@dev10/-/blob/lib/seahorn/BvOpSem2.cc#L1793

Does the original pattern also have the same problem?

lib/seahorn/BvOpSem2.cc
priyasiddharth: See comment above.

lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
agurfinkel: I did not know that there are named parameters in C++. A comment would be useful. 

In other contexts, `v = v` evaluates to `v`, so code like `foo(v = v)` is legal in C, but it means `v = v; foo(v);`

priyasiddharth:
Will submit this and fix comments in next commit.

PR #363, opened 2020-11-24
feat(build): add CI using github actions

priyasiddharth:


priyasiddharth:
Currently the CI is failing because `seahorn` repo does not define the following secrets

```
DOCKER_HUB_USERNAME
DOCKER_HUB_ACCESS_TOKEN 
```

Current run can be seen at: https://github.com/priyasiddharth/seahorn/runs/1449131154?check_suite_focus=true

A previous successful run can be seen at: https://github.com/priyasiddharth/seahorn/actions/runs/381504324

agurfinkel:
This is fantastic. I've added secrets, so they should be accessible.

Upon successful test, the action should push the result to docker-hub. This is currently missing.
Unlike our current travis, we should push `seahorn-builder` into `seahorn/seahorn-builder-llvm10:nightly` independently of tests succeeding and failing. And then push `seahorn` into `seahorn/seahorn-llvm10:nightly` if the tests pass.

Once this is up and running, we'll need a second job for coverage (maybe not as frequent). I've added a secret with access token for codecov.io already.

Then we'll have to experiment how to do builds for PRs.

PR #364, opened 2020-11-26
feat(ci): add ci for dev10

priyasiddharth:
scheduled actions must exist in default branch to run. Checking out dev10 only to limit this action to that branch. See https://stackoverflow.com/questions/58798886/github-actions-schedule-operation-on-branch

### DIFF
--- a/.github/workflows/test-seahorn-docker.yml
+++ b/.github/workflows/test-seahorn-docker.yml
@@ -1,0 +1,55 @@
+# workflow to build and test seahorn
+
+name: CI
+
+# Controls when the action will run. 
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
+  push:
+    branches: dev10
+  pull_request:
+    branches: dev10
+  schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-seahorn-docker.yml
```diff
@@ -0,0 +1,55 @@
...
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - name: Check Out Repo 
+        uses: actions/checkout@v2
+        with:
+          ref: dev10  # only checkout dev10
```
agurfinkel, line 28: will this work on PRs? Maybe we should have separate actions. One for PR, one cron. The cron one can use a fixed branch.

agurfinkel: oh, this only lives in master

PR #366, opened 2020-12-03
fix(opsem): convert signed and unsigned integers to symbolic values

agurfinkel:
This addresses a problem that some address constants are interpreted as negative
numbers.

### DIFF
--- a/lib/seahorn/BvOpSem2.cc
+++ b/lib/seahorn/BvOpSem2.cc
@@ -1240,7 +1240,11 @@ public:
       Expr isNoOverflow =
           m_ctx.alu().IsUaddNoOverflow(op0, op1, ty->getScalarSizeInBits());
       assert(addRes && isNoOverflow);
-      Expr maxVal = m_ctx.alu().si(~0UL, ty->getScalarSizeInBits());
+      mpz_class maxValZ;
+      for (unsigned i = 0; i < ty->getScalarSizeInBits(); ++i) {
+        maxValZ.setbit(i);
+      }
+      Expr maxVal = m_ctx.alu().num(maxValZ, ty->getScalarSizeInBits());
       Expr res = boolop::lite(isNoOverflow, addRes, maxVal);
       setValue(I, res);
     } break;
@@ -1361,7 +1365,7 @@ public:
                                           unsigned opResultBitWidth,
                                           unsigned carryBitPadwidth) {
     Expr carry = m_ctx.alu().Concat(
-        {m_ctx.alu().si(0U, carryBitPadwidth), carryBitPadwidth},
+        {m_ctx.alu().ui(0U, carryBitPadwidth), carryBitPadwidth},
         {carryBit, 1});
     return m_ctx.alu().Concat({carry, carryBitPadwidth},
                               {opResult, opResultBitWidth});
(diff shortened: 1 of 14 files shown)


lib/seahorn/BvOpSem2.cc
```diff
@@ -1240,7 +1240,11 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemVisitorBase {
       Expr isNoOverflow =
           m_ctx.alu().IsUaddNoOverflow(op0, op1, ty->getScalarSizeInBits());
       assert(addRes && isNoOverflow);
-      Expr maxVal = m_ctx.alu().si(~0UL, ty->getScalarSizeInBits());
+      mpz_class maxValZ;
+      for (unsigned i = 0; i < ty->getScalarSizeInBits(); ++i) {
+        maxValZ.setbit(i);
+      }
```
priyasiddharth, line 1246: why is this better than `maxValZ = 2^(ty->getScalarSizeInBits()) - 1`?

lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
```diff
@@ -407,12 +407,11 @@ ExtraWideMemManager<T>::ptrAdd(ExtraWideMemManager::PtrTy base,
   // do concrete computation if possible
   if (m_ctx.alu().isNum(base.getOffset())) {
     signed conc_offset = m_ctx.alu().toNum(base.getSize()).get_si() + _offset;
```
priyasiddharth, line 409: unrelated to this change:  `s/base.getSize()/base.getOffset()`

lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
```diff
@@ -407,12 +407,11 @@ ExtraWideMemManager<T>::ptrAdd(ExtraWideMemManager::PtrTy base,
...
-    new_offset = m_ctx.alu().si(conc_offset, ptrSizeInBits());
+    new_offset = m_ctx.alu().ui(conc_offset, ptrSizeInBits());
```
priyasiddharth, line 410: why is this `ui` and not `si`, since we allow negative offset in symbolic computation in line 412

priyasiddharth (changes requested):

lib/seahorn/BvOpSem2.cc
agurfinkel: The question is what space to do the computation at. This way, the computation is in infinite precision space, so it is guaranteed to be precise. If we do it in machine arithmetic, it might underflow depending on the architecture (i.e., size of `unsigned`) and requested number of bits. This was the problem with the previous solution that caused some test to fail.

lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
agurfinkel: goot catch. will be fixed in next revision.

lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
agurfinkel: fixed

priyasiddharth:
LGTM.

priyasiddharth (approved):

PR #367, opened 2020-12-04
refactor(opsem): move to a single mixin for mem managers

priyasiddharth:
All conditional logic is now in MemoryManagerMixin.


lib/seahorn/BvOpSem2MemManagerMixin.hh
```diff
@@ -306,25 +309,103 @@ public:
...
   Expr getFatData(PtrTy p, unsigned SlotIdx) override {
-    auto res = base().getFatData(BasePtrTy(std::move(p)), SlotIdx);
-    return res;
+    return hana::eval_if(
+        MemoryFeatures::has_fatmem(hana::type<BaseT>{}),
+        [&](auto _) {
+          auto res = _(base()).getFatData(BasePtrTy(std::move(p)), SlotIdx);
+          return res;
+        },
+        [&] {
```
agurfinkel, line 318: no need for brackets?

agurfinkel (approved):
LGTM

lib/seahorn/BvOpSem2MemManagerMixin.hh
priyasiddharth: nothing being passed in since we just print an error message.

PR #368, opened 2020-12-05
refactor(cex): new system for executable cex

danblitzhou:
- created new class `CexExeGenerator` to replace `CexHarness`; the new system modularized the process of storing evaluated expression and building LLVM IR for different type of functions.
- created APIs `bool isPtrTy(Expr e)` and `Expr ptrToAddr(Expr p)` for each BvOpsem2 memory managers; `isPtrTy` takes an expression and checks whether it is the correct format under current memory manager; `ptrToAddr` takes an expression and checks whether `isPtrTy` is true; if true, extracts and returns addressable part from it, otherwise returns empty expression.
- Added `Expr getPtrAddressable(Expr e)` function to `BmcEngine` and `SolverBmcEngine`; `getPtrAddressable` checks whether current opsem context is BvOpsem 2, if so, use the context's memory manager to convert encoded memory expression to addressable part only using `ptrToAddr`;  not sure if other operational semantics have special encoding as well, currently implementing special handling for BvOpsem 2 in `getPtrAddressable`.

### DIFF
--- a/include/seahorn/Bmc.hh
+++ b/include/seahorn/Bmc.hh
@@ -103,6 +103,14 @@ public:
     }
     return reg;
   }
+
+  // given an Expr encoding of pointer, return only addressable part
+  Expr getPtrAddressable(Expr e) {
+    if (!m_semCtx)
+      return Expr();
+    return m_semCtx->ptrToAddr(e);
+  }
+
   /// Dump unsat core
   /// Exposes internal details. Intendent to be used for debugging only
   virtual void unsatCore(ExprVector &out);
--- a/include/seahorn/CexExeGenerator.hh
+++ b/include/seahorn/CexExeGenerator.hh
@@ -1,0 +1,104 @@
+#pragma once
+
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Bitcode/BitcodeWriter.h"
+#include "llvm/IR/DataLayout.h"
+#include "llvm/IR/IRBuilder.h"
+#include "llvm/IR/LLVMContext.h"
+#include "llvm/IR/ValueMap.h"
+#include "llvm/IR/Verifier.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/ToolOutputFile.h"
+
+#include "boost/algorithm/string/replace.hpp"
+
+#include "seahorn/Bmc.hh"
+#include "seahorn/Expr/ExprLlvm.hh"
+#include "seahorn/Expr/ExprMemMap.hh"
+#include "seahorn/Expr/ExprOpBinder.hh"
+#include "seahorn/MemSimulator.hh"
+#include "seahorn/SolverBmc.hh"
+#include "seahorn/Support/SeaDebug.h"
+#include "seahorn/Transforms/Instrumentation/ShadowMemDsa.hh"
+#include <memory>
+
+namespace llvm {
+class TargetLibraryInfo;
+class DataLayout;
+class LLVMContext;
+class Module;
+} // namespace llvm
+
+/*
+    Converts bmc trace into a linkable LLVM IR
+    with all non-det functions implemented
+ */
+
+namespace seahorn {
+namespace cexGen {
+
+using MemMap = expr::exprMemMap::ExprMemMap;
+
+// helper functions
+namespace utils {
+
+/// \brief extract content from memory expression \p e using ExprMemMap;
+/// if sucessful, store default to \p e defaultValue, special id/value pairs
+/// to \p out
+template <typename kv>
+bool extractArrayContents(Expr e, kv &out, Expr &defaultValue);
+
+} // namespace utils
+
+template <class Trace> class CexExeGenerator {
+  Trace &m_trace;
+  const DataLayout &m_dl;
+  const TargetLibraryInfo &m_tli;
+  LLVMContext &m_context;
+  std::unique_ptr<Module> m_harness;
+
+  // map function calls to return value(s)
+  ValueMap<const Function *, ExprVector> m_func_val_map;
+  // list of <start addr, size> of havoc pointers
+  std::vector<std::pair<Expr, Expr>> m_memhavoc_args;
+
+  /// \brief fills m_memhavoc_args and m_func_val_map
+  void storeMemHavoc(unsigned loc, const Function *func, ImmutableCallSite cs,
+                     ImmutableCallSite prevCS);
+
+  /// \brief fills func val map, dsa data and memhavoc ptr info
+  void storeDataFromTrace();
+
+  /// \brief: Given llvm::Function \p func and a list of return values (Expr)
+  /// \p values in order of execution, convert \p values to a constant array,
+  /// implement \p func to return items in array in order per invocation
+  void buildNonDetFunction(const Function *func, ExprVector &values);
+
+  void buildMemhavoc(const Function *func, ExprVector &values);
+
+  /* build nd functions with primitive type return values, memhavoc and
+     shadow.mem.init */
+  void buildCexModule();
+
+  /// \brief converts Expr \p e into LLVM Constant with LLVM Type \p ty
+  Constant *exprToConstant(Type *ty, Expr e);
+
+  /// \brief extract default value and special id-val pairs from Expr encoding
+  /// of a shadow mem segment \p segment using ExprMemMap; returns a
+  /// ConstantArray of length \p size divided by content width, filled with
+  /// default and special values
+  Constant *exprToMemSegment(Expr segment, Expr startAddr, Expr size);
+
+public:
+  CexExeGenerator(Trace &trace, const DataLayout &dl,
+                  const TargetLibraryInfo &tli, LLVMContext &context)
+      : m_trace(trace), m_dl(dl), m_tli(tli), m_context(context) {
+    storeDataFromTrace();
+    buildCexModule();
+  }
+
+  void saveCexModuleToFile(llvm::StringRef CexFile);
+};
+
+} // namespace cexGen
+} // namespace seahorn
--- a/include/seahorn/SolverBmc.hh
+++ b/include/seahorn/SolverBmc.hh
@@ -87,6 +87,13 @@ public:
     return reg;
   }
 
+  // given an Expr encoding of pointer, return only addressable part
+  Expr getPtrAddressable(Expr e) {
+    if (!m_semCtx)
+      return Expr();
+    return m_semCtx->ptrToAddr(e);
+  }
+
   /// output current path condition in SMT-LIB2 format
   virtual raw_ostream &toSmtLib(raw_ostream &out);
 
--- a/lib/seahorn/BmcPass.cc
+++ b/lib/seahorn/BmcPass.cc
@@ -20,6 +20,7 @@
 #include "seahorn/BvOpSem.hh"
 #include "seahorn/BvOpSem2.hh"
 #include "seahorn/CallUtils.hh"
+#include "seahorn/CexExeGenerator.hh"
 #include "seahorn/CexHarness.hh"
 #include "seahorn/DfCoiAnalysis.hh"
 #include "seahorn/PathBmc.hh"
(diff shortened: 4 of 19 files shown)


lib/seahorn/CexExeGenerator.cc
```diff
@@ -0,0 +1,568 @@
...
+      auto *segmentConst = exprToLlvm(segmentElmTy, segmentE, ctx, dl);
+      LLVMValueSegment[index] = segmentConst;
+    } else
+      continue;
+  }
+  return ConstantArray::get(segmentAT, LLVMValueSegment);
+}
+
+bool extractArrayContents(Expr e, DsaKeyVal &out, Expr &defaultValue) {
+  const MemMap a_map(e);
```
danblitzhou, line 124: note that this is rewritten to leverage the more robust `ExprMemMap`; reading older version in `CexHarness` I notices the process is nearly identical to `ExprMemMap`'s array visitor

include/seahorn/Bmc.hh
```diff
@@ -103,6 +105,17 @@ public:
     }
     return reg;
   }
+
+  // given an Expr encoding of pointer, return only addressable part
+  Expr getPtrAddressable(Expr e) {
+    if (!m_semCtx)
+      return Expr();
+    if (auto *bv2SemCtx = dynamic_cast<Bv2OpSemContext *>(m_semCtx.get())) {
```
agurfinkel, line 113: no dynamic casts in seahorn 

just add the desired method to the base class

include/seahorn/SolverBmc.hh
```diff
@@ -87,6 +87,16 @@ public:
     return reg;
   }
 
+  // given an Expr encoding of pointer, return only addressable part
+  Expr getPtrAddressable(Expr e) {
+    if (!m_semCtx)
+      return Expr();
+    if (auto *bv2SemCtx = dynamic_cast<Bv2OpSemContext *>(m_semCtx.get())) {
```
agurfinkel, line 94: no dynamic casts

lib/seahorn/BmcPass.cc
```diff
@@ -360,9 +361,9 @@ class BmcPass : public llvm::ModulePass {
           if (CexFileRef.endswith(".ll") || CexFileRef.endswith(".bc")) {
             auto &tli = getAnalysis<TargetLibraryInfoWrapperPass>();
             auto const &dl = F.getParent()->getDataLayout();
-            BmcTraceWrapper<ZBmcTraceTy> trace_wrapper(trace);
-            dumpLLVMCex(trace_wrapper, CexFileRef, dl, tli.getTLI(F),
-                        F.getContext());
+            cexGen::CexExeGenerator<ZBmcTraceTy> cex(trace, dl, tli.getTLI(F),
+                                                     F.getContext());
+            cex.dumpCexModuleToFile(CexFileRef);
```
agurfinkel, line 366: save might be better than dump

lib/seahorn/BvOpSem2Allocators.cc
```diff
@@ -1,5 +1,5 @@
 #include "BvOpSem2Allocators.hh"
-#include "BvOpSem2Context.hh"
+#include "seahorn/BvOpSem2Context.hh"
```
agurfinkel, line 2: the inclusion used to be relative because we do not expose these classes to clients.

lib/seahorn/BvOpSem2Allocators.hh
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
-#include "BvOpSem2Context.hh"
 #include "BvOpSem2RawMemMgr.hh"
+#include "seahorn/BvOpSem2Context.hh"
```
agurfinkel, line 4: see comments above

lib/seahorn/BvOpSem2Alu.cc
```diff
@@ -1,4 +1,4 @@
-#include "BvOpSem2Context.hh"
+#include "seahorn/BvOpSem2Context.hh"
```
agurfinkel, line 1: see comment above

lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
```diff
@@ -97,6 +97,11 @@ ExtraWideMemManager<T>::getAddressable(ExtraWideMemManager::PtrTy p) const {
   }
   return m_ctx.alu().doAdd(p.getBase(), p.getOffset(), ptrSizeInBits());
 }
+
+template <class T> bool ExtraWideMemManager<T>::isPtrTy(Expr e) const {
+  return strct::isStructVal(e) && e->arity() == g_num_slots;
```
agurfinkel, line 102: rename into isPtrTyVal

because the check is on the value, not on the type

agurfinkel (changes requested):

include/seahorn/CexExeGenerator.hh
```diff
@@ -0,0 +1,116 @@
...
+/*
+    Converts bmc trace into a linkable LLVM IR
+    with all non-det functions implemented
+ */
+
+namespace seahorn {
+namespace cexGen {
+
+using MemMap = expr::exprMemMap::ExprMemMap;
+using DsaKeyVal = std::map<Expr, Expr>;
```
agurfinkel, line 41: consider using unordered map or DenseMap from llvm

include/seahorn/CexExeGenerator.hh
```diff
@@ -0,0 +1,116 @@
...
+
+// helper functions
+namespace utils {
+
+/// \brief converts Expr \p e into LLVM Constant
+Constant *exprToLlvm(Type *ty, Expr e, LLVMContext &ctx, const DataLayout &dl);
```
agurfinkel, line 47: not the most intuitive name. 

why not put it in a class and cache results

include/seahorn/CexExeGenerator.hh
```diff
@@ -0,0 +1,116 @@
...
+
+/// \brief extract default value and special id-val pairs from Expr encoding of
+/// a shadow mem segment \p segment using ExprMemMap; returns a ConstantArray
+/// of length \p size divided by content width, filled with default and special
+/// values
+Constant *exprToMemSegment(Expr segment, Expr startAddr, Expr size,
```
agurfinkel, line 53: again, move into some class to simplify the args at each call. same class as prev method

include/seahorn/CexExeGenerator.hh
```diff
@@ -0,0 +1,116 @@
...
+                           llvm::LLVMContext &ctx, const llvm::DataLayout &dl);
+
+/// \brief extract content from array expression \p e using ExprMemMap;
+/// if sucessful, store default to \p e defaultValue, special id/value pairs
+/// to \p out
+bool extractArrayContents(Expr e, DsaKeyVal &out, Expr &defaultValue);
```
agurfinkel, line 59: why Dsa in the name

include/seahorn/CexExeGenerator.hh
```diff
@@ -0,0 +1,116 @@
...
+  Trace &m_trace;
+  const DataLayout &m_dl;
+  const TargetLibraryInfo &m_tli;
+  LLVMContext &m_context;
+  std::unique_ptr<Module> m_harness;
+
+  // map function calls to return value(s)
+  ValueMap<const Function *, ExprVector> m_func_val_map;
+  // map a dsa node to start and end addresses
+  std::map<unsigned, std::pair<Expr, Expr>> m_dsa_alloc_map;
```
agurfinkel, line 73: can't see why this would be useful

include/seahorn/CexExeGenerator.hh
```diff
@@ -0,0 +1,116 @@
...
+  // map a dsa node to its contents (as a pair of a default value
+  // and a map from index to value)
+  std::map<unsigned, std::pair<Expr, DsaKeyVal>> m_dsa_content_map;
```
agurfinkel, line 76: starting to get very confused...

include/seahorn/CexExeGenerator.hh
```diff
@@ -0,0 +1,116 @@
...
+  // list of <start addr, size> of havoc pointers
+  std::vector<std::pair<Expr, Expr>> m_havoc_ptrs;
```
agurfinkel, line 78: what are havoc pointers?

lib/seahorn/CexExeGenerator.cc
```diff
@@ -0,0 +1,568 @@
+#include "seahorn/CexExeGenerator.hh"
+
+namespace seahorn {
+namespace cexGen {
+namespace utils {
+Constant *exprToLlvm(Type *ty, Expr e, LLVMContext &ctx, const DataLayout &dl) {
+  if (isOpX<TRUE>(e)) {
```
agurfinkel, line 7: this seems like old not updated code  with scary sounding comments

lib/seahorn/CexExeGenerator.cc
```diff
@@ -0,0 +1,568 @@
...
+      // like it is.
+      return Constant::getIntegerValue(
+          ty, toAPInt(dl.getTypeStoreSizeInBits(ty), mpz));
+    }
+    llvm_unreachable("Unhandled type");
+  } else {
+    // HACK: in structures, first argument is usually the base value we are
+    // after
+    if (strct::isStructVal(e)) {
+      return exprToLlvm(ty, e->arg(0), ctx, dl);
```
agurfinkel, line 36: arg(0)

agurfinkel (commented):
some more comments. This will need more time to review. 
I was expecting a better design. This seems to be a band-aid on top of what we had before. Better, but needs more thought

lib/seahorn/CexExeGenerator.cc
danblitzhou: good catch, will use mem manager to extract base address instead

include/seahorn/CexExeGenerator.hh
danblitzhou: start address and size of each `memhavoc` invocation, maybe change to something like `m_memhavoc_args`?

lib/seahorn/CexExeGenerator.cc
```diff
@@ -0,0 +1,575 @@
...
+    out.os() << *m_harness;
+  else
+    WriteBitcodeToFile(*m_harness, out.os());
+  out.os().close();
+  out.keep();
+}
+
+template <class Trace>
+Constant *CexExeGenerator<Trace>::exprToConstant(Type *ty, Expr e) {
+  if (isOpX<TRUE>(e)) {
```
danblitzhou, line 454: currently looks like this function is "inferring" type of `LLVM::Constant` to convert to based on `e`, instead of enforcing the input `Type` directly; perhaps we should always try to create a `Constant` of type `ty` here, if somehow input e is not compatible, create a null value and print a debug message.

lib/seahorn/BmcPass.cc
```diff
@@ -360,9 +361,9 @@ class BmcPass : public llvm::ModulePass {
           if (CexFileRef.endswith(".ll") || CexFileRef.endswith(".bc")) {
             auto &tli = getAnalysis<TargetLibraryInfoWrapperPass>();
             auto const &dl = F.getParent()->getDataLayout();
-            BmcTraceWrapper<ZBmcTraceTy> trace_wrapper(trace);
-            dumpLLVMCex(trace_wrapper, CexFileRef, dl, tli.getTLI(F),
-                        F.getContext());
+            cexGen::CexExeGenerator<ZBmcTraceTy> cex(trace, dl, tli.getTLI(F),
```
agurfinkel, line 364: I think at this stage it is better to keep both old and new functionality and have a command line switch to decide which one to use.

For the new functionality, we do not want to port all of the existing one -- only what makes sense in the current context. However, there are users of old functionality so we can't just remove it.

lib/seahorn/CexExeGenerator.cc
agurfinkel: that's great

lib/seahorn/CexExeGenerator.cc
agurfinkel: Depends on the client (but this is now under your control). This does sound most reasonable that the output is of a type requested.

There is still a disconnect because types might have different representation in memory. In particular, the memory size of `i1` is `i8`. Probably the client must be aware of that and request appropriate type, or type cast as needed.

danblitzhou: this suggestion is mostly to resolve original author's concern; the idea is to leave the size of type under client control, so that when it calls `ConstantArray::get(AT, LLVMarray)`, element type of `AT` will not mismatch with `ty`; 
if the client happened to send mismatched Type `ty` and expression `e`, I hope new implementation will make it easier to debug

agurfinkel: great

agurfinkel (approved):
This looks good. Next is another pass on cleaning up the 'new' cex gen

PR #373, opened 2020-12-14
Inline assembly of integer swap

LinerSu:
Add implementation of handling (X86) byte swap lower intrinsic assembly.
The current implementation handles byte-swapping under `16`, `32`, `64` bits.
Note that: The current implementation could not handle assembly made by `i386`.


test/opsem2/bswap/bswap-ntohs-inline-asm_sat.c
```diff
@@ -0,0 +1,20 @@
+//; RUN: %sea "%s" 2>&1 | OutputCheck %s
+// CHECK: ^Error: assertion failed
+// CHECK: ^sat$
+
+#include "seahorn/seahorn.h"
+#include <arpa/inet.h>
```
priyasiddharth, line 6: is this available on all linux systems... or will we have to install in Continuous integration server which runs ubuntu?

test/opsem2/bswap/bswap-ntohs-inline-asm_sat.c
```diff
@@ -0,0 +1,20 @@
...
+#include <stdio.h>
+
+extern uint16_t nd_uint16_t(void);
+
+int main(int argc, char **argv) {
+  uint16_t a = nd_uint16_t();
+  uint16_t b = nd_uint16_t();
+  __VERIFIER_assume(a == b);
```
priyasiddharth, line 14: we can use assume(...). : https://sourcegraph.com/github.com/seahorn/seahorn@dev10/-/blob/include/seahorn/seahorn.h#L39

test/opsem2/bswap/bswap-ntohs-inline-asm_sat.c
```diff
@@ -0,0 +1,20 @@
...
+  uint16_t res_a = ntohs(a);
+  uint16_t res_b = ntohs(b);
+  __VERIFIER_assert(res_a != res_b);
```
priyasiddharth, line 17: we can use `sassert` only: see https://sourcegraph.com/github.com/seahorn/seahorn@dev10/-/blob/include/seahorn/seahorn.h#L43

`__VERIFIER_assert` is not a replacement for `sassert`. It is an instrinsic used for vacuity check

test/opsem2/bswap/bswap-ntohs-inline-asm_unsat.c
```diff
@@ -0,0 +1,21 @@
...
+#include "seahorn/seahorn.h"
+#include <arpa/inet.h>
+#include <stdio.h>
+
+extern uint16_t nd_uint16_t(void);
+
+int main(int argc, char **argv) {
+  uint16_t a = nd_uint16_t();
+  uint16_t b = nd_uint16_t();
+  __VERIFIER_assume(a == b);
```
priyasiddharth, line 15: same as above

test/opsem2/bswap/bswap-ntohs-inline-asm_unsat.c
```diff
@@ -0,0 +1,21 @@
...
+  uint16_t res_a = ntohs(a);
+  uint16_t res_b = ntohs(b);
+  __VERIFIER_assert(res_a == res_b);
```
priyasiddharth, line 18: same as above

test/opsem/bswap-ntohs-inline-asm_sat.ll
```diff
@@ -0,0 +1,81 @@
+; RUN: %seabmc "%s" 2>&1 | %oc %s
```
priyasiddharth, line 1: As far as i can tell... the `ll` file is an output for the `c` source. We only need one usually. In this case to test that `ror` --> bswap --> ir, I would assume we necessarily need the `c` file but not the `ll` file.

test/opsem/bswap-ntohs-inline-asm_unsat.ll
```diff
@@ -0,0 +1,81 @@
+; RUN: %seabmc "%s" 2>&1 | %oc %s
+; RUN: %seabmc --horn-bv2-lambdas --log=opsem3 "%s" 2>&1 | %oc %s
```
priyasiddharth, line 2: Do we need this file? Same comment as previous `ll` file.

lib/seahorn/BvOpSem2.cc
```diff
@@ -617,20 +634,42 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemVisitorBase {
...
+                            [](CallInst &CI) {
+                              return IntrinsicLowering::LowerToByteSwap(&CI);
+                            })) {
+          LOG("opsem", ERR << "Cannot handle inline assembly of integer swap: "
+                           << *CS.getInstruction());
+          return;
+        }
+      } else {
+        // Copy input to output
+        Expr readVal = lookup(*CS.getArgument(0));
```
priyasiddharth, line 670: For my own understanding,  if flag registers are clobbered, we don't
1. replace instruction with `bswap`... why? i.e. can we always use `byteswap` when we see `rorw`?
2. print a warning in the `else` branch. Should we let the user know something like "Cannot handle inline assembly: "?

priyasiddharth (changes requested):
Thank you for this change.

1.The core logic looks good. 
2. A few questions and comments (on unit tests). 
3. The CI tests are failing - some test in `opsem` directory.

test/opsem/bswap-ntohs-inline-asm_sat.ll
LinerSu: Arie suggests us to remain `ll` files. I made new updates and removed `c` files.

lib/seahorn/BvOpSem2.cc
LinerSu: 1. The x86 contains some specific intrinsics we need to handle. Otherwise, it won't work for several harnesses under a Linux machine.

agurfinkel:
@LinerSu commit message is not passing a check. Still waiting on the CI for the build

LinerSu:
> @LinerSu commit message is not passing a check. Still waiting on the CI for the build

Updated commit messages.

agurfinkel (approved):

PR #374, opened 2020-12-18
Remove personality if a function is empty

fshaked:
Needed when running on bitcode that was generated by rustc.

When a function definition is transformed to a function declaration, remove the personality, otherwise later passes fail.


agurfinkel:
@fshaked thanks!

PR #375, opened 2020-12-22
feat(opsem): track alloc/free for memory objects

priyasiddharth:
1. Generalize Tracking memory managers to track read, writes and allocations.

2. Add builtins for
isRead(...)
resetRead(...)
isAlloc(...)
free(...)

3. read instructions are currently not wired in opsem

### DIFF
--- a/lib/Analysis/SeaBuiltinsInfo.cc
+++ b/lib/Analysis/SeaBuiltinsInfo.cc
@@ -315,7 +315,6 @@ Function *SeaBuiltinsInfo::mkTrackingOnFn(Module &M) {
   auto FC = M.getOrInsertFunction(SEA_TRACKING_ON, Type::getVoidTy(C));
   auto *FN = dyn_cast<Function>(FC.getCallee());
   if (FN) {
-    FN->setDoesNotReadMemory();
     FN->setDoesNotThrow();
     FN->setDoesNotFreeMemory();
     FN->setDoesNotRecurse();
@@ -328,7 +327,6 @@ Function *SeaBuiltinsInfo::mkTrackingOffFn(Module &M) {
   auto FC = M.getOrInsertFunction(SEA_TRACKING_OFF, Type::getVoidTy(C));
   auto *FN = dyn_cast<Function>(FC.getCallee());
   if (FN) {
-    FN->setDoesNotReadMemory();
     FN->setDoesNotThrow();
     FN->setDoesNotFreeMemory();
     FN->setDoesNotRecurse();
@@ -344,6 +342,9 @@ Function *SeaBuiltinsInfo::mkFreeFn(Module &M) {
   if (FN) {
     FN->setDoesNotThrow();
     FN->setDoesNotRecurse();
+    // This only marks the memory as freed an does not have the semantics for
+    // actually freeing memory.
+    FN->setDoesNotFreeMemory();
     FN->addParamAttr(0, Attribute::NoCapture);
     // XXX maybe even add the following
     // FN->setDoesNotAccessMemory();
--- a/lib/Analysis/SeaBuiltinsInfo.cc
+++ b/lib/Analysis/SeaBuiltinsInfo.cc
@@ -23,8 +23,12 @@ using namespace llvm;
 #define SEA_BRANCH_SENTINEL "sea.branch_sentinel"
 #define SEA_IS_MODIFIED "sea.is_modified"
 #define SEA_RESET_MODIFIED "sea.reset_modified"
+#define SEA_IS_READ "sea.is_read"
+#define SEA_RESET_READ "sea.reset_read"
+#define SEA_IS_ALLOC "sea.is_alloc"
 #define SEA_TRACKING_ON "sea.tracking_on"
 #define SEA_TRACKING_OFF "sea.tracking_off"
+#define SEA_FREE "sea.free"
 
 SeaBuiltinsOp
 seahorn::SeaBuiltinsInfo::getSeaBuiltinOp(const llvm::CallBase &cb) const {
@@ -45,8 +49,12 @@ seahorn::SeaBuiltinsInfo::getSeaBuiltinOp(const llvm::CallBase &cb) const {
       .Case(SEA_BRANCH_SENTINEL, SBIOp::BRANCH_SENTINEL)
       .Case(SEA_IS_MODIFIED, SBIOp::IS_MODIFIED)
       .Case(SEA_RESET_MODIFIED, SBIOp::RESET_MODIFIED)
+      .Case(SEA_IS_READ, SBIOp::IS_READ)
+      .Case(SEA_RESET_READ, SBIOp::RESET_READ)
+      .Case(SEA_IS_ALLOC, SBIOp::IS_ALLOC)
       .Case(SEA_TRACKING_ON, SBIOp::TRACKING_ON)
       .Case(SEA_TRACKING_OFF, SBIOp::TRACKING_OFF)
+      .Case(SEA_FREE, SBIOp::FREE)
       .Default(SBIOp::UNKNOWN);
 }
 
@@ -76,10 +84,18 @@ llvm::Function *SeaBuiltinsInfo::mkSeaBuiltinFn(SeaBuiltinsOp op,
     return mkIsModifiedFn(M);
   case SBIOp::RESET_MODIFIED:
     return mkResetModifiedFn(M);
+  case SBIOp::IS_READ:
+    return mkIsReadFn(M);
+  case SBIOp::RESET_READ:
+    return mkResetReadFn(M);
+  case SBIOp::IS_ALLOC:
+    return mkIsAllocFn(M);
   case SBIOp::TRACKING_ON:
     return mkTrackingOnFn(M);
   case SBIOp::TRACKING_OFF:
     return mkTrackingOffFn(M);
+  case SBIOp::FREE:
+    return mkFreeFn(M);
   }
   llvm_unreachable(nullptr);
 }
@@ -163,6 +179,56 @@ Function *SeaBuiltinsInfo::mkResetModifiedFn(Module &M) {
                                   Type::getInt8PtrTy(C));
   auto *FN = dyn_cast<Function>(FC.getCallee());
   if (FN) {
+    FN->setDoesNotThrow();
+    FN->setDoesNotFreeMemory();
+    FN->setDoesNotRecurse();
+    FN->addParamAttr(0, Attribute::NoCapture);
+    // XXX maybe even add the following
+    // FN->setDoesNotAccessMemory();
+  }
+  return FN;
+}
+
+Function *SeaBuiltinsInfo::mkIsReadFn(Module &M) {
+  auto &C = M.getContext();
+  auto FC = M.getOrInsertFunction(SEA_IS_READ, Type::getInt1Ty(C),
+                                  Type::getInt8PtrTy(C));
+  auto *FN = dyn_cast<Function>(FC.getCallee());
+  if (FN) {
+    FN->setOnlyReadsMemory();
+    FN->setDoesNotThrow();
+    FN->setDoesNotFreeMemory();
+    FN->setDoesNotRecurse();
+    FN->addParamAttr(0, Attribute::NoCapture);
+    // XXX maybe even add the following
+    // FN->setDoesNotAccessMemory();
+  }
+  return FN;
+}
+
+Function *SeaBuiltinsInfo::mkResetReadFn(Module &M) {
+  auto &C = M.getContext();
+  auto FC = M.getOrInsertFunction(SEA_RESET_READ, Type::getVoidTy(C),
+                                  Type::getInt8PtrTy(C));
+  auto *FN = dyn_cast<Function>(FC.getCallee());
+  if (FN) {
+    FN->setDoesNotThrow();
+    FN->setDoesNotFreeMemory();
+    FN->setDoesNotRecurse();
+    FN->addParamAttr(0, Attribute::NoCapture);
+    // XXX maybe even add the following
+    // FN->setDoesNotAccessMemory();
+  }
+  return FN;
+}
+
+Function *SeaBuiltinsInfo::mkIsAllocFn(Module &M) {
+  auto &C = M.getContext();
+  auto FC = M.getOrInsertFunction(SEA_IS_ALLOC, Type::getInt1Ty(C),
+                                  Type::getInt8PtrTy(C));
+  auto *FN = dyn_cast<Function>(FC.getCallee());
+  if (FN) {
+    FN->setOnlyReadsMemory();
     FN->setDoesNotThrow();
     FN->setDoesNotFreeMemory();
     FN->setDoesNotRecurse();
@@ -266,6 +332,21 @@ Function *SeaBuiltinsInfo::mkTrackingOffFn(Module &M) {
     FN->setDoesNotThrow();
     FN->setDoesNotFreeMemory();
     FN->setDoesNotRecurse();
+  }
+  return FN;
+}
+
+Function *SeaBuiltinsInfo::mkFreeFn(Module &M) {
+  auto &C = M.getContext();
+  auto FC = M.getOrInsertFunction(SEA_FREE, Type::getVoidTy(C),
+                                  Type::getInt8PtrTy(C));
+  auto *FN = dyn_cast<Function>(FC.getCallee());
+  if (FN) {
+    FN->setDoesNotThrow();
+    FN->setDoesNotRecurse();
+    FN->addParamAttr(0, Attribute::NoCapture);
+    // XXX maybe even add the following
+    // FN->setDoesNotAccessMemory();
   }
   return FN;
 }
--- a/lib/Transforms/Scalar/PromoteMalloc.cc
+++ b/lib/Transforms/Scalar/PromoteMalloc.cc
@@ -55,8 +55,7 @@ public:
         if (!nv)
           nv = new AllocaInst(v->getType()->getPointerElementType(), addrSpace,
                               CS.getArgument(0), "malloc", &I);
-        IRBuilder<> Builder(F.getContext());
-        Builder.SetInsertPoint(&I);
+
         v->replaceAllUsesWith(nv);
 
         changed = true;
--- a/lib/Transforms/Scalar/PromoteMalloc.cc
+++ b/lib/Transforms/Scalar/PromoteMalloc.cc
@@ -53,9 +53,10 @@ public:
         }
 
         if (!nv)
-          nv = new AllocaInst(v->getType()->getPointerElementType(),
-                              addrSpace, CS.getArgument(0), "malloc", &I);
-
+          nv = new AllocaInst(v->getType()->getPointerElementType(), addrSpace,
+                              CS.getArgument(0), "malloc", &I);
+        IRBuilder<> Builder(F.getContext());
+        Builder.SetInsertPoint(&I);
         v->replaceAllUsesWith(nv);
 
         changed = true;
--- a/lib/Transforms/Scalar/PromoteVerifierCalls.cc
+++ b/lib/Transforms/Scalar/PromoteVerifierCalls.cc
@@ -51,8 +51,12 @@ bool PromoteVerifierCalls::runOnModule(Module &M) {
   m_assert_if = SBI.mkSeaBuiltinFn(SBIOp::ASSERT_IF, M);
   m_is_modified = SBI.mkSeaBuiltinFn(SBIOp::IS_MODIFIED, M);
   m_reset_modified = SBI.mkSeaBuiltinFn(SBIOp::RESET_MODIFIED, M);
+  m_is_read = SBI.mkSeaBuiltinFn(SBIOp::IS_READ, M);
+  m_reset_read = SBI.mkSeaBuiltinFn(SBIOp::RESET_READ, M);
+  m_is_alloc = SBI.mkSeaBuiltinFn(SBIOp::IS_ALLOC, M);
   m_tracking_on = SBI.mkSeaBuiltinFn(SBIOp::TRACKING_ON, M);
   m_tracking_off = SBI.mkSeaBuiltinFn(SBIOp::TRACKING_OFF, M);
+  m_free = SBI.mkSeaBuiltinFn(SBIOp::FREE, M);
 
   // XXX DEPRECATED
   // Do not keep unused functions in llvm.used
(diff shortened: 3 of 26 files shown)


priyasiddharth:
Looking at test failures...

priyasiddharth:
Ready for review...

lib/Analysis/SeaBuiltinsInfo.cc
```diff
@@ -270,6 +336,21 @@ Function *SeaBuiltinsInfo::mkTrackingOffFn(Module &M) {
...
 
+Function *SeaBuiltinsInfo::mkFreeFn(Module &M) {
+  auto &C = M.getContext();
+  auto FC = M.getOrInsertFunction(SEA_FREE, Type::getVoidTy(C),
+                                  Type::getInt8PtrTy(C));
+  auto *FN = dyn_cast<Function>(FC.getCallee());
+  if (FN) {
+    FN->setDoesNotThrow();
+    FN->setDoesNotRecurse();
+    FN->addParamAttr(0, Attribute::NoCapture);
```
agurfinkel, line 347: should also be marked with `doesNotFreeMemory()`. While you can the function `free`, a more appropriate name is `mark_free`. The function marks a memory region as freed, but it does not free the memory region itself. The freeing will be done by calling libc `free` or C++ `delete`.

lib/Transforms/Scalar/PromoteVerifierCalls.cc
```diff
@@ -209,6 +213,33 @@ bool PromoteVerifierCalls::runOnFunction(Function &F) {
       if (cg)
         (*cg)[&F]->addCalledFunction(ci, (*cg)[ci->getCalledFunction()]);
 
+      I.replaceAllUsesWith(ci);
+      toKill.push_back(&I);
+    } else if (fn && (fn->getName().equals("sea_is_read"))) {
+      IRBuilder<> Builder(F.getContext());
```
agurfinkel, line 219: the bodies of these if statements are all the same, why not refactor the code into a (lambda) function and set a map between function name and function that replaces it. So the code looks like that:
```
if (map.count(fn->getName())) {
   replace(fn, map.at(fn->getName());
 }
```

lib/seahorn/BvOpSem2Context.hh
```diff
@@ -15,6 +15,14 @@
 namespace seahorn {
 namespace details {
 
+// This enumerates the different kinds of metadata memory used in Tracking
+// memory.
+enum MetadataKind {
```
agurfinkel, line 20: use `enum class`

lib/seahorn/BvOpSem2Context.hh
```diff
@@ -698,6 +706,9 @@ public:
   /// \brief returns a constant that represents zero-initialized memory region
   virtual MemValTy zeroedMemory() const = 0;
 
+  /// \brief returns a constant that represents zero-initialized memory region
+  virtual MemValTy setMemory(unsigned int val) const = 0;
```
agurfinkel, line 710: comment does not match method name

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -558,6 +558,11 @@ class FatMemManager : public MemManagerCore {
                     m_slot1.zeroedMemory());
   }
 
+  FatMemValTy setMemory(unsigned int val) const {
```
agurfinkel, line 561: is this correct? I am not sure what `setMemory` supposed to do, but it is strange that all slots are rewritten using the same value

lib/seahorn/BvOpSem2RawMemMgr.cc
```diff
@@ -795,7 +795,7 @@ bool RawMemManagerCore::ignoreAlignment() const { return m_ignoreAlignment; }
 
 RawMemManagerCore::MemValTy
 RawMemManagerCore::resetModified(PtrTy p, RawMemManagerCore::MemValTy mem) {
-  LOG("opsem", WARN << "resetModified() not implemented!\n");
+  LOG("opsem", WARN << "resetMetadata() not implemented!\n");
```
agurfinkel, line 798: should be `ERR` to be consistent with other usage

lib/seahorn/BvOpSem2RawMemMgr.cc
```diff
@@ -816,6 +816,12 @@ RawMemManagerCore::getGlobalVariableInitValue(const GlobalVariable &gv) {
 bool RawMemManagerCore::isPtrTyVal(Expr e) {
   return (!e || !strct::isStructVal(e));
 }
+RawMemManagerCore::MemValTy
+RawMemManagerCore::setMemory(unsigned int val) const {
+  assert(llvm::Log2_64(val) + 1 <= wordSizeInBits());
```
agurfinkel, line 821: there should be a warning printed if assertion fails. This can be hard to trace in non-debug mode

lib/seahorn/BvOpSem2RawMemMgr.cc
```diff
@@ -816,6 +816,12 @@ RawMemManagerCore::getGlobalVariableInitValue(const GlobalVariable &gv) {
 bool RawMemManagerCore::isPtrTyVal(Expr e) {
   return (!e || !strct::isStructVal(e));
 }
+RawMemManagerCore::MemValTy
+RawMemManagerCore::setMemory(unsigned int val) const {
```
agurfinkel, line 820: the name `setMemory` is confusing because it is interpreted either as a setter (which it is not), or something that changes memory region (which it is not). The implementation seems to be a wrapper over `FilledMemory` that returns initialized memory region. Maybe rename all into `InitMem` or something similar to reflect the intention of the function.

lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
```diff
@@ -5,7 +5,7 @@ namespace details {
 
 static const unsigned int g_MetadataBitWidth = 8;
```
agurfinkel, line 6: why are all of these globals? Can we merge them into class definition?

lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
```diff
@@ -113,9 +146,11 @@ TrackingRawMemManager::MemValTy TrackingRawMemManager::storeValueToMem(
   const unsigned byteSz =
       m_sem.getTD().getTypeStoreSize(const_cast<llvm::Type *>(&ty));
   ExprFactory &efac = ptr->efac();
-  // TODO: use zeroed memory on m_main, m_metadata instead of explicit
+  // TODO: use zeroed memory on m_main, m_w_metadata instead of explicit
```
agurfinkel, line 149: still a TODO?

lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
```diff
@@ -113,9 +146,11 @@ TrackingRawMemManager::MemValTy TrackingRawMemManager::storeValueToMem(
...
   // init
   MemValTy res(m_ctx.alu().ui(0UL, wordSizeInBits()),
```
agurfinkel, line 151: I am confused about this initialization. This should only be used when `store` operation fails. Should it be just a null indicating that operation failed?

lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
```diff
@@ -74,36 +102,41 @@ TrackingRawMemManager::MemValTy TrackingRawMemManager::MemFill(
...
 }
 TrackingRawMemManager::MemValTy TrackingRawMemManager::MemSet(
     TrackingRawMemManager::PtrTy ptr, Expr _val, unsigned int len,
     TrackingRawMemManager::MemValTy mem, uint32_t align) {
   RawMemValTy rawVal = m_main.MemSet(ptr, _val, len, mem.getRaw(), align);
-  return MemValTy(rawVal, mem.getMetadata());
+  return MemValTy(rawVal, mem.getMetadata(READ), mem.getMetadata(WRITE),
+                  mem.getMetadata(ALLOC));
 }
 TrackingRawMemManager::MemValTy TrackingRawMemManager::storeValueToMem(
```
agurfinkel, line 141: this function should be factored out into the mixin. It simply dispatches into appropriate store function based on type of value.

lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
```diff
@@ -213,29 +250,32 @@ Expr TrackingRawMemManager::loadIntFromMem(TrackingRawMemManager::PtrTy ptr,
...
+                                   len, memIn.getMetadata(kind),
+                                   m_metadata_map[kind]->wordSzInBytes()));
 }
-Expr TrackingRawMemManager::getMetaData(PtrTy ptr, MemValTy memIn,
-                                        unsigned int byteSz) {
-  return m_metadata.loadIntFromMem(ptr, memIn.getMetadata(), byteSz,
-                                   0 /* TODO: fix */);
+Expr TrackingRawMemManager::getMetaData(MetadataKind kind, PtrTy ptr,
+                                        MemValTy memIn, unsigned int byteSz) {
+  return m_metadata_map.at(kind)->loadIntFromMem(ptr, memIn.getMetadata(kind),
```
agurfinkel, line 277: should not the load be assumed to be aligned?

lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
```diff
@@ -255,8 +295,10 @@ Expr TrackingRawMemManager::coerce(Expr sort, Expr val) {
     assert(sort->arity() == g_num_slots);
     kids.push_back(m_main.coerce(sort->arg(0), val->arg(0)));
     // when havocing a value; don't havoc(ignore) value for metadata memory,
-    // instead use 'zeroed' memory.
-    kids.push_back(m_metadata.zeroedMemory());
+    // instead intialize memory to a constant value.
+    kids.push_back(m_metadata_map.at(READ)->zeroedMemory());
+    kids.push_back(m_metadata_map.at(WRITE)->zeroedMemory());
+    kids.push_back(m_metadata_map.at(ALLOC)->setMemory(1U));
```
agurfinkel, line 301: interesting choice for initial value for havoced memory...

lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
```diff
@@ -327,16 +371,22 @@ TrackingRawMemManager::mkStackPtr(unsigned int offset) {
...
+  LOG("opsem", WARN << "isMetadataSet() not implemented!\n");
   return Expr();
 }
 TrackingRawMemManager::MemValTy
-TrackingRawMemManager::resetModified(TrackingRawMemManager::PtrTy p,
+TrackingRawMemManager::resetMetadata(MetadataKind kind,
+                                     TrackingRawMemManager::PtrTy p,
                                      TrackingRawMemManager::MemValTy mem) {
-  LOG("opsem", WARN << "resetModified() not implemented!\n");
+  LOG("opsem", WARN << "resetMetadata() not implemented!\n");
```
agurfinkel, line 389: all `not implemented` messages should go into `ERR`

lib/seahorn/BvOpSem2TrackingRawMemMgr.hh
```diff
@@ -40,23 +45,53 @@ public:
...
-    MemValTyImpl(RawMemValTy &&raw_val, Expr &&metadata_val) {
+    MemValTyImpl(RawMemValTy &&raw_val, RawMemValTy &&r_metadata_val,
+                 RawMemValTy &&w_metadata_val, RawMemValTy &&a_metadata_val) {
       assert(!strct::isStructVal(raw_val));
-      assert(!strct::isStructVal(metadata_val));
-      m_v = strct::mk(std::move(raw_val), std::move(metadata_val));
+      assert(!strct::isStructVal(r_metadata_val));
+      assert(!strct::isStructVal(w_metadata_val));
+      assert(!strct::isStructVal(a_metadata_val));
+      llvm::SmallVector<RawMemValTy, 4> kids;
```
agurfinkel, line 54: can use stack data structure like `llvm::ArrayRef` or `std::array`

lib/seahorn/BvOpSem2TrackingRawMemMgr.hh
```diff
@@ -66,19 +101,36 @@ public:
 
     RawMemValTy getRaw() { return strct::extractVal(m_v, 0); }
 
-    Expr getMetadata() { return strct::extractVal(m_v, 1); }
+    Expr getMetadata(MetadataKind kind) {
+      return strct::extractVal(m_v, kind + 1);
```
agurfinkel, line 105: fragile. order of metadata and order of enum can be different. at least comment somewhere that they should match. But better, add a function that converts kind to index in tracking memory manager and use that function.

lib/seahorn/BvOpSem2TrackingRawMemMgr.hh
```diff
@@ -225,15 +282,19 @@ public:
 
   MemValTy zeroedMemory() const;
 
+  MemValTy setMemory(unsigned int val) const;
```
agurfinkel, line 285: I know believe that this should be `initializedMemory` and `zeroedMemory` should be `zeroInitializedMemory`
Note that `zeroinitialized` is a keyword LLVM uses to mark a global memory region `zeroinitialized`, so the name will make sense to anyone familiar with llvm

test/opsem2/trackingmem/freed_unsat.c
```diff
@@ -0,0 +1,21 @@
...
+extern bool sea_is_alloc(char *);
+extern void sea_tracking_on();
+extern void sea_tracking_off();
+
+extern int nd_int();
+int main() {
+  sea_tracking_on();
+  int *a = malloc(1024);
+  *a = nd_int();
+  free(a);
```
agurfinkel, line 18: I missed where `free` is mapped to `sea.free`...

agurfinkel (changes requested):
Looks very good. A few minor things.

lib/seahorn/BvOpSem2TrackingRawMemMgr.hh
priyasiddharth: Added a comment for now. This can be made less fragile but I want to solve it at the right level. 

e.g. Ideally there should be one class that deals with arranging metadata memory and all client  - MemMgr, MemToExpr classes should use its method rather than knowing internal details. We are not there yet and changing it as part of this feature appears out of scope to me.

lib/seahorn/BvOpSem2TrackingRawMemMgr.hh
priyasiddharth: done

lib/seahorn/BvOpSem2TrackingRawMemMgr.hh
```diff
@@ -40,23 +45,53 @@ public:
...
+      kids.push_back(std::move(a_metadata_val));
+
+      m_v = strct::mk(kids);
+    }
+
+    MemValTyImpl(MetadataKind kind, const MemValTyImpl &orig_val,
+                 const RawMemValTy &raw_val) {
+      llvm::SmallVector<RawMemValTy, 4> kids;
+      for (auto i = 0; i < orig_val.toExpr()->arity(); i++) {
+        kids.push_back(i == kind + 1 ? raw_val : orig_val.toExpr()->arg(i));
```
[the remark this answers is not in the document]
priyasiddharth: added comment.

priyasiddharth (commented):
Please take a look

agurfinkel:
are the build failures current @priyasiddharth ?

lib/seahorn/BvOpSem2Context.hh
agurfinkel: Effective Modern C++, Item 10, might be useful to review

test/opsem2/trackingmem/freed_unsat.c
agurfinkel: we should not map `free` to `sea.free`. They have different semantics... This also interferes with other LLVM passes that are aware of semantics of `free`. Note that there is also `delete` in C++ that should be treated similarly.

agurfinkel:
there are still some unresolved/unanswered comments from last review

lib/Transforms/Scalar/PromoteMalloc.cc
```diff
@@ -53,9 +53,10 @@ class PromoteMalloc : public FunctionPass {
         }
 
         if (!nv)
-          nv = new AllocaInst(v->getType()->getPointerElementType(),
-                              addrSpace, CS.getArgument(0), "malloc", &I);
-
+          nv = new AllocaInst(v->getType()->getPointerElementType(), addrSpace,
+                              CS.getArgument(0), "malloc", &I);
+        IRBuilder<> Builder(F.getContext());
```
agurfinkel, line 58: this looks like dead code...

lib/seahorn/BvOpSem2Context.hh
```diff
@@ -17,6 +17,7 @@ namespace details {
 
 // This enumerates the different kinds of metadata memory used in Tracking
 // memory.
+// TODO: move to enum class when int value is not needed.
```
agurfinkel, line 20: where is the implicit cast to underlying type is used?

lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
```diff
@@ -642,17 +637,16 @@ unsigned int ExtraWideMemManager<T>::getMetaDataMemWordSzInBits() {
...
-                                      ExtraWideMemManager::PtrTy ptr,
-                                      ExtraWideMemManager::MemValTy mem) {
-  if (!m_ctx.isTrackingOn()) {
+typename ExtraWideMemManager<T>::MemValTy ExtraWideMemManager<T>::setMetadata(
+    MetadataKind kind, ExtraWideMemManager::PtrTy ptr,
+    ExtraWideMemManager::MemValTy mem, unsigned val) {
+  if (!m_ctx.isTrackingOn() && kind != ALLOC) {
     LOG("opsem.memtrack.verbose",
-        errs() << "Ignoring resetMetadata();Memory tracking is off"
+        errs() << "Ignoring setMetadata();Memory tracking is off"
```
agurfinkel, line 645: probably should be `WARN`

lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
```diff
@@ -274,8 +265,9 @@ TrackingRawMemManager::memsetMetaData(MetadataKind kind, PtrTy ptr,
 }
 Expr TrackingRawMemManager::getMetaData(MetadataKind kind, PtrTy ptr,
                                         MemValTy memIn, unsigned int byteSz) {
-  return m_metadata_map.at(kind)->loadIntFromMem(ptr, memIn.getMetadata(kind),
-                                                 byteSz, 0 /* TODO: fix */);
+  return m_metadata_map.at(kind)->loadIntFromMem(
```
agurfinkel, line 268: this should be aligned load

lib/seahorn/BvOpSem2TrackingRawMemMgr.hh
```diff
@@ -67,18 +68,17 @@ public:
       assert(!strct::isStructVal(r_metadata_val));
       assert(!strct::isStructVal(w_metadata_val));
       assert(!strct::isStructVal(a_metadata_val));
-      llvm::SmallVector<RawMemValTy, 4> kids;
-      kids.push_back(std::move(raw_val));
-      kids.push_back(std::move(r_metadata_val));
-      kids.push_back(std::move(w_metadata_val));
-      kids.push_back(std::move(a_metadata_val));
-
+      std::array<RawMemValTy, 4> kids = {raw_val, r_metadata_val,
```
agurfinkel, line 71: `std::move` got lost in conversion from `SmallVector` to `array`

lib/seahorn/BvOpSem2TrackingRawMemMgr.hh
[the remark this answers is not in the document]
agurfinkel: also comment what is `i == kind + 1` -- this depends on internal representation of meta-data fields.
update `for` loop per my comment -- only call `arity()` ones, use pre-inc instead of post-inc. These are my nits, but they are not hard to change...

tools/seahorn/seahorn.cpp
```diff
@@ -316,8 +316,8 @@ int main(int argc, char **argv) {
   }
   pass_manager.add(new seahorn::RemoveUnreachableBlocksPass());
 
-  pass_manager.add(seahorn::createPromoteMallocPass());
   pass_manager.add(seahorn::createPromoteVerifierCallsPass());
+  pass_manager.add(seahorn::createPromoteMallocPass());
```
agurfinkel, line 320: hmm... this might be masking the problem of mapping `free` to `sea.free`. The order of these two passes should not matter in general...

agurfinkel (changes requested):
I've lost quite a bit of context, so just looking for things that stand out. 
Looks good overall. I only have minor things + build failure...

lib/Transforms/Scalar/PromoteMalloc.cc
priyasiddharth: removed change... this was for debugging.

lib/seahorn/BvOpSem2Context.hh
priyasiddharth: When changing MemValTy to Expression and vice versa

tools/seahorn/seahorn.cpp
priyasiddharth: removed change... was debugging and forgot to remove.

test/opsem2/trackingmem/freed_unsat.c
agurfinkel: The semantics of `sea.free()` is to mark a memory region as freed. The memory region itself, is not freed until the call to `free()`. The semantics of `free()` is to release the address space to be allocated again.

We can, for example, provide an implementation of `free()` that first calls `sea.free()` to mark memory as freed, and then calls the builtin `free()`.  We do similar things for functions like `memcmp` that have builtin versions.

The reason why we cannot change `free()` to any other name is that both `malloc()` and `free()` are understood by LLVM. That is, they are builtin functions with well defined semantics.

To me, `sea.free` is more like `llvm.lifetime.end`, if that analogy helps.

agurfinkel:
build still fails. some prior comments are still not resolved. semantics of `sea.free` might need face-to-face discussion.

priyasiddharth (commented):
Build failure may be due to https://github.com/seahorn/sea-dsa/pull/111 .. can only verify after that is checked-in.

lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
agurfinkel: please add comment in the code for that. Probably should expose this functionality eventually, but I'm ok with the current solution

priyasiddharth: Done

agurfinkel:
los track. Am I the bottleneck? I see CI is failing, maybe it needs to be re-triggered?

test/opsem2/trackingmem/freed_unsat.c
priyasiddharth: If changing `free` to `sea.free` is blocking this change then we need to discuss this face to face. Otherwise, I can submit it. The cause of build failure is well understood and will be fixed with next commit. I want to send a final cleanup commit once the above discussion is concluded.

priyasiddharth: Added a `sea_free` user facing function which maps to `sea.free` builtin.

priyasiddharth:
This is ready for review.

I can't find an easy way to look at open review comments in a single list. If I missed any, please highlight if possible.

Once this is approved. I will rebase, squash the commits and submit to `dev10`.

agurfinkel (approved):

PR #377, opened 2020-12-23
refactor(cex): cleanup new cex gen system; add tests and python pipeline

danblitzhou:
- Cleaned up CexExeGenerator to use `InstVisitor` implementation for collecting nd function values
- added new python command pipeline `CexGen`; given a source (C or IR) and counterexample harness, usage: `sea cex <options> <source> <harness> -o <outfile>`; if -o is not provided, an executable name will be generated in `/tmp/` with default name format: `<target_name>_debug.exe`; I did not add seahorn solve step to the pipeline since right now there are many different ways: `bpf`, `fpf`, `pf`... each with different options, from my personal dev experience I find it more flexible to do seahorn solve step separately
- added back `cex` tests, need insight on `cex_mem_fail.c` and `cex_non_linear_arith.fail.c` about how to reproduce fail in meaningful way.

### DIFF
--- a/.github/workflows/test-seahorn-docker.yml
+++ b/.github/workflows/test-seahorn-docker.yml
@@ -39,5 +39,8 @@ jobs:
       - name: Run mcfuzz tests
         run: docker run -v $(pwd):/host seahorn_container /bin/bash -c "cd seahorn/share/seahorn && lit -a --time-tests --max-time=1200 test/mcfuzz"
       - name: Run smc tests
-        run: docker run -v $(pwd):/host seahorn_container /bin/bash -c "cd seahorn/share/seahorn && lit -a --time-tests --max-time=1200 test/smc" 
-
+        run: docker run -v $(pwd):/host seahorn_container /bin/bash -c "cd seahorn/share/seahorn && lit -a --time-tests --max-time=1200 test/smc"
+      - name: Run cex tests
+        run: docker run -v $(pwd):/host seahorn_container /bin/bash -c "cd seahorn/share/seahorn && lit -a --time-tests --max-time=1200 test/cex"
+      - name: Run cex tests with new cexgen
+        run: docker run -v $(pwd):/host seahorn_container /bin/bash -c "cd seahorn/share/seahorn && lit -a --time-tests --max-time=1200 -D cex_bmc_opts=--horn-bmc-cexgen test/cex"
--- a/include/seahorn/CexExeGenerator.hh
+++ b/include/seahorn/CexExeGenerator.hh
@@ -13,6 +13,7 @@
 #include "boost/algorithm/string/replace.hpp"
 
 #include "seahorn/Bmc.hh"
+#include "seahorn/CallUtils.hh"
 #include "seahorn/Expr/ExprLlvm.hh"
 #include "seahorn/Expr/ExprMemMap.hh"
 #include "seahorn/Expr/ExprOpBinder.hh"
@@ -59,12 +60,8 @@ template <class Trace> class CexExeGenerator {
 
   // map function calls to return value(s)
   ValueMap<const Function *, ExprVector> m_func_val_map;
-  // list of <start addr, size> of havoc pointers
-  std::vector<std::pair<Expr, Expr>> m_memhavoc_args;
-
-  /// \brief fills m_memhavoc_args and m_func_val_map
-  void storeMemHavoc(unsigned loc, const Function *func, ImmutableCallSite cs,
-                     ImmutableCallSite prevCS);
+  // list of <start addr, size> of havoc non det pointers
+  SmallVector<std::pair<Expr, Expr>, 20> m_nondet_ptrs;
 
   /// \brief fills func val map, dsa data and memhavoc ptr info
   void storeDataFromTrace();
@@ -98,6 +95,24 @@ public:
   }
 
   void saveCexModuleToFile(llvm::StringRef CexFile);
+
+  const DataLayout &getDataLayout() { return m_dl; }
+
+  const TargetLibraryInfo &getTargetLibraryInfo() { return m_tli; }
+
+  Trace &trace() { return m_trace; }
+
+  /// \brief Add int or pointer value \p val to list of return values of
+  /// Function \p func
+  void addValueToFunc(const Function *func, Expr val) {
+    m_func_val_map[func].push_back(val);
+  }
+
+  /// \brief Add pointer with starting address \p start and size of \p size to
+  /// list of pointers with non-det content
+  void addNonDetPtr(Expr start, Expr size) {
+    m_nondet_ptrs.emplace_back(std::make_pair(start, size));
+  }
 };
 
 } // namespace cexGen
--- a/lib/seahorn/BvOpSem2Context.hh
+++ b/lib/seahorn/BvOpSem2Context.hh
@@ -333,6 +333,8 @@ public:
 
   Expr ptrToAddr(Expr p) override;
 
+  Expr getRawMem(Expr p) override;
+
   void resetSolver();
   void addToSolver(const Expr e);
   boost::tribool solve();
@@ -729,6 +731,14 @@ public:
   /// \brief given a properly encoded pointer Expr \p p , return the raw
   /// expression representing memory address only
   virtual Expr ptrToAddr(Expr p) = 0;
+
+  /// \brief given an Expression \p e , return true if \p e has expected
+  /// encoding of a MemValTy
+  virtual bool isMemVal(Expr e) = 0;
+
+  /// \brief given a properly encoded memory map Expr \p p , return the base
+  /// expression representing raw memory only
+  virtual Expr getRawMem(Expr p) = 0;
 };
 
 OpSemMemManager *mkRawMemManager(Bv2OpSem &sem, Bv2OpSemContext &ctx,
--- a/lib/seahorn/BvOpSem2Context.hh
+++ b/lib/seahorn/BvOpSem2Context.hh
@@ -724,17 +724,9 @@ public:
   /// \brief reset memory modified state; used in conjuction with isModified
   virtual MemValTy resetModified(PtrTy p, MemValTy mem) = 0;
 
-  /// \brief given an Expression \p e , return true if \p e has expected
-  /// encoding of a PtrTyImpl
-  virtual bool isPtrTyVal(Expr e) = 0;
-
   /// \brief given a properly encoded pointer Expr \p p , return the raw
   /// expression representing memory address only
   virtual Expr ptrToAddr(Expr p) = 0;
-
-  /// \brief given an Expression \p e , return true if \p e has expected
-  /// encoding of a MemValTy
-  virtual bool isMemVal(Expr e) = 0;
 
   /// \brief given a properly encoded memory map Expr \p p , return the base
   /// expression representing raw memory only
--- a/lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
+++ b/lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
@@ -106,12 +106,12 @@ ExtraWideMemManager<T>::getAddressable(ExtraWideMemManager::PtrTy p) const {
 }
 
 template <class T> bool ExtraWideMemManager<T>::isPtrTyVal(Expr e) const {
-  return strct::isStructVal(e) && e->arity() == g_num_slots;
+  return e && strct::isStructVal(e) && e->arity() == g_num_slots;
 }
 
 template <class T> bool ExtraWideMemManager<T>::isMemVal(Expr e) const {
   // Our base is a struct of three exprs
-  return strct::isStructVal(e) && e->arity() == 3;
+  return e && strct::isStructVal(e) && e->arity() == g_num_slots;
 }
 
 template <class T>
--- a/lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
+++ b/lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
@@ -109,6 +109,11 @@ template <class T> bool ExtraWideMemManager<T>::isPtrTyVal(Expr e) const {
   return strct::isStructVal(e) && e->arity() == g_num_slots;
 }
 
+template <class T> bool ExtraWideMemManager<T>::isMemVal(Expr e) const {
+  // Our base is a struct of three exprs
+  return strct::isStructVal(e) && e->arity() == 3;
+}
+
 template <class T>
 Expr ExtraWideMemManager<T>::isDereferenceable(ExtraWideMemManager::PtrTy p,
                                                Expr byteSz) {
--- a/lib/seahorn/BvOpSem2FatMemMgr.cc
+++ b/lib/seahorn/BvOpSem2FatMemMgr.cc
@@ -580,12 +580,12 @@ public:
 
   bool isPtrTyVal(Expr e) const {
     // struct with raw ptr + fat slots
-    return strct::isStructVal(e) && e->arity() <= (1 + g_maxFatSlots);
+    return e && strct::isStructVal(e) && e->arity() <= (1 + g_maxFatSlots);
   }
 
   bool isMemVal(Expr e) const {
-    // struct with 3 fields
-    return (strct::isStructVal(e) && e->arity() == 3);
+    // struct with raw ptr + fat slots
+    return e && strct::isStructVal(e) && e->arity() == (1 + g_maxFatSlots);
   }
 };
 
--- a/lib/seahorn/BvOpSem2FatMemMgr.cc
+++ b/lib/seahorn/BvOpSem2FatMemMgr.cc
@@ -45,6 +45,7 @@ public:
     Expr v() const { return m_v; }
     Expr toExpr() const { return v(); }
     explicit operator Expr() const { return toExpr(); }
+    Expr getRaw() { return strct::extractVal(m_v, 0); }
   };
 
   using FatMemTag = MemoryFeatures::FatMem_tag;
@@ -580,6 +581,11 @@ public:
   bool isPtrTyVal(Expr e) const {
     // struct with raw ptr + fat slots
     return strct::isStructVal(e) && e->arity() <= (1 + g_maxFatSlots);
+  }
+
+  bool isMemVal(Expr e) const {
+    // struct with 3 fields
+    return (strct::isStructVal(e) && e->arity() == 3);
   }
 };
 
--- a/lib/seahorn/BvOpSem2FatMemMgr.cc
+++ b/lib/seahorn/BvOpSem2FatMemMgr.cc
@@ -580,7 +580,7 @@ public:
 
   bool isPtrTyVal(Expr e) const {
     // struct with raw ptr + fat slots
-    return e && strct::isStructVal(e) && e->arity() <= (1 + g_maxFatSlots);
+    return e && strct::isStructVal(e) && e->arity() == (1 + g_maxFatSlots);
   }
 
   bool isMemVal(Expr e) const {
--- a/lib/seahorn/BvOpSem2RawMemMgr.cc
+++ b/lib/seahorn/BvOpSem2RawMemMgr.cc
@@ -814,11 +814,11 @@ RawMemManagerCore::getGlobalVariableInitValue(const GlobalVariable &gv) {
 }
 
 bool RawMemManagerCore::isPtrTyVal(Expr e) {
-  return (!e || !strct::isStructVal(e));
+  return (e && !strct::isStructVal(e));
 }
 
 bool RawMemManagerCore::isMemVal(Expr e) {
-  return (!e || !strct::isStructVal(e));
+  return (e && !strct::isStructVal(e));
 }
 
 } // namespace details
--- a/lib/seahorn/BvOpSem2RawMemMgr.cc
+++ b/lib/seahorn/BvOpSem2RawMemMgr.cc
@@ -817,5 +817,9 @@ bool RawMemManagerCore::isPtrTyVal(Expr e) {
   return (!e || !strct::isStructVal(e));
 }
 
+bool RawMemManagerCore::isMemVal(Expr e) {
+  return (!e || !strct::isStructVal(e));
+}
+
 } // namespace details
 } // namespace seahorn
--- a/lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
+++ b/lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
@@ -35,6 +35,11 @@ bool TrackingRawMemManager::isPtrTyVal(Expr e) const {
   return !e || !strct::isStructVal(e);
 }
 
+bool TrackingRawMemManager::isMemVal(Expr e) const {
+  // raw and metadata
+  return strct::isStructVal(e) && e->arity() == 2;
+}
+
 Expr TrackingRawMemManager::isDereferenceable(TrackingRawMemManager::PtrTy p,
                                               Expr byteSz) {
   // isDereferenceable should never be called in a 'RawMemMgr'
--- a/lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
+++ b/lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
@@ -32,12 +32,12 @@ TrackingRawMemManager::getAddressable(TrackingRawMemManager::PtrTy p) const {
 
 bool TrackingRawMemManager::isPtrTyVal(Expr e) const {
   // same PtrTy as RawMemManager
-  return !e || !strct::isStructVal(e);
+  return e && !strct::isStructVal(e);
 }
 
 bool TrackingRawMemManager::isMemVal(Expr e) const {
   // raw and metadata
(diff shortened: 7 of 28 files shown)


agurfinkel:
it is good to cleanup commit history when submitting a PR, but don't squash everything into one commit. The history should be meaningful. For example, changes to python scripts should be separate from other changes

py/sea/commands.py
```diff
@@ -770,7 +770,7 @@ def run (self, args, extra):
         self.harnessCmd = sea.ExtCmd (eharness)
         if args.cpu is None: args.cpu = 10
 
-        retval = self.harnessCmd.run (args, [eharness], encoding='utf-8', errors='replace', stdout=subprocess.PIPE)
+        retval = self.harnessCmd.run (args, [], encoding='utf-8', errors='replace', stdout=subprocess.PIPE)
```
agurfinkel, line 773: is this intentional? (`eharness` has been removed)

py/sea/commands.py
```diff
@@ -1614,6 +1614,88 @@ def run (self, args, extra):
 
         return self.seainspectCmd.run (args, argv)
 
+class CexGen(sea.LimitedCmd):
```
agurfinkel, line 1617: good idea to split it into separate file. The file `commands.py` is getting too large. See for example `yama` command

py/sea/commands.py
```diff
@@ -1614,6 +1614,88 @@ def run (self, args, extra):
...
+        if args.out_file is None:
+            target = os.path.basename(args.in_files[0]).split(sep='.')[0]
+            args.out_file = "/tmp/{}_debug.exe".format(target)
+
+        linkRtCmd = LinkRt()
+        self.clangCmd = linkRtCmd
+        cmds.append(linkRtCmd)
+
+        pipeline = sea.SeqCmd('', '', cmds)
+        argv = ['-m64', '-g']
```
agurfinkel, line 1686: why always `m64`? Maybe check based on the platform or let user control

py/sea/commands.py
```diff
@@ -1614,6 +1614,88 @@ def run (self, args, extra):
...
+        argv.extend(extra)
+        res = pipeline.run(args, argv)
+        if res:
+            print("\n Executable Counterexample generation failed...")
```
agurfinkel, line 1690: there should not be a new line before print. Only after. If it is required, usually indicates an issue elsewhere

test/cex/cex_mem.c
```diff
@@ -14,10 +17,11 @@ struct st {
 };
 
 extern struct st *nd_st(void);
+extern void memhavoc(void *ptr, size_t size);
 
 int main(int argc, char**argv) {
-    struct st *p;
-    p = nd_st();
+    struct st *p = malloc(sizeof(struct st));
+    memhavoc(p, sizeof(struct st));
```
agurfinkel, line 24: seems like you are removing a test for `exe-cex` to `cex`. But why are you deleting the old test?

test/cex/cex_nonlinear_arith.c
```diff
@@ -13,8 +16,8 @@ extern int nd_int(void);
 
 int main(int argc, char**argv) {
   int x = nd_int();
-  __VERIFIER_assume (x >= 1);
-  __VERIFIER_assume (x <= 100);  
+  assume (x >= 1);
```
agurfinkel, line 19: `__VERIFIER_assume` was correct. `assume` is expanded into unknown external function because `seahorn/seahorn.h` is not included

test/cex/cex_nonlinear_arith.c
```diff
@@ -13,8 +16,8 @@ extern int nd_int(void);
...
+  assume (x <= 100);
```
agurfinkel, line 20: I think the problem here is with wrong assume. With opsem2 this example should be solvable.

lib/seahorn/CexExeGenerator.cc
```diff
@@ -37,122 +37,133 @@ bool extractArrayContents(Expr e, kv &out, Expr &defaultValue) {
...
-  Expr shadowMemRaw = m_trace.engine().getPtrAddressable(shadowMem);
-  Expr hPtrStartRaw = m_trace.engine().getPtrAddressable(hPtrStart);
-  if (!shadowMemRaw || !hPtrStartRaw || !size) {
-    LOG("cex", errs() << "Skipping harness for memhavoc due to lacking info");
-    return;
+
+  void visitCallSite(CallSite CS) {
+    Instruction &I = *CS.getInstruction();
+    const Value *CV = CS.getCalledValue();
+    const Function *CF = dyn_cast<Function>(CV->stripPointerCasts());
```
agurfinkel, line 118: we have getCalledFunction utility somewhere in our code

lib/seahorn/CexExeGenerator.cc
```diff
@@ -37,122 +37,133 @@ bool extractArrayContents(Expr e, kv &out, Expr &defaultValue) {
...
+    // shadow.mem.load(i32 x, i32 %sm_n, i8* null)
+    if (I.getPrevNonDebugInstruction() == nullptr)
+      return;
+    const Instruction *prevI = I.getPrevNonDebugInstruction();
+    if (const CallInst *prevCI = dyn_cast<CallInst>(prevI)) {
+      ImmutableCallSite prevCS(prevCI);
+      const Value *prevCV = prevCS.getCalledValue();
+      const Function *prevF = dyn_cast<Function>(prevCV->stripPointerCasts());
+      if (!(prevF && prevF->getName().equals("shadow.mem.load"))) {
+        LOG("cex", errs() << "Skipping harness for memhavoc"
```
agurfinkel, line 70: use `ERR << ` instead of `errs() << `

lib/seahorn/CexExeGenerator.cc
```diff
@@ -37,122 +37,133 @@ bool extractArrayContents(Expr e, kv &out, Expr &defaultValue) {
...
+      }
+      Expr shadowMemRaw = m_cex.trace().engine().getPtrAddressable(shadowMem);
+      Expr hPtrStartRaw = m_cex.trace().engine().getPtrAddressable(hPtrStart);
+      if (!shadowMemRaw || !hPtrStartRaw || !size) {
+        LOG("cex",
+            errs() << "Skipping harness for memhavoc due to lacking info \n");
+        return;
+      }
+      LOG("cex", errs() << "Producing harness for " << CF->getName() << "\n";);
+      m_cex.addValueToFunc(CF, shadowMemRaw);
```
agurfinkel, line 110: I do not understand `shadowMemRaw`. The argument to `shadow.mem.load` is a version of a memory region. it is not a pointer... I must be missing something.

agurfinkel (changes requested):
all looks good except I'm not sure about what you are doing with `shadowMemRaw`

py/sea/commands.py
danblitzhou: yes, not sure why `eharness`(name of executable) was part of arguments in the first place; before removal, given an executable `debug.exe`, `ExecHarness` would call `debug.exe debug.exe`

test/cex/cex_mem.c
danblitzhou: good point, lets keep old tests but disable them for now

agurfinkel:
I believe we are still waiting for next iteration, right @danblitzhou ?

test/cex/cex_nonlinear_arith.c
danblitzhou: `seahorn/seahorn.h` is included on line 13 though

lib/seahorn/CexExeGenerator.cc
danblitzhou: in sea-dsa we injected `shadow.mem.load` in front of every `memhavoc` to associate every havoc pointer with corresponding shadow mem region; the idea is to evaluate the memory region for memory content.

agurfinkel:
commit lint is failing @danblitzhou

test/cex/cex_nonlinear_arith.c
agurfinkel: I am looking at diff on github, so must have missed that line

lib/seahorn/CexExeGenerator.cc
```diff
@@ -221,7 +229,7 @@ void CexExeGenerator<Trace>::buildNonDetFunction(const Function *func,
     else
       Args.push_back(ConstantInt::get(Type::getInt32Ty(m_context), 0));
   } else {
-    errs() << "WARNING: Unknown type: " << *RT << "\n";
+    ERR << "WARNING: Unknown type: " << *RT << "\n";
```
agurfinkel, line 232: either `ERR` or `WARN` but then no `WARNING` since the prefix is added automatically

agurfinkel (changes requested):
Please clean up commit history. I am getting lost. There are commits that mix configs, test, C++, and python code, and then commits on top of these that are better. 

Still a few issues to address from previous review. 

I am still not clear that `shdow.mem.load` is processed correctly, although I do understand the intention. We should talk... (but clean commit history first pls)

lib/seahorn/CexExeGenerator.cc
```diff
@@ -37,122 +37,130 @@ bool extractArrayContents(Expr e, kv &out, Expr &defaultValue) {
...
+    Instruction &I = *CS.getInstruction();
+    auto *CF = getCalledFunction(CS);
+    // previous instruction should be
+    // shadow.mem.load(i32 x, i32 %sm_n, i8* null)
+    if (I.getPrevNonDebugInstruction() == nullptr)
+      return;
+    const Instruction *prevI = I.getPrevNonDebugInstruction();
+    if (const CallInst *prevCI = dyn_cast<CallInst>(prevI)) {
+      ImmutableCallSite prevCS(prevCI);
+      const Function *prevF = prevCS.getCalledFunction();
```
danblitzhou, line 66: skipping strip pointer cast should be fine here `shadow.mem.load` is prepended directly

.github/workflows/test-seahorn-docker.yml
```diff
@@ -39,5 +39,8 @@ jobs:
       - name: Run mcfuzz tests
         run: docker run -v $(pwd):/host seahorn_container /bin/bash -c "cd seahorn/share/seahorn && lit -a --time-tests --max-time=1200 test/mcfuzz"
       - name: Run smc tests
-        run: docker run -v $(pwd):/host seahorn_container /bin/bash -c "cd seahorn/share/seahorn && lit -a --time-tests --max-time=1200 test/smc" 
-
+        run: docker run -v $(pwd):/host seahorn_container /bin/bash -c "cd seahorn/share/seahorn && lit -a --time-tests --max-time=1200 test/smc"
```
agurfinkel, line 42: split this commit into two commits. one for test (as commit message says), and one for ci.

include/seahorn/CexExeGenerator.hh
```diff
@@ -59,12 +60,8 @@ template <class Trace> class CexExeGenerator {
...
   // map function calls to return value(s)
   ValueMap<const Function *, ExprVector> m_func_val_map;
-  // list of <start addr, size> of havoc pointers
-  std::vector<std::pair<Expr, Expr>> m_memhavoc_args;
-
-  /// \brief fills m_memhavoc_args and m_func_val_map
-  void storeMemHavoc(unsigned loc, const Function *func, ImmutableCallSite cs,
-                     ImmutableCallSite prevCS);
+  // list of <start addr, size> of havoc non det pointers
+  std::vector<std::pair<Expr, Expr>> m_nondet_ptrs;
```
agurfinkel, line 64: maybe `llvm::SmallVector`

include/seahorn/CexExeGenerator.hh
```diff
@@ -98,6 +95,20 @@ public:
...
 
   void saveCexModuleToFile(llvm::StringRef CexFile);
+
+  const DataLayout &getDataLayout() { return m_dl; }
+
+  const TargetLibraryInfo &getTargetLibraryInfo() { return m_tli; }
+
+  Trace &trace() { return m_trace; }
+
+  void addValueToFunc(const Function *func, Expr val) {
```
agurfinkel, line 105: comment on what this does

include/seahorn/CexExeGenerator.hh
```diff
@@ -98,6 +95,20 @@ public:
...
+    m_func_val_map[func].push_back(val);
+  }
+
+  void addNonDetPtr(Expr start, Expr size) {
+    m_nondet_ptrs.push_back(std::make_pair(start, size));
```
agurfinkel, line 110: use `emplace_back`

lib/seahorn/CexExeGenerator.cc
```diff
@@ -37,122 +37,130 @@ bool extractArrayContents(Expr e, kv &out, Expr &defaultValue) {
...
+  BmcTraceVisitor(CexExeGenerator<Trace> &cex) : m_cex(cex) {}
+
+  void setLoc(unsigned loc) { m_loc = loc; }
+
+  void visitMemhavoc(CallSite CS) {
+    Instruction &I = *CS.getInstruction();
+    auto *CF = getCalledFunction(CS);
+    // previous instruction should be
+    // shadow.mem.load(i32 x, i32 %sm_n, i8* null)
+    if (I.getPrevNonDebugInstruction() == nullptr)
```
agurfinkel, line 61: only call `getPrevNon...` function once

lib/seahorn/CexExeGenerator.cc
```diff
@@ -37,122 +37,130 @@ bool extractArrayContents(Expr e, kv &out, Expr &defaultValue) {
...
+      auto *shadowMemI = dyn_cast<Instruction>(prevCS.getArgOperand(1));
+      if (!shadowMemI)
+        return;
+      Expr shadowMem = m_cex.trace().eval(m_loc, *shadowMemI, true);
+      // get memhavoc size from second operand of memhavoc
+      const Value *sizeArg = CS.getArgOperand(1);
+      Expr size;
+      if (auto *sizeI = dyn_cast<Instruction>(sizeArg)) {
+        size = m_cex.trace().eval(m_loc, *sizeI, true);
+      } else if (auto *sizeConst = dyn_cast<ConstantInt>(sizeArg)) {
```
agurfinkel, line 82: seems like `trace().eval()` should know how to evaluation `ConstantInt`

lib/seahorn/CexExeGenerator.cc
```diff
@@ -37,122 +37,130 @@ bool extractArrayContents(Expr e, kv &out, Expr &defaultValue) {
...
+      if (!(prevF && prevF->getName().equals("shadow.mem.load"))) {
+        LOG("cex", ERR << "Skipping harness for memhavoc"
+                       << " because shadow.mem.load cannot be found\n");
+        return;
+      }
+      // get memhavoc content from corresponding shadow mem region
+      auto *shadowMemI = dyn_cast<Instruction>(prevCS.getArgOperand(1));
+      if (!shadowMemI)
+        return;
+      Expr shadowMem = m_cex.trace().eval(m_loc, *shadowMemI, true);
```
agurfinkel, line 76: I believe that `shadowMem` is a memory map (i.e., array or lambda), right?

lib/seahorn/CexExeGenerator.cc
```diff
@@ -37,122 +37,130 @@ bool extractArrayContents(Expr e, kv &out, Expr &defaultValue) {
...
+      const Value *hPtrAlloc = CS.getArgOperand(0)->stripPointerCasts();
+      Expr hPtrStart;
+      if (auto *hPtrAllocInst = dyn_cast<Instruction>(hPtrAlloc)) {
+        hPtrStart = m_cex.trace().eval(m_loc, *hPtrAllocInst, true);
+      } else {
+        LOG("cex",
+            ERR << "unhandled Value of memhavoc ptr: " << *hPtrAlloc << "\n");
+        return;
+      }
+      Expr shadowMemRaw = m_cex.trace().engine().getPtrAddressable(shadowMem);
```
agurfinkel, line 100: This is wrong. 

`getAddressable` takes a pointer as input and returns an address. But `shadowMem` is not a pointer, it is a memory map, so, probably, a structure of 3 maps. Not sure why this works...

There should be a different function to get `raw` part of the memory map.

lib/seahorn/CexExeGenerator.cc
agurfinkel: I think I finally understand what I think is wrong. See comments above about `shadowMem`

agurfinkel (changes requested):
see comments

lib/seahorn/CexExeGenerator.cc
danblitzhou: yes, which is why we can use `ExprMemMap` to extract memory content

lib/seahorn/CexExeGenerator.cc
danblitzhou: the `Expr` encoding of a memory map looks very similar to pointer encoding, so I just used `getPtrAddressable`

agurfinkel: they both happen to be structs with 3 elements. But this is only in some configurations. The fields themselves are very different. In our default configuration, a pointer is `{base, offset, size}` and `getAddressable` returns `base + offset`. Not sure how this works for memory map which is `{raw, mem_offset, nem_size}` where `raw`, `mem_offset`, and `mem_size` are all memory maps themselves.

agurfinkel: @priyasiddharth what is preferred way to get raw memory from the manager?

priyasiddharth: Assuming given an expression which is a `MemValTy` we want the raw part.
Currently there is no API for this.
Discussing with Xiang, best is to add a `ExprToRaw(e)` method to `OpSemMemManagerBase` and implement it only in the needed `MemManagers`

agurfinkel: Name should be something like `getRawMem` to be similar to `getAddressable` for ptr. It converts memory map to its raw part (i.e., main memory without shadows). The term `Expr` should not be in the name since `Expr` is just a low level data structure, not a concept at this level.

lib/seahorn/CexExeGenerator.cc
danblitzhou: I tried using `eval` on `sizeConst` but the evaluation returned an empty Expr; does the trace keep track of constant variables?

lib/seahorn/CexExeGenerator.cc
danblitzhou: added `getRawMem` for all memory managers in use, now `shadowMemRaw` is extracted with that instead; the function checks whether an `Expr` is in the right form of `MemValTy`, if so, return `getRaw` of the MemValTy. @priyasiddharth please take a look on the new mem manager functions.

lib/seahorn/CexExeGenerator.cc
agurfinkel: I was suggesting to move this code into eval so that it does support eval of constant ints

agurfinkel:
lost track. is this waiting for me or @danblitzhou ? Still see some comments that are not marked resolved. Maybe they are addressed and I'm the only one who can mark them resolved... let me know

agurfinkel (approved):
LGTM

agurfinkel:
I'm good. @danblitzhou @priyasiddharth good to merge?

danblitzhou:
Should be good, let's merge once @priyasiddharth sign off the changed in mem manager

lib/seahorn/BvOpSem2Context.hh
```diff
@@ -729,6 +731,14 @@ public:
   /// \brief given a properly encoded pointer Expr \p p , return the raw
   /// expression representing memory address only
   virtual Expr ptrToAddr(Expr p) = 0;
+
+  /// \brief given an Expression \p e , return true if \p e has expected
+  /// encoding of a MemValTy
+  virtual bool isMemVal(Expr e) = 0;
```
priyasiddharth, line 737: This is not needed in the interface (context) class since no clients of context use this. Please remove this method.

lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
```diff
@@ -109,6 +109,11 @@ template <class T> bool ExtraWideMemManager<T>::isPtrTyVal(Expr e) const {
   return strct::isStructVal(e) && e->arity() == g_num_slots;
 }
 
+template <class T> bool ExtraWideMemManager<T>::isMemVal(Expr e) const {
+  // Our base is a struct of three exprs
+  return strct::isStructVal(e) && e->arity() == 3;
```
priyasiddharth, line 114: `s/3/g_numslots + 1`

lib/seahorn/BvOpSem2FatMemMgr.cc
```diff
@@ -581,6 +582,11 @@ class FatMemManager : public MemManagerCore {
     // struct with raw ptr + fat slots
     return strct::isStructVal(e) && e->arity() <= (1 + g_maxFatSlots);
   }
+
+  bool isMemVal(Expr e) const {
+    // struct with 3 fields
+    return (strct::isStructVal(e) && e->arity() == 3);
```
priyasiddharth, line 588: s/3/g_maxFatSlots+1

lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
```diff
@@ -35,6 +35,11 @@ bool TrackingRawMemManager::isPtrTyVal(Expr e) const {
   return !e || !strct::isStructVal(e);
 }
 
+bool TrackingRawMemManager::isMemVal(Expr e) const {
+  // raw and metadata
+  return strct::isStructVal(e) && e->arity() == 2;
```
priyasiddharth, line 40: `s/2/g_num_slots+1`

lib/seahorn/BvOpSem2RawMemMgr.cc
```diff
@@ -817,5 +817,9 @@ bool RawMemManagerCore::isPtrTyVal(Expr e) {
   return (!e || !strct::isStructVal(e));
 }
 
+bool RawMemManagerCore::isMemVal(Expr e) {
+  return (!e || !strct::isStructVal(e));
```
priyasiddharth, line 821: Are we ok accepting a `NULL` value for e? If so, why?

I would assume it should be `e && !strct::isStructVal(e)`

Note when constructing a `MemValTy`, we allow `NULL` values since that is a mechanism to signal that the resulting `MemValTy` is bad (e.g. an operation failed). I don't see a use case where we take a bad `MemValTy` and then `getRaw()` on that.

lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
priyasiddharth: Do we expect `e` to ever be NULL?

lib/seahorn/BvOpSem2Context.hh
```diff
@@ -729,6 +731,14 @@ public:
...
+
+  /// \brief given a properly encoded memory map Expr \p p , return the base
+  /// expression representing raw memory only
+  virtual Expr getRawMem(Expr p) = 0;
```
priyasiddharth, line 741: if `p` is `NULL` will we crash or return NULL?

I think some Memory managers crash and others return `NULL` expression('RawMemManagerCore`). I would suggest making it consistent across all.

lib/seahorn/BvOpSem2WideMemMgr.cc
```diff
@@ -483,6 +483,11 @@ bool WideMemManager::isPtrTyVal(Expr e) const {
   return strct::isStructVal(e) && e->arity() == g_num_slots;
 }
 
+bool WideMemManager::isMemVal(Expr e) const {
+  // struct of raw and size
+  return strct::isStructVal(e) && e->arity() == 2;
```
priyasiddharth, line 488: `s\2\g_num_slots+1`

priyasiddharth (changes requested):
Looks good for Memory manager changes. Some minor comments to improve code behaviour and readability.

lib/seahorn/BvOpSem2RawMemMgr.cc
danblitzhou: good point, when I first wrote this I was just trying to follow the constructor of `MemValTy`; Let's make all `isMemVal` return false when `e` is NULL and in that case `getRawMem` will just return an empty `Expr` (NULL); should probably do the same for `isPtrTyVal`

lib/seahorn/BvOpSem2Context.hh
danblitzhou: it should return NULL, which will be properly handled in `CexExeGenerator`

lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
danblitzhou: In extrawide g_num_slots == 3, should I just use `g_num_slots`? or make a separate global variable like `g_num_mem_slots = 3`?

priyasiddharth: g_num_slots should be fine.

priyasiddharth (approved):

PR #386, opened 2021-03-08
Improve memcpy

agurfinkel:
Support memcpy of unaligned regions

### DIFF
--- a/lib/seahorn/BvOpSem2MemRepr.cc
+++ b/lib/seahorn/BvOpSem2MemRepr.cc
@@ -3,7 +3,7 @@
 
 namespace {
 template <typename T, typename... Rest>
-auto as_std_array(const T &t, const Rest &... rest) ->
+auto as_std_array(const T &t, const Rest &...rest) ->
     typename std::array<T, sizeof...(Rest) + 1> {
   return {t, rest...};
 }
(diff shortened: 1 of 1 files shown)


lib/seahorn/BvOpSem2MemRepr.cc
```diff
@@ -358,12 +370,50 @@ OpSemMemLambdaRepr::MemCpy(PtrTy dPtr, PtrTy sPtr, unsigned len,
...
+        Expr lastWordValSrc =
+            op::bind::fapp(srcMem.toExpr(), readPtrInSrc.toExpr());
+
+        // -- compute the last word by taking chunks of source and destination
+        // -- words. source word comes first
+        unsigned wordSzInBits = wordSzInBytes * 8;
+        unsigned remainderBits = remainderBytes * 8;
+        auto &alu = m_ctx.alu();
+        Expr srcChunk =
+            alu.Extract({lastWordValSrc, wordSzInBits}, 0, remainderBits - 1);
```
priyasiddharth, line 398: Please confirm that `end` is inclusive

i.e. if `remainderBits = 4` then bits 0, 1, 2, 3 will be extracted.

lib/seahorn/BvOpSem2MemRepr.cc
```diff
@@ -347,8 +347,20 @@ OpSemMemLambdaRepr::MemCpy(PtrTy dPtr, PtrTy sPtr, unsigned len,
...
+      unsigned lastByteToCopy;
+      unsigned remainderBytes;
+      if (m_memManager.isIgnoreAlignment()) {
+        // if alignment is ignored, we treat it as alignment of 1
+        lastByteToCopy = len - 1;
+        remainderBytes = 0;
+      } else {
+        unsigned wordsToCopy = (len / wordSzInBytes);
+        // -- -1 because ptrInRangeCheck is inclusive
+        lastByteToCopy = (wordsToCopy - 1) * wordSzInBytes;
```
priyasiddharth, line 359: This is really `lastAlignedBytePosToCopy` isn't it? Do you think that is a better name?

priyasiddharth (commented):
Minor questions for my understanding.

lib/seahorn/BvOpSem2MemRepr.cc
agurfinkel: yes. definitely inclusive. Forgot `-1` at first and z3 complained with type error.

lib/seahorn/BvOpSem2MemRepr.cc
agurfinkel: done. will commit soon.

agurfinkel:
@priyasiddharth : good to merge?
@danblitzhou : ? (this is related to what you doing, so would be great if you can understand this code)

PR #387, opened 2021-03-11
feat(opsem): support for filters in UfoOpSem

ScottWe:
UfoOpSem inherits from OperationalSemantics, which supports instruction filtering.
In UfoOpSem, the filter checks are disabled through function overrides.
This commit restores the filter check in UfoOpSem::isTracked.

Note that filter support breaks some assumptions made by OpSemVisitor.
Therefore, additional guards are added against m_sem::isTracked.

### DIFF
--- a/lib/seahorn/UfoOpSem.cc
+++ b/lib/seahorn/UfoOpSem.cc
@@ -219,6 +219,9 @@ struct OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemBase {
   }
 
   void visitCmpInst(CmpInst &I) {
+    if (!m_sem.isTracked(I))
+      return;
+
     Expr lhs = havoc(I);
 
     const Value &v0 = *I.getOperand(0);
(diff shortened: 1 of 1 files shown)


agurfinkel:
@ScottWe some tests still fail. Do they fail for you locally as well? They pass on my version without your new commits. (but only in release mode)

ScottWe:
@agurfinkel I can reproduce the failures locally.

The filter is only used in BmcPass, correct? So the failed tests should be unrelated to the filter. However, my commit also calls isTracked on new instructions. I suspect that isTracked doesn't work properly for one of those instructions.

I will take a closer look.

ScottWe:
The last line is `UfoOpSem::isTracked(v)` is `return v.getType()->isIntegerTy();`. This returns false on call instructions with a void return type (e.g., assumes, asserts, function that might fail).

To fix this, I've added a conditional that will track call instructions, after all other checks have been made.

agurfinkel:
why not make `isTracked` track all void functions by changing
`return v.getType()->isIntegerTy() || v.getType()->isVoidTy()` ?

ScottWe:
That makes sense. I can't think of any reason not to. Updated.

agurfinkel:
let's see that all tests pass

ScottWe:
@agurfinkel Are you waiting on any revisions from me?

agurfinkel:
no. was waiting for checks to pass and forgot to check

PR #388, opened 2021-03-18
Support for predicate synthesis from partial definitions in small step semantics.

ScottWe:
This PR allows Seahorn to synthesize predicates from partial definitions using small step semantics.

- A function `f` that is marked with a `partial` annotation is treated as a partial function definition. If `f` returns 1 on an input, the input is included in the base case. If `f` calls to an external function, then the call is treated as `assume(0);`.
- New intrinsics, `sea.synth.assume` and `sea.synth.assert` are added to provide specifications to a partial function `f`. These intrinsics are used in place of `verifier.assume` and `verifier.assert` when the argument is a partial function. 
- If `sea.synth.assume` appears in a program, then the predicate is assumed to be true.
- If `sea.synth.assert` appears in a program, then the predicate must be true. In terms of the CHC encoding, this means that the predicate appears on the rhs of a rule.

The test cases demonstrate use case for the feature.

### DIFF
--- a/include/seahorn/HornifyFunction.hh
+++ b/include/seahorn/HornifyFunction.hh
@@ -37,10 +37,15 @@ protected:
   EZ3 &m_zctx;
   ExprFactory &m_efac;
 
+  llvm::Function *m_synthAssertFn = nullptr;
+
   /// whether encoding is inter-procedural (i.e., with summaries)
   bool m_interproc;
 
   void extractFunctionInfo(const BasicBlock &BB);
+
+  llvm::SmallVector<llvm::Instruction *, 8> getPartialFnsToSynth(Function &F);
+  void expandEdgeFilter(llvm::Instruction &I);
 
 public:
   HornifyFunction(HornifyModule &parent, bool interproc = false)
@@ -55,6 +60,7 @@ public:
 };
 
 class SmallHornifyFunction : public HornifyFunction {
+  void mkBBSynthRules(const LiveSymbols &ls, Function &F, SymStore &store);
 
 public:
   SmallHornifyFunction(HornifyModule &parent, bool interproc = false)
--- a/include/seahorn/Transforms/Instrumentation/GeneratePartialFnPass.hh
+++ b/include/seahorn/Transforms/Instrumentation/GeneratePartialFnPass.hh
@@ -1,0 +1,22 @@
+#pragma once
+
+/**
+ * Locates external function calls annotated by "partial" and generates
+ * placeholder implementations.
+ */
+
+#include "llvm/IR/Function.h"
+#include "llvm/Pass.h"
+
+using namespace llvm;
+
+namespace seahorn {
+
+// Returns true if GeneratePartialFnPass has marked F as a stub of a partial
+// function.
+bool isInferable(const Function &F);
+
+// Returns true if F has the "partial" annotation.
+bool isPartialFn(const Function &F);
+
+} // namespace seahorn
--- a/lib/Transforms/Instrumentation/GeneratePartialFnPass.cc
+++ b/lib/Transforms/Instrumentation/GeneratePartialFnPass.cc
@@ -1,0 +1,161 @@
+#include "seahorn/Transforms/Instrumentation/GeneratePartialFnPass.hh"
+
+#include "seahorn/Analysis/SeaBuiltinsInfo.hh"
+#include "seahorn/InitializePasses.hh"
+#include "seahorn/Passes.hh"
+#include "seahorn/Support/SeaDebug.h"
+
+#include "llvm_seahorn/IR/LLVMContext.h"
+#include "llvm/IR/Constants.h"
+#include "llvm/IR/IRBuilder.h"
+
+#define PARTIAL_FN_ANNOTATION "partial"
+#define PARTIAL_FN_STUB_PREFIX "sea.synth.inferable."
+
+#define PASS_NAME "generate-partial-fn"
+#define DEBUG_TYPE PASS_NAME
+
+using namespace llvm;
+
+namespace {
+
+// Returns true if I is marked by the "partial" metadata.
+bool hasPartialAnnotation(const llvm::Instruction &I) {
+  if (auto *meta = I.getMetadata(llvm_seahorn::LLVMContext::MD_annotation)) {
+    auto *tuple = cast<MDTuple>(meta);
+    for (auto &N : tuple->operands()) {
+      if (cast<MDString>(N.get())->getString() == PARTIAL_FN_ANNOTATION) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+} // namespace
+
+namespace seahorn {
+
+// Delcared in GeneratePartialFnPass.h
+bool isInferable(const Function &F) {
+  return F.getName().startswith(PARTIAL_FN_STUB_PREFIX);
+}
+
+// Delcared in GeneratePartialFnPass.h
+bool isPartialFn(const Function &F) {
+  // Conservative check: Is the first instruction marked as partial?
+  auto &I = F.getEntryBlock().front();
+  return hasPartialAnnotation(I);
+}
+
+namespace {
+
+// Instruments partial functions. This consists of identifying partial functions
+// (Boolean functions annotated with partial), locating all stubs (external
+// functions called from within the partial function), and instrumenting all
+// stubs as required in VC generations.
+class GeneratePartialFnPass : public llvm::ModulePass {
+private:
+  Function *m_assumeFn = nullptr;
+
+public:
+  static char ID;
+
+  GeneratePartialFnPass() : ModulePass(ID) {}
+
+  virtual void getAnalysisUsage(llvm::AnalysisUsage &AU) const {
+    AU.addRequired<SeaBuiltinsInfoWrapperPass>();
+    AU.setPreservesAll();
+  }
+
+  bool runOnModule(Module &M) {
+    auto &SBI = getAnalysis<SeaBuiltinsInfoWrapperPass>().getSBI();
+    m_assumeFn = SBI.mkSeaBuiltinFn(SeaBuiltinsOp::ASSUME, M);
+
+    bool changed = false;
+    for (auto &F : M) {
+      if (!F.empty()) {
+        changed |= runOnFunction(SBI, F);
+      }
+    }
+    return changed;
+  }
+
+  bool runOnFunction(SeaBuiltinsInfo &SBI, Function &F) {
+    // Checks that this is a partial function, with a Boolean return type.
+    if (!isPartialFn(F))
+      return false;
+
+    if (!F.getReturnType()->isIntegerTy(1)) {
+      DOG(errs() << "Partial function without Bool return type: "
+                 << F.getName().str() << "\n";);
+      return false;
+    }
+
+    // Provides a placeholder implementation to each partial function stub.
+    bool changed = false;
+    for (auto &BB : F) {
+      IRBuilder<> pfnBuilder(&BB);
+
+      for (auto &I : BB) {
+        // Does I call a user defined function.
+        CallBase *CB = dyn_cast<CallBase>(&I);
+        if (!CB || SBI.isSeaBuiltin(*CB))
+          continue;
+
+        // Is the callee defined.
+        Function *CF = CB->getCalledFunction();
+        if (!CF || !CF->isDeclaration())
+          continue;
+
+        // Does the function return an integer?
+        if (!CF->getReturnType() || !CF->getReturnType()->isIntegerTy())
+          continue;
+
+        // Records changes.
+        changed = true;
+        DOG(errs() << "Found external call from partial function "
+                   << F.getName().str() << " to " << CF->getName().str()
+                   << ".\n");
+
+        // Renames the stub so that it can be identified in other passes.
+        CF->setName(PARTIAL_FN_STUB_PREFIX + CF->getName().str());
+
+        // The stub should have a body. The body will force the ModuleHornify
+        // pass to generate a call to the stub. Consequently, arguments passed
+        // to the stub will be live within the partial function that calls it.
+        BasicBlock *block = BasicBlock::Create(CF->getContext(), "entry", CF);
+        IRBuilder<> stubBuilder(block);
+        stubBuilder.CreateRet(ConstantInt::get(CF->getReturnType(), 0));
+
+        // Ensures that the stub is not optimized away by sequential LLVM
+        // passes.
+        CF->addFnAttr(Attribute::NoInline);
+        CF->addFnAttr(Attribute::OptimizeNone);
+
+        // Within the partial function, assume that the call to the stub never
+        // happens. This allows for the body to be synthesized, starting from
+        // the location of the call.
+        pfnBuilder.SetInsertPoint(&I);
+        pfnBuilder.CreateCall(m_assumeFn, pfnBuilder.getFalse());
+      }
+    }
+    return changed;
+  }
+
+  virtual StringRef getPassName() const { return PASS_NAME; }
+};
+
+char GeneratePartialFnPass::ID = 0;
+
+} // namespace
+
+llvm::Pass *createGeneratePartialFnPass() {
+  return new GeneratePartialFnPass();
+}
+
+} // namespace seahorn
+
+using namespace seahorn;
+INITIALIZE_PASS(GeneratePartialFnPass, PASS_NAME,
+                "Marks and instruments partial function stubs.", false, false)
--- a/lib/Transforms/Scalar/PromoteVerifierCalls.cc
+++ b/lib/Transforms/Scalar/PromoteVerifierCalls.cc
@@ -1,5 +1,7 @@
 #include "seahorn/Transforms/Scalar/PromoteVerifierCalls.hh"
+
 #include "seahorn/Analysis/SeaBuiltinsInfo.hh"
+#include "seahorn/Transforms/Instrumentation/GeneratePartialFnPass.hh"
 
 #include "llvm/IR/CallSite.h"
 #include "llvm/IR/InstIterator.h"
@@ -32,6 +34,35 @@ Value *coerceToBool(Value *arg, IRBuilder<> &builder, Instruction &I) {
     arg = builder.CreateICmpNE(arg, ConstantInt::get(arg->getType(), 0));
   return arg;
 }
+
+/// Determines if \p v represents a partial function call. If so, then the
+/// function call is returned.
+Value *extractPartialFnCall(Value *v) {
+  // The partial call may be cast to bool by comparison to 0.
+  if (auto cmp = dyn_cast<ICmpInst>(v)) {
+    auto *lhs = cmp->getOperand(0);
+    auto *rhs = cmp->getOperand(1);
+    v = isa<CallInst>(lhs) ? lhs : rhs;
+  }
+
+  // Strips zext if applicable.
+  if (auto *ze = dyn_cast<ZExtInst>(v))
+    v = ze->getOperand(0);
+
+  // The value must now be a CallInst.
+  auto CI = dyn_cast<CallInst>(v);
+  if (!CI)
+    return nullptr;
+
+  // Ensures that fn is a Boolean function with a "partial" annotation.
+  const Function &F = (*CI->getCalledFunction());
+  if (!F.getReturnType() || !F.getReturnType()->isIntegerTy(1))
+    return nullptr;
+  if (!seahorn::isPartialFn(F))
+    return nullptr;
+  return v;
+}
+
 } // namespace
 namespace seahorn {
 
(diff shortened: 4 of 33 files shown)


ScottWe:
@agurfinkel Should be good to review, provided tests pass.

include/seahorn/HornifyFunction.hh
```diff
@@ -37,11 +37,16 @@ protected:
...
   ExprFactory &m_efac;
 
+  llvm::Function *m_synthAssertFn = nullptr;
+
   /// whether encoding is inter-procedural (i.e., with summaries)
   bool m_interproc;
 
   void extractFunctionInfo(const BasicBlock &BB);
 
+  std::list<llvm::Value *> getPartialFnsToSynth(BasicBlock &BB);
```
agurfinkel, line 47: why use linked list here? Why not SmallVector or std::vector?

include/seahorn/Transforms/Instrumentation/GeneratePartialFnPass.hh
```diff
@@ -0,0 +1,49 @@
+
+#ifndef __GENERATE_PARTIAL_FN_PASS__HH_
+#define __GENERATE_PARTIAL_FN_PASS__HH_
```
agurfinkel, line 3: #pragma once

include/seahorn/Transforms/Instrumentation/GeneratePartialFnPass.hh
```diff
@@ -0,0 +1,49 @@
...
+  static char ID;
+
+  GeneratePartialFnPass();
+
+  virtual void getAnalysisUsage(llvm::AnalysisUsage &AU) const;
+
+  bool runOnModule(Module &M);
+  bool runOnFunction(SeaBuiltinsInfo &SBI, Function &F);
+
+  virtual StringRef getPassName() const { return "PromotePartial"; }
```
agurfinkel, line 38: unify name of class and name or module. One is Generate another is Promote

include/seahorn/Transforms/Instrumentation/GeneratePartialFnPass.hh
```diff
@@ -0,0 +1,49 @@
...
+#include <set>
+#include <string>
+
+using namespace llvm;
+
+namespace seahorn {
+
+class SeaBuiltinsInfo;
+
+class GeneratePartialFnPass : public llvm::ModulePass {
```
agurfinkel, line 22: this pass does not expose any public api. it does not need to be exposed in a header file

lib/Transforms/Instrumentation/GeneratePartialFnPass.cc
```diff
@@ -0,0 +1,118 @@
...
+using namespace llvm;
+
+namespace seahorn {
+
+const std::string GeneratePartialFnPass::PREFIX("sea.synth.inferable.");
+
+GeneratePartialFnPass::GeneratePartialFnPass() : ModulePass(ID) {}
+
+void GeneratePartialFnPass::getAnalysisUsage(AnalysisUsage &AU) const {
+  AU.addRequired<SeaBuiltinsInfoWrapperPass>();
```
agurfinkel, line 24: preserves all is missing

lib/Transforms/Instrumentation/GeneratePartialFnPass.cc
```diff
@@ -0,0 +1,118 @@
...
+      changed = true;
+      LOG("partial-fn", errs() << "Found partial function external call "
+                               << CF->getName().str() << " in "
+                               << F.getName().str() << ".\n");
+
+      // Assumes that the call never happens.
+      builder.SetInsertPoint(&I);
+      CallInst *ci = builder.CreateCall(m_assumeFn, builder.getFalse());
+
+      // Generates a placeholder body.
```
agurfinkel, line 75: do not follow what this code does. please expand comment.

use SmallVector instead of std::vector

lib/Transforms/Scalar/PromoteVerifierCalls.cc
```diff
@@ -32,6 +34,41 @@ Value *coerceToBool(Value *arg, IRBuilder<> &builder, Instruction &I) {
...
+    auto *lhs = cmp->getOperand(0);
+    auto *rhs = cmp->getOperand(1);
+    v = isa<CallInst>(lhs) ? lhs : rhs;
+  }
+
+  // The value must now be a call instruction.
+  if (!isa<CallInst>(v))
+    return nullptr;
+
+  // Annotation2Metadata marks all instructions of the partial function with the
```
agurfinkel, line 52: why not mark function declaration only?

lib/Transforms/Scalar/PromoteVerifierCalls.cc
```diff
@@ -32,6 +34,41 @@ Value *coerceToBool(Value *arg, IRBuilder<> &builder, Instruction &I) {
...
+  // "partial" annotation. Partial functions must have at least one instruction
+  // (i.e., a return) so this check is robust.
+  CallSite CS(v);
```
agurfinkel, line 55: callsite is deprecated and removed in llvm11. should be a way to use callbase instead

lib/Transforms/Scalar/PromoteVerifierCalls.cc
```diff
@@ -32,6 +34,41 @@ Value *coerceToBool(Value *arg, IRBuilder<> &builder, Instruction &I) {
...
+  const Function &F = (*CS.getCalledFunction());
+  for (auto &BB : F) {
+    for (auto &I : BB) {
+      if (seahorn::GeneratePartialFnPass::hasPartialAnnotation(I)) {
+        // Strips zext if there is one.
+        if (auto *ze = dyn_cast<ZExtInst>(v))
```
agurfinkel, line 61: seems like this should be done earlier. and checked that v is a call

lib/Transforms/Scalar/PromoteVerifierCalls.cc
```diff
@@ -171,28 +210,57 @@ bool PromoteVerifierCalls::runOnFunction(Function &F) {
...
+          nfn = m_synthAssumeFn;
+        else if (fn->getName().equals("__VERIFIER_assert"))
+          nfn = m_synthAssertFn;
+        else
+          assert(0);
+
+        // Generates code.
+        IRBuilder<> Builder(F.getContext());
+        Builder.SetInsertPoint(&I);
+        auto *sentinel = ConstantInt::get(partialFn->getType(), 1);
```
agurfinkel, line 229: in C 0 means false and non 0 means true. I think you are assuming that 1 means true instead

lib/seahorn/HornifyFunction.cc
```diff
@@ -37,6 +38,7 @@ void HornifyFunction::extractFunctionInfo(const BasicBlock &BB) {
     return;
 
   FunctionInfo &fi = m_sem.getFunctionInfo(F);
+  fi.isInferable = GeneratePartialFnPass::isInferable(F);
```
agurfinkel, line 41: isInferable should be declared outside of a class.

lib/seahorn/HornifyFunction.cc
```diff
@@ -137,13 +142,64 @@ void HornifyFunction::extractFunctionInfo(const BasicBlock &BB) {
...
+std::list<llvm::Value *> HornifyFunction::getPartialFnsToSynth(BasicBlock &BB) {
+  // Gets reference to sea.synth.assert.
+  if (!m_synthAssertFn) {
+    auto &SBI = m_parent.getSBI();
+    auto &M = (*BB.getParent()->getParent());
+    m_synthAssertFn = SBI.mkSeaBuiltinFn(SeaBuiltinsOp::SYNTH_ASSERT, M);
+  }
+
+  // Records all functions passed to sea.synth.assert within BB.
+  std::list<llvm::Value *> partials;
```
agurfinkel, line 167: why linked list. use SmallVector

lib/seahorn/HornifyFunction.cc
```diff
@@ -137,13 +142,64 @@ void HornifyFunction::extractFunctionInfo(const BasicBlock &BB) {
...
+    // Must be a call to sea.synth.assert.
+    if (!isa<CallInst>(&I))
+      continue;
+
+    // Analyzes the call.
+    CallSite CS(&I);
+    if (const Function *fn = CS.getCalledFunction()) {
+      if (fn->getName().equals(m_synthAssertFn->getName())) {
+        // The return value of the partial function is compared to 1. It is
+        // possible that the check is optimized away.
```
agurfinkel, line 178: would be easier to require that output of function to synthesize is i1

you can use bool in C which￼ is defined in stddef

agurfinkel (changes requested):
some comments

lib/seahorn/HornifyFunction.cc
```diff
@@ -45,53 +47,56 @@ void HornifyFunction::extractFunctionInfo(const BasicBlock &BB) {
...
-      if (cf && (cf->getName().equals("shadow.mem.in") ||
-                 cf->getName().equals("shadow.mem.out"))) {
-        const Value &v = *CS.getArgument(1);
-        Expr r = m_sem.symb(v);
-        if (!r)
-          continue;
-        fi.regions.push_back(&v);
-        sorts.push_back(bind::typeOf(r));
+  // memory regions (omitted for partial function stubs)
+  if (!fi.isInferable) {
```
agurfinkel, line 51: splitting everything on `isInferable` makes the code harder to follow. At least add big comment at the beginning. Alternatively, refactor the code. For example, use lambdas to capture big blocks of code, then call it only under appropriate condition.

lib/seahorn/HornifyFunction.cc
```diff
@@ -137,13 +142,64 @@ void HornifyFunction::extractFunctionInfo(const BasicBlock &BB) {
...
+  std::list<llvm::Value *> partials;
+  for (auto &I : BB) {
+    // Must be a call to sea.synth.assert.
+    if (!isa<CallInst>(&I))
+      continue;
+
+    // Analyzes the call.
+    CallSite CS(&I);
+    if (const Function *fn = CS.getCalledFunction()) {
+      if (fn->getName().equals(m_synthAssertFn->getName())) {
```
agurfinkel, line 176: instead of iterating over all instructions, iterate over all uses of `m_synthAssertFn`

lib/seahorn/HornifyFunction.cc
```diff
@@ -137,13 +142,64 @@ void HornifyFunction::extractFunctionInfo(const BasicBlock &BB) {
...
+    }
+  }
+
+  return partials;
+}
+
+void HornifyFunction::expandEdgeFilter(llvm::Instruction &I) {
+  m_sem.addToFilter(I);
+  if (!isa<PHINode>(&I)) {
+    // If the block is recursive, then the PHI node can reference instructions
```
agurfinkel, line 194: the comment makes it sound like `I` is expected to be `PHINode` but the condition checks for the opposite

agurfinkel (changes requested):
more comments

lib/seahorn/HornifyFunction.cc
```diff
@@ -175,6 +231,14 @@ void SmallHornifyFunction::runOnFunction(Function &F) {
       extractFunctionInfo(BB);
   }
 
+  // If F is an partial function stub, it should not have a body.
+  const FunctionInfo &fi = m_sem.getFunctionInfo(F);
+  if (fi.isInferable) {
+    LOG("seahorn",
+        errs() << "Adding partial fn stub: " << F.getName().str() << "\n");
```
agurfinkel, line 238: the print message says that something will be added, but the code returns without adding anything

lib/seahorn/HornifyFunction.cc
```diff
@@ -227,6 +291,62 @@ void SmallHornifyFunction::runOnFunction(Function &F) {
   side.clear();
   s.reset();
 
+  // Generates rules for synthesis.
+  for (auto &BB : F) {
+    auto partials = getPartialFnsToSynth(BB);
```
agurfinkel, line 296: potential performance penalty when synthesis is not enabled

lib/seahorn/HornifyFunction.cc
```diff
@@ -227,6 +291,62 @@ void SmallHornifyFunction::runOnFunction(Function &F) {
...
+    if (partials.empty())
+      continue;
+
+    // Generates a rule for each synthesis assert.
```
agurfinkel, line 300: factor out into a separate function for readability

lib/seahorn/HornifyFunction.cc
```diff
@@ -227,6 +291,62 @@ void SmallHornifyFunction::runOnFunction(Function &F) {
...
+        Expr post = side.back();
+        side.pop_back();
+
+        // Assume a-priori that the partial function call returns 1. It is
+        // possible that the return value of I has been optimized to Int1Ty.
+        auto rv = m_sem.lookup(s, I);
+        if (I.getType()->isIntegerTy(1)) {
+          side.push_back(mk<EQ>(rv, mk<TRUE>(m_efac)));
+        } else {
+          side.push_back(mk<EQ>(rv, mkTerm<mpz_class>(1, m_efac)));
```
agurfinkel, line 325: enforce synthesized functions to be `i1` instead of handling other choices here

lib/seahorn/HornifyFunction.cc
```diff
@@ -227,6 +291,62 @@ void SmallHornifyFunction::runOnFunction(Function &F) {
...
+        Expr tau = mknary<AND>(mk<TRUE>(m_efac), side);
+
+        expr::filter(tau, bind::IsConst(),
+                     std::inserter(allVars, allVars.begin()));
+        expr::filter(post, bind::IsConst(),
+                     std::inserter(allVars, allVars.begin()));
+
+        LOG("seahorn", errs()
+                           << "Adding synthesis rule : "
+                           << *mk<IMPL>(boolop::land(pre, tau), post) << "\n";);
```
agurfinkel, line 336: factor out the rule that is being added and print what exactly is being added

test/synth-sm/04_loop_sat.c
```diff
@@ -0,0 +1,46 @@
...
+extern int nd1();
+extern int nd2();
+extern int nd3();
+extern int nd4();
+
+// Loop invariant.
+extern int infer(int x, int n);
+int __attribute__((annotate("partial")))
+    __attribute__((noinline))
+    inv(int x, int n)
```
agurfinkel, line 16: are synthesized functions restricted to scalar arguments? Can a pointer argument be passed? What would that mean? Do we need this for our use case?

If we allow for pointer arguments, we need to pass implicit memory regions (arrays) that these pointers index into. This is not hard conceptually, but might be harder due to interplay with alias analysis. 

Synthesized functions are read-only, this might simplify things...

test/synth-sm/06_object_sat.c
```diff
@@ -0,0 +1,71 @@
...
+extern int nd4();
+extern int nd5();
+extern int nd6();
+extern int nd7();
+extern int nd8();
+extern int nd9();
+
+// Object invariant.
+extern int infer(int count, int max);
+int __attribute__((annotate("partial")))
```
agurfinkel, line 19: factor out all attributes into `#define` . It will make it easier to read. Something like
```
#define PARTIAL_FN __attribute__ ...
int PARTIAL_FN inv(int count, int max);
```

agurfinkel (changes requested):
see comments

agurfinkel:
@ScottWe done with review. let me know if you have questions.

test/synth-sm/06_object_sat.c
ScottWe: Good idea. Should this be added to `seahorn.h`? Or just added to the test cases for now?

test/synth-sm/04_loop_sat.c
ScottWe: I was working under the assumption that the synthesized functions were read-only. We do not need to pass pointers for our use-case.

Is `read-only` an attribute in llvm? I recall seeing that in the source code, but I can't find documentation for it.

test/synth-sm/06_object_sat.c
agurfinkel: you can add it to seahorn.h for convenience

test/synth-sm/04_loop_sat.c
agurfinkel: you can mark function in LLVM as only reading memory. This affects optimizations. I was wondering less about marking and optimizations, and more about the behavior of the function. If there are no pointers for now, we don't need to get into this discussion.

lib/seahorn/HornifyFunction.cc
ScottWe: I've made some changes. Is this what you had in mind with respect to lambdas?

agurfinkel: Yes, but the result is not any more readable than before. Maybe if all the lambda functions are moved up.

ScottWe: I've tried rearranging a bit.

lib/seahorn/HornifyFunction.cc
ScottWe: Updated.

lib/Transforms/Scalar/PromoteVerifierCalls.cc
ScottWe: This is how `Annotation2Metadata` is implemented. It reads the from `llvm.global.annotations`, extracts the function annotations, and then annotates each instruction within the function.

agurfinkel: you are right. I did not see this at first. I was expecting the annotation to be attached to the function only. Oh well

lib/Transforms/Instrumentation/GeneratePartialFnPass.cc
ScottWe: Revised

lib/seahorn/HornifyFunction.cc
ScottWe: Updated. The code only looks at blocks that call `sea.synth.assert`. The check is constant in the size of the module, and linear in the number of calls to `sea.synth.assert`.

lib/seahorn/HornifyFunction.cc
```diff
@@ -45,54 +113,12 @@ void HornifyFunction::extractFunctionInfo(const BasicBlock &BB) {
...
-      continue;
-
-    const Value *val = getTerm<const Value *>(u);
-    if (!m_sem.isTracked(*val))
-      continue;
-    if (const GlobalVariable *gv = dyn_cast<const GlobalVariable>(val)) {
-      fi.globals.push_back(gv);
-      sorts.push_back(bind::typeOf(v));
-    }
+  if (fi.isInferable) {
```
agurfinkel, line 116: I think this is clearer. Thanks!

ScottWe:
All feedback should be addressed now.

lib/Transforms/Instrumentation/GeneratePartialFnPass.cc
```diff
@@ -0,0 +1,162 @@
...
+bool isInferable(const Function &F) {
+  const std::string prefix(PARTIAL_FN_STUB_PREFIX);
+  return (F.getName().str().compare(0, prefix.size(), prefix) == 0);
+}
+
+// Delcared in GeneratePartialFnPass.h
+bool isPartialFn(const Function &F) {
+  for (auto &BB : F) {
+    for (auto &I : BB) {
+      return hasPartialAnnotation(I);
```
agurfinkel, line 46: shouldn't this be more conservative. If a function is inlined, its instructions will all have annotations, and the annotations are likely preserved during inlining.

We could check that, for example, the entry basic block has annotations, or maybe that all instructions in the entry basic block have annotations.

lib/Transforms/Instrumentation/GeneratePartialFnPass.cc
```diff
@@ -0,0 +1,162 @@
...
+          continue;
+
+        // Records changes.
+        changed = true;
+        LOG("partial-fn", errs() << "Found external call from partial function "
+                                 << F.getName().str() << " to "
+                                 << CF->getName().str() << ".\n");
+
+        // Renames the stub so that it can be identified in other passes.
+        CF->setName(std::string(PARTIAL_FN_STUB_PREFIX) + CF->getName().str());
```
agurfinkel, line 123: should be albe to use llvm::StringRef or llvm::Twine instead of std::string

lib/Transforms/Instrumentation/GeneratePartialFnPass.cc
```diff
@@ -0,0 +1,162 @@
...
+} // namespace
+
+llvm::Pass *createGeneratePartialFnPass() {
+  return new GeneratePartialFnPass();
+}
+
+} // namespace seahorn
+
+using namespace seahorn;
+INITIALIZE_PASS(GeneratePartialFnPass, "partial-fn-generator",
```
agurfinkel, line 161: short pass name should match string used in `LOG`. better use `DOG` macro

agurfinkel (changes requested):
almost there

lib/Transforms/Instrumentation/GeneratePartialFnPass.cc
ScottWe: What is the `DOG` macro?

agurfinkel: Debug LOG. grep code for its definition

ScottWe:
Changes made.

agurfinkel (approved):

PR #389, opened 2021-03-19
feat(fat): add vacuity check under fat bound checks

LinerSu:
Add vacuity check under fat bound checks. Concretely, the `verifierAssert` call inserts when there exists a comparison for bound checks. To invoke such checks, add flag `--enable-fat-vacuity-check` under `sea` binary or add `horn-bnd-chk-vac` under `seapp` binary.

### DIFF
--- a/py/sea/commands.py
+++ b/py/sea/commands.py
@@ -899,6 +899,8 @@ class FatBoundsCheck(sea.LimitedCmd):
                         metavar='STR', help='Log level')
         ap.add_argument('--no-fat-fns', dest='no_fat_fns', default=None,
                         type=str, metavar='STR,STR,...', help='List of functions to NOT instrument')
+        add_bool_argument(ap, 'enable-fat-vacuity-check', dest='enable_fat_vacuity_check', default=False,
+                         help='Enable vacuity check for fat bound checks')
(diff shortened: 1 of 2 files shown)


py/sea/commands.py
```diff
@@ -899,6 +899,8 @@ def mk_arg_parser (self, ap):
                         metavar='STR', help='Log level')
         ap.add_argument('--no-fat-fns', dest='no_fat_fns', default=None,
                         type=str, metavar='STR,STR,...', help='List of functions to NOT instrument')
+        ap.add_argument ('--enable-fat-vacuity-check', dest='enable_fat_vacuity_check', default=False,
```
agurfinkel, line 902: we have special handling of boolean options. use that.

agurfinkel (approved):
one minor issue. mostly looks good

py/sea/commands.py
LinerSu: Do you mean we add options into `sea.yaml`?

agurfinkel: no. how bool options are added to args. there is a special function for that

LinerSu: Made it. Used `add_bool_argument`.

PR #390, opened 2021-03-23
test(dsa): add two tests for checking global initializations

LinerSu:
Issue: Assigned function pointers on global structure are ignored by `sea-dsa`.


caballa:
I've moved these tests to sea-dsa where I think they should be. 
So I'll close this PR without merging it. If you or @agurfinkel  want the tests in SeaHorn feel free to reopen the PR.

PR #392, opened 2021-03-26
Extends support of partial function synthesis to large step semantics.

ScottWe:
Extends the changes in PR #388 to large step semantics. Some new helper methods are introduced.


agurfinkel:
@ScottWe is this ready for review

ScottWe:
@agurfinkel All changes should be on this branch now.

lib/seahorn/HornifyFunction.cc
```diff
@@ -448,6 +448,26 @@ void SmallHornifyFunction::runOnFunction(Function &F) {
...
 }
 
+namespace {
+
+/// \brief Returns all successor edges of \p cp that contain the basic block
+/// \c target between the source and destination (inclusively).
+llvm::SmallVector<const CpEdge *, 8>
+filterCpEdgesByBB(const CutPoint &cp, llvm::BasicBlock &target) {
+  llvm::SmallVector<const CpEdge *, 8> matchingEdges;
+  for (auto *e : boost::make_iterator_range(cp.succ_begin(), cp.succ_end())) {
```
agurfinkel, line 458: `llvm::make_range`

lib/seahorn/HornifyFunction.cc
```diff
@@ -468,6 +468,88 @@ filterCpEdgesByBB(const CutPoint &cp, llvm::BasicBlock &target) {
...
+      break;
+    for (auto &I : BB)
+      expandEdgeFilter(I);
+  }
+
+  // Uses iterators rather than `for (auto I : target)` since an iterator is
+  // required by execRange.
+  const BasicBlock &source = edge.source().bb();
+  const ExprVector &live = ls.live(&source);
+  for (auto itr = target.begin(); itr != target.end(); ++itr) {
```
agurfinkel, line 487:
```
for (auto itr = target.begin(), end = target.end(); itr != end; ++itr ) ...
```

lib/seahorn/HornifyFunction.cc
```diff
@@ -732,6 +739,23 @@ void LargeHornifyFunction::runOnFunction(Function &F) {
...
+  s.reset();
+  for (auto partial : getPartialFnsToSynth(F)) {
+    for (auto &cp : cpg) {
+      auto &srcBB = cp.bb();
+      if (reached.count(&srcBB) <= 0)
+        continue;
+
+      auto &targetBB = (*partial->getParent());
+      for (auto *edge : filterCpEdgesByBB(cp, targetBB)) {
+        assert(mkEdgeSynthRules(ls, *partial, *edge, targetBB, vcgen, s));
```
agurfinkel, line 752: `assert` is not compiled in  Release mode. Is the intention that `mkEdgeSynthRules` is used during `Debug` mode only?

agurfinkel (approved):
LGTM

minor stuff

lib/seahorn/HornifyFunction.cc
ScottWe: This and other issues are updated.

lib/seahorn/VCGen.cc
```diff
@@ -146,6 +146,10 @@ void VCGen::genVcForCpEdge(OpSemContext &ctx, const CpEdge &edge) {
       checkSideAtBb(head, ctx.side(), bbV, *smt, edge, bb);
   }
 
+  // -- if the block is truncated, omit phi nodes
+  if (&*edge.rbegin() == &target.bb())
```
agurfinkel, line 150: why not add a boolean flag for this behaviour?

I've looked at this code several times. I'm sure it works for the examples you have. But I can't really understand the condition. Why not add `includePhi` boolean argument that is set to true by default, and set it to false when necessary?

agurfinkel (approved):
Still stuck on one little thing...

lib/seahorn/VCGen.cc
ScottWe: I see your point. My solution is unnecessarily complex.

How the code works: There is a structural difference between CPG edges, and the "truncated edges" I generate. The condition identifies "truncated edges".

I think that a flag would be better for the following reasons.
1. A flag is a more robust solution. It will work even if the implementation of CPG changes.
2. If I use a flag, then the intent of the code is more explicit.
3. If I use a flag, then the feature can be reused by code without "truncated edges".
I've updated `genVcForCpEdge` accordingly.

agurfinkel (approved):
LGTM

PR #393, opened 2021-03-28
fix(scalar): avoids convertion of assert calls if conseq is true const

LinerSu:
If the assert instruction is `verifier.assert(i1 true)`, skips the transformation into `sea.assert.if`.

### DIFF
--- a/lib/Transforms/Scalar/UnifyAssumes.cc
+++ b/lib/Transforms/Scalar/UnifyAssumes.cc
@@ -138,9 +138,14 @@ bool UnifyAssumesPass::runOnFunction(Function &F) {
   for (auto ci : assumes)
     ci->eraseFromParent();
 
+  // -- process all asserts
   for (auto ci : asserts) {
     processAssertInst(*ci, *assumeFlag);
   }
+
+  // -- cleanup the asserts vector
+  asserts.clear();
+
   // insert call to assume before the last instructions of the exit block
   // (maybe before seahorn.fail)
   CallInst *seaFailCall = findSeahornFail(F);
@@ -172,10 +177,17 @@ void UnifyAssumesPass::processAssertInst(CallInst &CI, AllocaInst &flag) {
   llvm::CallSite CS(&CI);
   IRBuilder<> B(bb);
   B.SetInsertPoint(&CI);
+  Value *conseq = CS.getArgument(0);
+  // remove instruction if verifier.assert(true)
+  if (auto *conseq_const = dyn_cast<llvm::ConstantInt>(conseq)) {
+    if (!conseq_const->isZero()) {
+      CI.eraseFromParent();
+      return;
+    }
+  }
   auto ante = B.CreateLoad(&flag);
   // negate condition if verifier.assert.not seen
   bool isNot = m_SBI->getSeaBuiltinOp(CI) == seahorn::SeaBuiltinsOp::ASSERT_NOT;
-  Value *conseq = CS.getArgument(0);
   if (isNot) {
     conseq = B.CreateNot(conseq);
   }


lib/Transforms/Scalar/UnifyAssumes.cc
```diff
@@ -172,10 +172,15 @@ void UnifyAssumesPass::processAssertInst(CallInst &CI, AllocaInst &flag) {
   llvm::CallSite CS(&CI);
   IRBuilder<> B(bb);
   B.SetInsertPoint(&CI);
+  Value *conseq = CS.getArgument(0);
+  // skip transformation if verifier.assert(true)
+  if (auto *conseq_const = dyn_cast<llvm::Constant>(conseq)) {
+    if (conseq_const->isOneValue())
```
agurfinkel, line 178: in C 0 is false and non zero is true

you are assuming that 1 is true 

while this does not matter because the argument is i1 it is better to do it properly. I think there is also a method to check that a constant is false

agurfinkel (commented):
if this works then it is ok. 

but removing assert all together would be better

lib/Transforms/Scalar/UnifyAssumes.cc
LinerSu: Will update by using `isZeroValue`.

LinerSu:
I add code to remove instruction if assert(true). Let me know if this is not your expectation.

lib/Transforms/Scalar/UnifyAssumes.cc
```diff
@@ -172,10 +172,17 @@ void UnifyAssumesPass::processAssertInst(CallInst &CI, AllocaInst &flag) {
   llvm::CallSite CS(&CI);
   IRBuilder<> B(bb);
   B.SetInsertPoint(&CI);
+  Value *conseq = CS.getArgument(0);
+  // remove instruction if verifier.assert(true)
+  if (auto *conseq_const = dyn_cast<llvm::Constant>(conseq)) {
+    if (!conseq_const->isZeroValue()) {
+      CI.eraseFromParent();
```
agurfinkel, line 179: are you sure this does not invalidate any iterator?

LinerSu: Are we running on multithreaded mode? I think the iterator only affected is here:
https://github.com/seahorn/seahorn/blob/dev10/lib/Transforms/Scalar/UnifyAssumes.cc#L141-L143
Is that ok I created a new vector to store `assert(true)` instruction and delete them at once after invoking `processAssertInst`?

agurfinkel: Oh, all is good. We are traversive the `asserts` and deleting them not form the list we are traversing.
Just empty the `asserts` vector after the for loop since it might contain deallocated memory. We should not traverse the asserts more than once.

LinerSu: Added cleanup code after used for `assert` vector on new commit.

agurfinkel:
squash commits and we are good to go

LinerSu:
DONE!! I did by forced push.

lib/Transforms/Scalar/UnifyAssumes.cc
```diff
@@ -172,10 +177,17 @@ void UnifyAssumesPass::processAssertInst(CallInst &CI, AllocaInst &flag) {
   llvm::CallSite CS(&CI);
   IRBuilder<> B(bb);
   B.SetInsertPoint(&CI);
+  Value *conseq = CS.getArgument(0);
+  // remove instruction if verifier.assert(true)
+  if (auto *conseq_const = dyn_cast<llvm::Constant>(conseq)) {
+    if (!conseq_const->isZeroValue()) {
```
agurfinkel, line 183: apparently `isZeroValue` is for floating point numbers. 
We should use `isZero` of `ConstantInt`.

Sorry, should have spotted this earlier, but only looked at LLVM code now before merging.

LinerSu: Ok. I made this change.

agurfinkel (approved):
LGTM

PR #395, opened 2021-04-01
feat(seahorn): small step synthesis does not rely on opsem order

ScottWe:
Small step synthesis relies on the order of instructions in `side`. However, opsem does not place guarantees on the order of terms in `side`. To fix this, I have:
1. exposed a new interface (`execRange`) through `OperationalSemantics` to assemble individual instructions,
2. used `execRange` within synthesis to assemble the rhs of each synthesis rule.

Note: clang-format required that I fix `LegacyOperationalSemantics.hh` and `OperationalSemantics.hh`.

### DIFF
--- a/lib/seahorn/HornifyFunction.cc
+++ b/lib/seahorn/HornifyFunction.cc
@@ -243,8 +243,8 @@ void SmallHornifyFunction::mkBBSynthRules(const LiveSymbols &ls, Function &F,
 
     // Searches for a matching instruction.
     bool found = false;
-    for (auto &I : BB) {
-      expandEdgeFilter(I);
+    for (auto itr = BB.begin(); itr != BB.end(); ++itr) {
+      auto &I = (*itr);
 
       found = (partial == &I);
       if (found) {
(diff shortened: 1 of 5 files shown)


lib/seahorn/HornifyFunction.cc
```diff
@@ -243,8 +243,8 @@ void SmallHornifyFunction::mkBBSynthRules(const LiveSymbols &ls, Function &F,
 
     // Searches for a matching instruction.
     bool found = false;
-    for (auto &I : BB) {
-      expandEdgeFilter(I);
+    for (auto itr = BB.begin(); itr != BB.end(); ++itr) {
```
agurfinkel, line 246: why change to this form?

lib/seahorn/UfoOpSem.cc
```diff
@@ -1206,6 +1206,17 @@ void UfoOpSem::execBr(SymStore &s, const BasicBlock &src, const BasicBlock &dst,
   }
 }
 
+void UfoOpSem::execRange(SymStore &s, const llvm::BasicBlock::iterator begin,
+                         const llvm::BasicBlock::iterator end, ExprVector &side,
+                         Expr act) {
+  OpSemVisitor v(s, *this, side);
+  v.setActiveLit(act);
+  for (const auto &I : boost::make_iterator_range(begin, end)) {
```
agurfinkel, line 1214: use `llvm::make_range` instead of `boost`

agurfinkel (approved):
LGTM

lib/seahorn/UfoOpSem.cc
ScottWe: Fixed

ScottWe: Fixed.

PR #399, opened 2021-04-24
widemem: initialize size of null ptrs to zero

priyasiddharth:


lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
```diff
@@ -260,8 +260,8 @@ ExtraWideMemManager<T>::storeValueToMem(Expr _val,
       m_sem.getTD().getTypeStoreSize(const_cast<llvm::Type *>(&ty));
   ExprFactory &efac = base.v()->efac();
   // init memval to a default value
-  MemValTy res(m_main.zeroedMemory(),
-               m_ctx.alu().ui(g_uninit_small, wordSizeInBits()), m_uninit_size);
+  MemValTy res(m_main.zeroedMemory(), m_offset.zeroedMemory(),
```
agurfinkel, line 263: why? this will treat a skipped store as zero of all memory. We usually treated this as nondet instead by returning just zero constructed res

priyasiddharth: What is a zero constructed res?

If you mean `res(expr())`, that does not work with the current ctors of `MemValTy`

agurfinkel: Yes. In Raw where Res is Expr we simply let it be Expr() initially

We can update MemValTy to return Expr() if say first component is Expr()

priyasiddharth: `strct` operations can't work with `Expr()`. I now pass `MemValTy(Expr())` and do the right thing in `MemValTy`

agurfinkel (approved):
LGTM

PR #400, opened 2021-04-30
Hybrid memory representation

danblitzhou:
- added `OpSemMemHybridRepr`: store: array theory, load: ITE
- iterative rewrite function for `store` Expr during `loadAlignedWordFromMem` for hybrid repr


lib/seahorn/BvOpSem2MemRepr.cc
```diff
@@ -14,11 +15,93 @@ auto as_std_array(const T &t, const Rest &... rest) ->
...
+  Expr m_ptr;
+  OpSemAlu &m_alu;
+  unsigned m_ptrSz;
+  ArrayStoreRewriter(Expr ptr, OpSemAlu &alu, unsigned ptrSz)
+      : m_ptr(ptr), m_alu(alu), m_ptrSz(ptrSz) {}
+
+  Expr doPtrEq(Expr p1, Expr p2) { return m_alu.doEq(p1, p2, m_ptrSz); }
+
+  Expr operator()(Expr e) {
+    if (isOp<STORE>(e)) {
```
priyasiddharth, line 34: will adding an `assert` that we expect a struct with three fields help with debugging?

lib/seahorn/BvOpSem2MemRepr.cc
```diff
@@ -183,6 +265,24 @@ OpSemMemRepr::MemValTy OpSemMemArrayRepr::MemFill(PtrTy dPtr, char *sPtr,
   return MemValTy(res);
 }
 
+Expr OpSemMemHybridRepr::loadAlignedWordFromMem(PtrTy ptr, MemValTy mem) {
+  LOG("opsem-hybrid", INFO << "Load ptr " << *ptr.toExpr() << "\n");
+  LOG("opsem-hybrid", INFO << "From ptr " << *mem.toExpr() << "\n");
+
+  /** rewrite store into ITE **/
+  ArrayStoreRewriter rw(ptr.toExpr(), m_ctx.alu(),
```
priyasiddharth, line 273: IIUC when we `load` we lazily rewrite the `store` instructions in the memory array rather than rewriting everytime we store... is this correct?

priyasiddharth (commented):
Left some comments mainly for my understanding...

lib/seahorn/BvOpSem2MemRepr.cc
agurfinkel: yes. optimizing while doing so. next layer is to optimize store instructions to simplify load optimizations later on

include/seahorn/Expr/ExprRewriter.hh
```diff
@@ -0,0 +1,144 @@
...
+#include "seahorn/Expr/ExprCore.hh"
+#include "seahorn/Expr/ExprRewriteRule.hh"
+
+namespace expr {
+using namespace seahorn;
+
+Expr rewriteITEComp(Expr exp);
+
+struct RewriteFrame {
+  Expr m_exp;
```
agurfinkel, line 12: comment the role of the fields

include/seahorn/Expr/ExprRewriter.hh
```diff
@@ -0,0 +1,144 @@
...
+};
+
+using ExprVisitedMap = std::unordered_map<ENode *, bool>;
+using RewriteFrameVector = std::vector<RewriteFrame>;
+
+class ExprRewriterConfig {
+  /* apply rewrite rules */
+protected:
+  ExprFactory &m_efac; // for making expr
+  EZ3 &m_zctx;         // for z3 simplifier if needed
```
agurfinkel, line 26: I don't see why generic rewriter should depend on Z3. Its specialization can, but the generic rewriter should not

include/seahorn/Expr/ExprRewriter.hh
```diff
@@ -0,0 +1,144 @@
...
+  rewrite_result applyRewriteRules(Expr exp);
+
+  bool shouldRewrite(Expr exp);
+};
+
+template <typename Config> class ExprRewriter {
+protected:
+  Config m_config;
+  ExprFactory &m_efac; // for making expr
+  EZ3 &m_zctx;         // for z3 simplifier if needed
```
agurfinkel, line 57: should not be part of generic rewriter

include/seahorn/Expr/ExprRewriter.hh
```diff
@@ -0,0 +1,144 @@
...
+  ExprVisitedMap m_visited;
+
+  /* visit e, return true if any of the following is true:
+    1. e has been cached, or
+    2. current depth is 0 or
+    3. current m_config is dictates e should not be further visited
+    in any of the above case, push e to top of result stack
+    otherwise return false and push e into top of rewriteStack
+  */
+  bool visited(Expr e, size_t depth) {
```
agurfinkel, line 71: the function has too many side-effects. The name does not reflect them all, neither does the comment.

agurfinkel: perhaps renaming the function into `visit` would go long way to address the comment above

include/seahorn/Expr/ExprRewriter.hh
```diff
@@ -0,0 +1,144 @@
...
+        return;
+    }
+    m_rewriteStack.pop_back();
+    // all kids of exp has been visited, collect rewritten kids to form
+    // new expression
+    SmallVector<Expr, 4> new_kids;
+    size_t end = m_resultStack.size();
+    size_t begin = end - arity;
+    for (size_t i = begin; i < end; i++) {
+      new_kids.push_back(m_resultStack[i]);
```
agurfinkel, line 115: optimization: check that at least one `new_kid` is different from current arg. If not, the expression does not need to be re-created

include/seahorn/Expr/ExprRewriter.hh
```diff
@@ -0,0 +1,144 @@
...
+    }
+    Expr new_exp =
+        exp->getFactory().mkNary(exp->op(), new_kids.begin(), new_kids.end());
+    m_resultStack.resize(begin);
+    // apply rewrite rules to expression with new kids
+    rewrite_result rwRes = m_config.applyRewriteRules(new_exp);
+    if (rwRes.status == rewrite_status::RW_DONE) {
+      m_resultStack.push_back(rwRes.rewritten);
+      if (exp->use_count() > 1) {
```
agurfinkel, line 124: replace `exp->use_count() > 1` with `shouldCache` function for readability. Here and elsewhere

include/seahorn/Expr/ExprRewriter.hh
```diff
@@ -0,0 +1,144 @@
...
+    size_t begin = end - arity;
+    for (size_t i = begin; i < end; i++) {
+      new_kids.push_back(m_resultStack[i]);
+    }
+    Expr new_exp =
+        exp->getFactory().mkNary(exp->op(), new_kids.begin(), new_kids.end());
+    m_resultStack.resize(begin);
+    // apply rewrite rules to expression with new kids
+    rewrite_result rwRes = m_config.applyRewriteRules(new_exp);
+    if (rwRes.status == rewrite_status::RW_DONE) {
```
agurfinkel, line 122: there has to be a way for a rule to indicate that no rewrite has happened. In this case, the result should not be cached.

include/seahorn/Expr/ExprVisitor.hh
```diff
@@ -84,11 +85,25 @@ public:
...
   }
-
 };
 
 using DagVisitCache = std::unordered_map<ENode *, Expr>;
 
+/* Pair <mem, ptr> */
+using DagVisitMemCache = std::unordered_map<ENode *, ExprPair>;
+
+inline Expr findInDagVisitCache(DagVisitCache &cache, Expr expr) {
```
agurfinkel, line 95: what is this for? Document

agurfinkel (changes requested):

include/seahorn/Expr/ExprRewriter.hh
danblitzhou: perhaps add a new status `RW_SKIPPED`

agurfinkel: `RW_SKIP`

include/seahorn/Expr/ExprVisitor.hh
danblitzhou: not sure if this is needed anymore

include/seahorn/Expr/ExprRewriteRule.hh
```diff
@@ -56,16 +56,24 @@ struct ITERewriteRule : public ExprRewriteRule {
...
+        return {mk<ITE>(i, t, e_t), rewrite_status::RW_1};
       }
     }
+    // ite(true, a, b) => a
+    if (isOpX<TRUE>(i)) {
+      return {t, rewrite_status::RW_1};
+    }
+    // ite(false, a, b) => b
+    if (isOpX<FALSE>(i)) {
+      return {e, rewrite_status::RW_1};
```
agurfinkel, line 76: `RW_DONE`

include/seahorn/Expr/ExprRewriteRule.hh
```diff
@@ -151,7 +152,45 @@ struct BoolOpRewriteRule : public ExprRewriteRule {
     if (isOpX<NEG>(exp)) {
       Expr neg = exp->arg(0);
       if (isOpX<NEG>(neg)) {
-        return {neg->arg(0), rewrite_status::RW_2};
+        return {neg->arg(0), rewrite_status::RW_1};
```
agurfinkel, line 155: RW_DONE

include/seahorn/Expr/ExprRewriteRule.hh
```diff
@@ -151,7 +152,45 @@ struct BoolOpRewriteRule : public ExprRewriteRule {
...
+      Expr arr = exp->arg(0);
+      Expr idx = exp->arg(1);
+      // select(ite(i, a, b), v) => ite(i, select(a, v), select(b, v))
+      if (isOpX<ITE>(arr)) {
+        Expr i = arr->arg(0);
+        Expr t = arr->arg(1);
+        Expr e = arr->arg(2);
+        Expr new_t = op::array::select(t, idx);
+        Expr new_e = op::array::select(e, idx);
+        return {mk<ITE>(i, new_t, new_e), rewrite_status::RW_2};
```
agurfinkel, line 181: overall, this is potentially very expensive in our application. It will create a lot of intermediate expressions while `select` is being pushed down. It would be more effective to implement `push_select_over_ite` method that pushes select to the bottom of the `ite` tree in one operation.

include/seahorn/Expr/ExprRewriteRule.hh
```diff
@@ -151,7 +152,45 @@ struct BoolOpRewriteRule : public ExprRewriteRule {
...
+      }
+      /** Read-over-write: select(store(arr_w, idx_w, val), idx_r)
+          ==> ite(idx_w == idx_x, val, select(arr_w, idx_r))
+       **/
+      if (isOpX<STORE>(arr)) {
```
agurfinkel, line 186: same comment as above. Need a single operation that pushes select down a chain of stores, and maybe a combination that pushes select through the chain of ite and store

include/seahorn/Expr/ExprRewriter.hh
```diff
@@ -0,0 +1,177 @@
...
+  CompareRewriteRule m_compRule;
+  BoolOpRewriteRule m_boolRule;
+  ArrayRewriteRule m_arrayRule;
+  ArithmeticRule m_arithRule;
+
+public:
+  ITECompRewriteConfig(ExprFactory &efac, DagVisitCache &cache,
+                       AddrRangeMap &arm)
+      : m_iteRule(efac, cache), m_compRule(efac, cache),
+        m_boolRule(efac, cache), m_arrayRule(efac, cache, arm),
```
LinerSu, line 60: Looks like you only use ARM for `select` only. Maybe I did not fully get the usage of the ARM. Can Arm be used for other pointer expressions?

lib/seahorn/Smt/ExprAddrRangeMap.cc
```diff
@@ -0,0 +1,79 @@
...
+  return OS;
+}
+
+std::ostream &operator<<(std::ostream &OS, AddrRangeMap const &arm) {
+  arm.print(OS);
+  return OS;
+}
+
+AddrRange zeroBitsRange(AddrRange &r, size_t bits) {
+  unsigned new_low = r.low >> bits;
```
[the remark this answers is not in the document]
danblitzhou: good catch so that we can skip some operations

lib/seahorn/Smt/ExprRewriter.cc
```diff
@@ -0,0 +1,191 @@
...
+        res = mk<ITE>(simpIN, res, newE);
+      }
+    }
+    nested.pop_back();
+  }
+  return res;
+}
+} // namespace utils
+
+Expr rewriteHybridLoadMemExpr(Expr loadMem, Expr ptr, AddrRangeMap &arm) {
```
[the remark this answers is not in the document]
danblitzhou: good catch

include/seahorn/Expr/ExprRewriter.hh
danblitzhou: we can perhaps use it during the initial rewrite from `store`s to `ITE`s

agurfinkel:
@danblitzhou can we close this PR?

danblitzhou:
Yes, the new PR should replace this one.

Arie Gurfinkel ***@***.***>于2022年12月9日 周五下午3:15写道：

PR #402, opened 2021-06-04
Fixed conflict between partial function synthesis and inlining.

ScottWe:


agurfinkel:
looks like there are formatting issues to fix

agurfinkel (commented):
LGTM

PR #403, opened 2021-06-04
Fixed bug in KleeInternalize.

ScottWe:
The new `__SEA_assume` primitive is not exempt for `KleeInternalize`. This breaks executable counter-examples.


agurfinkel:
formatting

PR #406, opened 2021-06-07
fix(snyth): synthesis assertions are preserved

ScottWe:
Previously, synthesis assertions were encoded using `__VERIFIER_assert(x)`, and did not generate calls to `__VERIFIER_error()`. If the assertion did not appear on a path to `__VERIFIER_error()`, then the assertion could be optimized away. This commit resolves the issue by extending `sassert(x)` for synthesis (similar to `VACCHECK`).

The synthesis version of `sassert` is enabled by defining `SEA_SYNTH`. The `PARTIAL_FN` macro is now guarded by `SEA_SYNTH`. By guarding `PARTIAL_FN`, it is ensured that should an end-user forget to define `SEA_SYNTH`, then compilation will fail.

This change also impacts the promotion of synthesis calls in `PromoteVerifierCalls.cc`. Before this change, PromoteVerifierCalls would assume the partial function after each synthesis call. This assumption is now generated automatically by `sassert(x)`. Following this change, PromoteVerifierCalls only generates assumptions for sea.synth.assume.

### DIFF
--- a/include/seahorn/seahorn.h
+++ b/include/seahorn/seahorn.h
@@ -55,14 +55,18 @@ extern void sea_reset_modified(char *);
 /* Convenience macros */
 #define assume __SEA_assume
 
+#ifdef VACCHECK
+/* See https://github.com/seahorn/seahorn/projects/5 for details */
+#define sassert(X)                                                             \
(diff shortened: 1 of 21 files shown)


agurfinkel (approved):
LGTM

ScottWe:
Is this waiting on any changes?

PR #411, opened 2021-06-17
fix(boost): resolve ambiguous under boost namepsace

LinerSu:
Fix ambiguity caused by using the namespace `expr::op::variant`.


agurfinkel:
@LinerSu are you sure that this is needed? I think the root cause is inclusion of a wrong boost header file somewhere where it is not needed. Can you explain the root cause. When did this happen, what commit causes it, etc.

LinerSu:
I think this is not needed. Close it.

PR #413, opened 2021-06-24
refactor(opsem): refactor mem mgr to allow object mem

priyasiddharth:
Templatize TrackingMemMgr and ExtraWideMemMgr so that they can take in any mem mgr, not just raw mem mgr

ObjectMemMgr (to be added) will be a sibling to RawMemMgr

### DIFF
--- a/lib/seahorn/BvOpSem2Context.hh
+++ b/lib/seahorn/BvOpSem2Context.hh
@@ -39,6 +39,11 @@ auto widemem_tag_of =
   return {};
 };
 
+auto objectmem_tag_of =
+    [](auto t) -> hana::type<typename decltype(t)::type::ObjectMemTag> {
+  return {};
+};
+
 // This empty class is used as a 'tag' to mark containing classes as enabling
 // features
 // feature: tracking.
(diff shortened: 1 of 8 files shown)


lib/seahorn/BvOpSem2Context.hh
```diff
@@ -261,6 +274,9 @@ public:
   /// \f setMemReadRegister
   Expr loadValueFromMem(Expr ptr, const llvm::Type &ty, uint32_t align);
 
+  /// \brief join two memories according to semantics of memory manager
```
agurfinkel, line 277: What is the semantics of this operation? I was expecting the join to be conditional on some path condition, but it is not in the interface

agurfinkel (commented):
what are semantics of join?

lib/seahorn/BvOpSem2Context.hh
priyasiddharth: Fixed. 
1. `joinMemories` will take in a join `cond`  e.g. at a `select` operation
2. The rest of the path conditions are carried in `mem1` and `mem2`.

PR #415, opened 2021-09-08
feat(opsem): integrate crab to lower sea-is-deref

LinerSu:
Use crab to infer the results of  `sea.is_dereferenceable`.

### DIFF
--- a/lib/seahorn/BvOpSem2.cc
+++ b/lib/seahorn/BvOpSem2.cc
@@ -43,6 +43,7 @@ using gep_type_iterator = generic_gep_type_iterator<>;
 
 namespace seahorn {
 extern bool isUnifiedAssume(const Instruction &CI);
+extern clam::CrabDomain::Type CrabDom;
 namespace details {
 enum class VacCheckOptions { NONE, ANTE, ALL };
 }
@@ -148,6 +149,11 @@ static llvm::cl::opt<bool>
     UseCrabInferRng("horn-bv2-crab-rng",
                     llvm::cl::desc("Use crab to infer rng invariants"),
                     llvm::cl::init(false));
+
+static llvm::cl::opt<bool> UseCrabLowerIsDeref(
+    "horn-bv2-crab-lower-is-deref",
+    llvm::cl::desc("Use crab to lower sea.is_dereferenceable"),
+    llvm::cl::init(false));
 
 static llvm::cl::opt<bool> UseLVIInferRng(
     "horn-bv2-lvi-rng",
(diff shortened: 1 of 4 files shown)


lib/seahorn/BvOpSem2.cc
```diff
@@ -149,6 +150,30 @@ static llvm::cl::opt<bool>
                     llvm::cl::desc("Use crab to infer rng invariants"),
                     llvm::cl::init(false));
 
+static llvm::cl::opt<bool> UseCrabLowerIsDeref(
+    "horn-bv2-crab-lower-is-deref",
+    llvm::cl::desc("Use crab to lower sea.is_dereferenceable"),
+    llvm::cl::init(false));
+
+static llvm::cl::opt<clam::CrabDomain::Type, false, clam::CrabDomainParser>
```
caballa, line 158: `PathBmc.cc` already defines a command-line option to choose the abstraction domain. 
Replace this definition of `CrabDom` with this code:

```
namespace seahorn {
  extern clam::CrabDomain::Type CrabDom;
}
```

lib/seahorn/BvOpSem2.cc
```diff
@@ -3228,9 +3273,14 @@ void Bv2OpSem::initCrabAnalysis(const llvm::Module &M) {
 
   // -- Set parameters for CFG
   clam::CrabBuilderParams cfg_builder_params;
-  LOG("opsem-rng", cfg_builder_params.print_cfg = true;);
+  LOG("opsem-crab", cfg_builder_params.print_cfg = true;);
   cfg_builder_params.setPrecision(clam::CrabBuilderPrecision::MEM);
   cfg_builder_params.interprocedural = true;
+  if (UseCrabLowerIsDeref) {
+    cfg_builder_params.add_bounds_checks = true;
```
caballa, line 3280: Remove `add_bounds_checks`. This option will adds also checks for load and stores. We don't want that.

lib/seahorn/BvOpSem2.cc
```diff
@@ -3228,9 +3273,14 @@ void Bv2OpSem::initCrabAnalysis(const llvm::Module &M) {
...
+    cfg_builder_params.add_is_deref = true;
+  }
+  cfg_builder_params.lowerUnsignedICmpIntoSigned();
```
caballa, line 3283: You need also this option:
```
cfg_builder_params.lower_arithmetic_with_overflow_intrinsics  = true;
```

lib/seahorn/BvOpSem2.cc
```diff
@@ -3242,12 +3292,12 @@ void Bv2OpSem::initCrabAnalysis(const llvm::Module &M) {
 void Bv2OpSem::runCrabAnalysis() {
   /// Set Crab parameters
   clam::AnalysisParams aparams;
-  aparams.dom = clam::CrabDomain::INTERVALS;
+  aparams.dom = CrabDom;
   aparams.run_inter = true;
-  aparams.check = clam::CheckerKind::NOCHECKS;
+  aparams.check = clam::CheckerKind::ASSERTION;
```
caballa, line 3297: I don't think we need the `ASSERTION` option. This will run the checker in Crab but we don't need/want that

lib/seahorn/BvOpSem2.cc
```diff
@@ -29,6 +29,7 @@
 #include "clam/CfgBuilder.hh"
 #include "clam/Clam.hh"
 #include "clam/ClamQueryAPI.hh"
+#include "clam/CrabDomainParser.hh"
```
caballa, line 32: I think you don't need this.

caballa:
This PR looks good to me but we need to add some tests.
@yusen: can you add some tests to exercise this new feature?

LinerSu:
I added two tests.

caballa:
Do the tests pass with zones? 
I don't want to use `pk` unless strictly necessary because the use of Elina/Apron is optional.

caballa:
Merged in commit 51f04bff838ae4a830987beb43442513e21b98ba

PR #425, opened 2021-11-17
fix(cex): handle invalid memmap when build memhavoc

### LINKED ISSUE #424: dev10 `cex` mode crash
**Describe the bug**

Modify the [file](https://sourcegraph.com/github.com/seahorn/verify-c-common/-/blob/seahorn/jobs/byte_buf_write_from_whole_string/aws_byte_buf_write_from_whole_string_harness.c?subtree=true) as given in this branch: https://github.com/priyasiddharth/verify-c-common/tree/bug

commit: https://github.com/priyasiddharth/verify-c-common/commit/ff1c47348db052db93bd09c46598043d0e9cfdfb

**To Reproduce**
1. Link to an input file (bitcode is preferred)

2. Full command line options
Use `dev10` seahorn branch
 ninja && ./verify --cex  seahorn/jobs/byte_buf_write_from_whole_string/ 
**Expected behavior**
cex printed and seahorn exits successfully

**Logs, Stacktraces, and Screenshots**
If applicable, add a log, a screenshots, or a stack trace to help explain your problem.
```
seahorn: /home/siddharth/seahorn/llvm-project/llvm/lib/IR/Type.cpp:245: static llvm::IntegerType* llvm::IntegerType::get(llvm::LLVMContext&,
 unsigned int): Assertion `NumBits >= MIN_INT_BITS && "bitwidth too small"' failed.
/home/siddharth/seahorn/seahorn/build-dbg/run/bin/seahorn[0x48a1fc1]
/home/siddharth/seahorn/seahorn/build-dbg/run/bin/seahorn[0x48a2058]
/home/siddharth/seahorn/seahorn/build-dbg/run/bin/seahorn[0x489fdb9]
/home/siddharth/seahorn/seahorn/build-dbg/run/bin/seahorn[0x48a1918]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x141f0)[0x7fb4cf5f91f0]
/lib/x86_64-linux-gnu/libc.so.6(gsignal+0xcb)[0x7fb4cd789fbb]
/lib/x86_64-linux-gnu/libc.so.6(abort+0x116)[0x7fb4cd76f864]
/lib/x86_64-linux-gnu/libc.so.6(+0x26749)[0x7fb4cd76f749]
/lib/x86_64-linux-gnu/libc.so.6(+0x383d6)[0x7fb4cd7813d6]
...
```

**Desktop (please complete the following information):**
 - OS: [e.g. Linux, Mac]
 - LLVM: [e.g., llvm10]
 - Z3: [e.g. z3-master]
Linux, llvm 10 (dev10)
**Additional context**

danblitzhou:
- when synthesizing `memhavoc`, return all 0 constant array if `ExprMemMap` cannot parse a valid mem map.
- explicitly checks if mem map content width is not less than `llvm::IntegerType::MIN_INT_BITS`; `is_valid` field should have been enough but just adding these as sanity checks.
- fixes #424 , tested using `debug` build against vcc `byte_buf_write_from_whole_string`

### DIFF
--- a/lib/seahorn/CexHarness.cc
+++ b/lib/seahorn/CexHarness.cc
@@ -69,7 +69,7 @@ Constant *exprToLlvm(Type *ty, Expr e, LLVMContext &ctx, const DataLayout &dl) {
 }
 
 /* Given Expr of a shadow.mem segment and pre-recorded size Value
-   use HexDump to extract const array with item size of current
+   use MemMap to extract const array with item size of current
    word size from Expr
 */
 Constant *exprToMemSegment(Expr segment, Expr startAddr, Expr size,
@@ -104,10 +104,16 @@ Constant *exprToMemSegment(Expr segment, Expr startAddr, Expr size,
     ArrayType *placeholderT = ArrayType::get(Type::getInt8PtrTy(ctx), 0);
     return ConstantArray::get(placeholderT, LLVMValueSegment);
   }
-
+  // use MemMap to extract mem segment info
   using MemMap = expr::exprMemMap::ExprMemMap;
   const MemMap m_map(segment);
   size_t elmWidth = m_map.getContentWidth();
+  if (!m_map.isValid() || elmWidth < IntegerType::MIN_INT_BITS) {
+    LOG("cex_verbose", WARN << "memhavoc: invalid memory expression: "
+                            << *m_map.getRawExpr() << "\n");
+    ArrayType *placeholderT = ArrayType::get(Type::getInt8PtrTy(ctx), 0);
+    return ConstantArray::get(placeholderT, LLVMValueSegment);
+  }
   size_t blocks = std::ceil((float)blockWidth / (float)elmWidth);
   auto *segmentElmTy = IntegerType::get(ctx, elmWidth * 8);
   ArrayType *segmentAT = ArrayType::get(segmentElmTy, blocks);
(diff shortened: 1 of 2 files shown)


lib/seahorn/CexHarness.cc
```diff
@@ -104,10 +104,16 @@ Constant *exprToMemSegment(Expr segment, Expr startAddr, Expr size,
     ArrayType *placeholderT = ArrayType::get(Type::getInt8PtrTy(ctx), 0);
     return ConstantArray::get(placeholderT, LLVMValueSegment);
   }
-
+  // use MemMap to extract mem segment info
   using MemMap = expr::exprMemMap::ExprMemMap;
   const MemMap m_map(segment);
   size_t elmWidth = m_map.getContentWidth();
+  if (!m_map.isValid() || elmWidth < IntegerType::MIN_INT_BITS) {
+    LOG("cex", errs() << "memhavoc: invalid memory expression: "
```
agurfinkel, line 112: maybe better to use `WARN` macro? Is it an expected or unexpected behavior? 

SMT solver will not produce an array value unless the value is used. So, for example, if region is first `memhavoc` but then properly initialized, the initial value of the region will not be used and will not be computed. This is expected and desired behavior. 

One thing that has changed is that @priyasiddharth added more `memhavoc` calls to fix the fuzzer, so perhaps that is why it was missed.

TL;DR: the log should probably be `WARN` and should probably be silenced by default in `cex` log level since it is expected behavior in many cases. Maybe put it under `cex_verbose` log level.

agurfinkel (approved):

lib/seahorn/CexHarness.cc
danblitzhou: will change the logging lvl, I wasn't sure if this is like the other special handlings. Not sure why there wasn't any division by zero errors in line `size_t blocks = std::ceil((float)blockWidth / (float)elmWidth);`

priyasiddharth (approved):

PR #427, opened 2021-11-25
feat(vac): dump formula on solver error

priyasiddharth:
when solver returns "indeterminate" then output smt2 formula to file

### DIFF
--- a/lib/seahorn/BvOpSem2.cc
+++ b/lib/seahorn/BvOpSem2.cc
@@ -975,7 +975,21 @@ public:
     m_ctx.addToSolver(m_ctx.getPathCond());
     m_ctx.addToSolver(solveFor);
 
-    return m_ctx.solve();
+    auto r = m_ctx.solve();
+    // if the solver returns unknown then dump smt2 formula to file for analysis
+    LOG("opsem.dump.incassert", {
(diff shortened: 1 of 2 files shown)


lib/seahorn/BvOpSem2.cc
```diff
@@ -975,7 +975,22 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemVisitorBase {
     m_ctx.addToSolver(m_ctx.getPathCond());
     m_ctx.addToSolver(solveFor);
 
-    return m_ctx.solve();
+    auto r = m_ctx.solve();
+    // if the solver returns unknown then dump smt2 formula to file for analysis
+    LOG(
+        "opsem.dump.incassert", if (boost::logic::indeterminate(r)) {
```
agurfinkel, line 981: too many ifs. 

put code into `{ }` for formatting to work better

don't rely on `boost::logic::indeterminate`. You can instead do
```if (r || !r) {} else { YOUR CODE GOES HERE }```

Easier to read and does not depend on any particular implementation of `tribool`

agurfinkel (approved):
good except for little change requested

agurfinkel (approved):

PR #428, opened 2021-12-07
fix(opsem): in lambda mode, convert const array to lambda after simplify

priyasiddharth:
Note that this doesn't overload the simplification process with the conversion task since I thought it better to not pass unnecessary flag information to the solver.

### DIFF
--- a/lib/seahorn/BvOpSem2.cc
+++ b/lib/seahorn/BvOpSem2.cc
@@ -2386,11 +2386,42 @@ Bv2OpSemContext::Bv2OpSemContext(SymStore &values, ExprVector &side,
   setPathCond(o.getPathCond());
 }
 
+// If running in lambda mode,
+// convert constant arrays returned by Z3 simplifier to a lambda.
+//
+// Details: When Z3 is given a lambda of the form lambda x: 0 to simplify, a
+// constant array is returned. If running in lambda mode, this will cause the
+// formula to have both lambdas and constant arrays which is not supported
+// (fully) by Z3 and y2.
+// This workaround fixes the issue by converting constant arrays gotten from
+// Z3 back to lambda x: 0.
+Expr coerceConstArrayToLambda(Expr e, ExprFactory &efac) {
+  // go into structs until a non struct is found; convert to lambda if needed.
+  if (strct::isStructVal(e)) {
+    llvm::SmallVector<Expr, 8> kids;
+    for (unsigned i = 0, sz = e->arity(); i < sz; ++i) {
+
+      kids.push_back(coerceConstArrayToLambda(e->arg(i), efac));
+    }
+    return strct::mk(kids);
+  } else if (isOpX<CONST_ARRAY>(e)) {
+    Expr dom = e->arg(0); // get domain
+    Expr val = e->arg(1); // get value
+    Expr addr = bind::mkConst(mkTerm<std::string>("addr", efac), dom);
+    Expr decl = bind::fname(addr);
+    Expr res = mk<LAMBDA>(decl, val);
+    return res;
+  } else {
+    return e;
+  }
+}
+
 Expr Bv2OpSemContext::simplify(Expr u) {
   ScopedStats _st_("opsem.simplify");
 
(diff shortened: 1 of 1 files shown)


lib/seahorn/BvOpSem2.cc
```diff
@@ -2386,11 +2386,37 @@ Bv2OpSemContext::Bv2OpSemContext(SymStore &values, ExprVector &side,
   setPathCond(o.getPathCond());
 }
 
+Expr Bv2OpSemContext::convertConstArrayToLambda(Expr e) {
```
agurfinkel, line 2389: we use the term `coerce` to express any form of coercion. Would be good to be consistent.

lib/seahorn/BvOpSem2.cc
```diff
@@ -2386,11 +2386,37 @@ Bv2OpSemContext::Bv2OpSemContext(SymStore &values, ExprVector &side,
...
+
+      kids.push_back(convertConstArrayToLambda(e->arg(i)));
+    }
+    return strct::mk(kids);
+  } else if (isOpX<CONST_ARRAY>(e)) {
+    Expr dom = e->arg(0); // get domain
+    Expr val = e->arg(1); // get value
+    Expr addr = bind::mkConst(mkTerm<std::string>("addr", efac()), dom);
+    Expr decl = bind::fname(addr);
+    Expr res = mk<LAMBDA>(decl, val);
```
agurfinkel, line 2403: I don't think this is correct construction. I will check and get back.
I believe `mk<LAMBDA>` expects the body already to be over bound variables.

lib/seahorn/BvOpSem2.cc
```diff
@@ -2386,11 +2386,37 @@ Bv2OpSemContext::Bv2OpSemContext(SymStore &values, ExprVector &side,
...
+
 Expr Bv2OpSemContext::simplify(Expr u) {
   ScopedStats _st_("opsem.simplify");
 
-  Expr _u;
-  _u = m_z3_simplifier->simplify(u);
+  Expr __u, _u;
+  __u = m_z3_simplifier->simplify(u);
+  if (UseLambdas) {
+    _u = convertConstArrayToLambda(__u);
```
agurfinkel, line 2416: will be more readable as a ternary ite operation
```
_u = UseLambdas ? ... 
```

lib/seahorn/BvOpSem2Context.hh
```diff
@@ -361,6 +361,17 @@ public:
...
+  // If running in lambda mode,
+  // convert constant arrays returned by Z3 simplifier to a lambda.
+  //
+  // Details: When Z3 is given a lambda of the form lambda x: 0 to simplify, a
+  // constant array is returned. If running in lambda mode, this will cause the
+  // formula to have both lambdas and constant arrays which is not supported
+  // (fully) by Z3 and y2.
+  // This workaround fixes the issue by converting constant arrays gotten from
+  // Z3 back to lambda x: 0.
+  Expr convertConstArrayToLambda(Expr e);
```
agurfinkel, line 373: why does this need to be a class method? just a private function.

lib/seahorn/BvOpSem2.cc
agurfinkel: I take my comment back, this is correct because this is a constant function. So `val` does not depend on the argument, hence, there is neither `addr` nor corresponding bound variable in `val`!

agurfinkel (commented):
there are only minor changes on name. everything looks great ow

lib/seahorn/BvOpSem2Context.hh
priyasiddharth: done

lib/seahorn/BvOpSem2.cc
priyasiddharth: good point! done.

lib/seahorn/BvOpSem2.cc
priyasiddharth: done

PR #429, opened 2021-12-08
feat(y2): marshal const-array w/ lambda

danblitzhou:
Idea from https://github.com/SRI-CSL/yices2/issues/271#issuecomment-664726905, use lambda function to represent const array for Yices2

### DIFF
--- a/lib/seahorn/Smt/MarshalYices.cc
+++ b/lib/seahorn/Smt/MarshalYices.cc
@@ -258,26 +258,26 @@ term_t marshal_yices::encode_term(Expr e, ycache_t &cache) {
       encode_term_fail(e, "unhandled arity 1 case");
     }
   } else if (arity == 2) {
-
+    term_t t1, t2;
     if (isOpX<REM>(e)) {
       encode_term_fail(e, "Integer remainder not supported in yices");
-    } else if (isOpX<CONST_ARRAY>(e)) {
+    }
+    if (!isOpX<CONST_ARRAY>(e)) {
+      t1 = encode_term(e->left(), cache);
+    }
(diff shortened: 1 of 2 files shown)


lib/seahorn/Smt/MarshalYices.cc
```diff
@@ -262,7 +262,13 @@ term_t marshal_yices::encode_term(Expr e, ycache_t &cache) {
...
       encode_term_fail(e, "Integer remainder not supported in yices");
     } else if (isOpX<CONST_ARRAY>(e)) {
-      encode_term_fail(e, "const-array term not supported in yices");
+      // (const-array sort v) --> lambda i:sort :: v
+      // idea from:
+      // https://github.com/SRI-CSL/yices2/issues/271#issuecomment-664726905
+      type_t indexType = encode_type(e->left());
+      term_t i = yices_new_variable(indexType);
+      term_t v = encode_term(e->right(), cache);
+      return yices_lambda(1, &i, v);
```
agurfinkel, line 271: because of the early return, you are missing caching. This supposed to assign `res` variable like the surrounding code, and then at the end of the function `res` is cached.

danblitzhou: Maybe do early caching here? I don't like the idea of going through rest of the `else if` blocks and do type checking.

agurfinkel: structure the code so that caching happens but of course other branches should not be evaluated

agurfinkel:
looks ready, right? @danblitzhou

agurfinkel:
ok, don't forget to merge or let me know. I don't get notified when CI completes and might forget by that time :)

PR #431, opened 2022-02-27
fix(opsem): init fresh ptr name in widememmgrs

priyasiddharth:
Added test to replicate failure and validate fix

Why was this failure not detected?

Because we never set the symbolic memory threshold very low thus escaping the need to use fresh pointers and exercise faulty code path

### DIFF
--- a/lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
+++ b/lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
@@ -447,7 +447,7 @@ typename ExtraWideMemManager<T>::PtrTy ExtraWideMemManager<T>::nullPtr() const {
 }
 template <class T>
 typename ExtraWideMemManager<T>::PtrTy ExtraWideMemManager<T>::freshPtr() {
-  Expr name = op::variant::variant(m_id++, m_freshPtrName);
+  Expr name = m_main.freshPtr();
   return mkAlignedPtr(name, m_alignment);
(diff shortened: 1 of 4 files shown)


lib/seahorn/BvOpSem2ExtraWideMemMgr.cc
```diff
@@ -26,6 +26,7 @@ ExtraWideMemManager<T>::ExtraWideMemManager(Bv2OpSem &sem, Bv2OpSemContext &ctx,
       m_offset(sem, ctx, ptrSz, ptrSz, useLambdas, true),
       m_size(sem, ctx, ptrSz, g_slotByteWidth, useLambdas, true),
       m_uninit_size(m_ctx.alu().ui(g_uninit_size, g_slotBitWidth)),
+      m_freshPtrName(mkTerm<std::string>("sea.ptr", m_efac)), m_id(0),
```
agurfinkel, line 29: why should this functionality be duplicated from RawMemManager and not simply reused from there. All the wide mem managers already depend on other, simpler, mem managers, and the raw one provides names to allocator.

If this is duplicated, at least the names should be different, i.e., not `sea.ptr`, to simplify debugging later on. But I really think it should be properly refactored.

priyasiddharth: Good point! fixed. Will submit after tests complete.

PR #432, opened 2022-02-28
fix(allocator): use assumed maximal allocation size if crab/LVI infers unbounded values

LinerSu:
The old implementation does not consider if the constant range is empty. Now fix the code to handle that situation.

### DIFF
--- a/lib/seahorn/BvOpSem2Allocators.cc
+++ b/lib/seahorn/BvOpSem2Allocators.cc
@@ -303,19 +303,22 @@ public:
       auto rangeValCrab = m_sem.getCrabInstRng(*operand_inst);
       auto rangeValLVI = m_sem.getLVIInstRng(*operand_inst);
       uint64_t ub_crab = rangeValCrab.getUpper().getLimitedValue();
+      if (rangeValCrab.isFullSet() || rangeValCrab.isEmptySet()) {
+        // not useful range inferred by crab
+        ub_crab = m_maxSymbAllocSz; // take max allocation size as default
+      }
       uint64_t ub_lvi = rangeValLVI.getUpper().getLimitedValue();
+      if (rangeValLVI.isFullSet() || rangeValLVI.isEmptySet()) {
+        // not useful range inferred by LVI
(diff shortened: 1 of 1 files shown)


lib/seahorn/BvOpSem2Allocators.cc
```diff
@@ -303,19 +303,22 @@ class StaticOpSemAllocator : public OpSemAllocator {
       auto rangeValCrab = m_sem.getCrabInstRng(*operand_inst);
       auto rangeValLVI = m_sem.getLVIInstRng(*operand_inst);
       uint64_t ub_crab = rangeValCrab.getUpper().getLimitedValue();
+      if (rangeValCrab.isFullSet() || rangeValCrab.isEmptySet()) {
```
caballa, line 306: This code is sound but maybe too conservative. If a range is empty then it means that the instruction is unreachable so the check is actually safe (see comments here https://github.com/seahorn/clam/blob/master/include/clam/ClamQueryAPI.hh#L27).

LinerSu: Should be inserted a call for `llvm_unreachable` if the set is empty?

caballa: No. My point was that it is perfectly fine that a memory allocation is unreachable and clam can detect that by returning an empty interval. `llvm_unreachable` has a different purpose. This function aborts and it's intended to be called when some code (at the llvm side) is executed but it shouldn't. The unreachability I'm talking about is in the analyzed code not in the SeaHorn code.

caballa: I thought this code was related to `is_dereferenceable` so that's why having an empty interval it can avoid a check. But then I realized this is different. It's just for helping the BMC engine to reason about allocation sizes. So I would leave it like that. Sorry about the digression.

LinerSu: Yes. This code is used for inferring allocation size. Thanks for telling me the `llvm_unreachable` thing.

agurfinkel (approved):

PR #435, opened 2022-05-09
fix(opsem): cast alloc size to slot size in salloc

### LINKED ISSUE #434: VCGen error due to type inconsistency with BMC+opsem2+extra widen mem manager
**Describe the bug**

`Error: failed to convert to z3: ite(main@%_11_0, 0:bv(64), main@%_6_0)`

The error happens when pointer size is 4 bytes. Since `main@%_6_0` is bv(32)  the construction of the ite term fails.
I get the error when compiling on Mac M1 but I think the problem should also happen if we pass the option `-m32` to clang.

The VCGen uses opsem2  with extra widen memory manager. I think the relevant part of the code is here:

https://github.com/seahorn/seahorn/blob/dev10/lib/seahorn/BvOpSem2ExtraWideMemMgr.cc#L29

The fact that `g_slotBitWidth` is fixed to 64 is suspicious:

https://github.com/seahorn/seahorn/blob/dev10/lib/seahorn/BvOpSem2ExtraWideMemMgr.cc#L12

**To Reproduce**

I attach the bitcode 
[array_eq_c_str_ignore_case.ir.bc.zip](https://github.com/seahorn/seahorn/files/8653389/array_eq_c_str_ignore_case.ir.bc.zip)

The full command is:

```
sea fpf -O3 --inline --enable-loop-idiom --enable-indvar --no-lower-gv-init-struct --externalize-addr-taken-functions --no-kill-vaarg --with-arith-overflow=true --horn-unify-assumes=true --horn-gsa --no-fat-fns=bcmp,memcpy,assert_bytes_match,ensure_linked_list_is_allocated,sea_aws_linked_list_is_valid --dsa=sea-cs-t --devirt-functions=types --bmc=opsem --horn-vcgen-use-ite --horn-vcgen-only-dataflow=true --horn-bmc-coi=true --sea-opsem-allocator=static --horn-explicit-sp0=false --horn-bv2-lambdas --horn-bv2-simplify=true --horn-bv2-extra-widemem  --horn-stats=true --bound=10 array_eq_c_str_ignore_case.ir.bc 
```

**Expected behavior**
You should get something like this:

```
!4 Alloca of (main@%_6_0 * 1) bytes:   %malloc8.i = alloca i8, i32 %_6, align 4, !sea.dsa.allocsite !730
Adding range condition: bvule(main@%_6_0, 4096:bv(32))
Error: failed to convert to z3: ite(main@%_11_0, 0:bv(64), main@%_6_0)

Error: Sorts (_ BitVec 64) and (_ BitVec 32) are incompatible
```

**Desktop (please complete the following information):**

SeaHorn `dev10` branch. 
The OS is irrelevant since I already give you the bitcode as generated from Mac M1.
The Z3 version is also irrelevant for this bug.

priyasiddharth:
When allocating an expression size of bytes, we need to make sure the expression sort matches the size slot sort

I've added a test from the original bug report

fixes #434


priyasiddharth:
Merging without review since this is small change

PR #436, opened 2022-05-18
refactor(opsem): decouple sub mem mgr tuple from trackingmemmgr

priyasiddharth:
details:
Replace hardcoded sub mem managers in TrackingMemMgr by a
configurable struct of Sub MemoryManagers.
This allows adding new shadow slots without necessarily
touching TrackingMemoryMgr.
Inside TrackingMemoryMgr, calls to sub memory managers are
unrolled at compile time except calls to set/get metadata
methods since they take in a metadata kind which is only
known at runtime. This can be optimized away in a separate effort.

more details:
Most of memory manager code involves taking a memory tuple(MemValTy),
operating on some or all elements and giving a memory tuple back.
Since most methods operate in this pure way, they are good candidates to
use a functional style of programming.
This change makes extensive use of boost::hana to work with memory tuples.
Hana has zero-cost abstraction since the loops are unrolled at
compile time so all iteration should be preprocessed to a bunch of
hand crafted statements.

### DIFF
--- a/lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
+++ b/lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
@@ -8,25 +8,35 @@ const unsigned int TrackingRawMemManager::g_MetadataByteWidth =
     TrackingRawMemManager::g_MetadataBitWidth / 8;
 const unsigned int TrackingRawMemManager::g_num_slots = 4;
 
+TrackingMemoryTuple::TrackingMemoryTuple(Bv2OpSem &sem, Bv2OpSemContext &ctx,
+                                         unsigned ptrSz, unsigned wordSz,
+                                         bool useLambdas)
+    : m_main(sem, ctx, ptrSz, wordSz, useLambdas),
+      m_w_metadata(sem, ctx, ptrSz, TrackingRawMemManager::g_MetadataByteWidth,
+                   useLambdas, true),
+      m_r_metadata(sem, ctx, ptrSz, TrackingRawMemManager::g_MetadataByteWidth,
+                   useLambdas, true),
+      m_a_metadata(sem, ctx, ptrSz, TrackingRawMemManager::g_MetadataByteWidth,
+                   useLambdas, true) {}
+
+TrackingMemoryTuple::TrackingMemoryTuple(Bv2OpSem &sem, Bv2OpSemContext &ctx,
+                                         unsigned ptrSz, unsigned wordSz,
+                                         bool useLambdas, bool ignoreAlignment)
+    : m_main(sem, ctx, ptrSz, wordSz, useLambdas, ignoreAlignment),
+      m_w_metadata(sem, ctx, ptrSz, TrackingRawMemManager::g_MetadataByteWidth,
+                   useLambdas, true),
+      m_r_metadata(sem, ctx, ptrSz, TrackingRawMemManager::g_MetadataByteWidth,
+                   useLambdas, true),
+      m_a_metadata(sem, ctx, ptrSz, TrackingRawMemManager::g_MetadataByteWidth,
+                   useLambdas, true) {}
+
 TrackingRawMemManager::TrackingRawMemManager(Bv2OpSem &sem,
                                              Bv2OpSemContext &ctx,
                                              unsigned ptrSz, unsigned wordSz,
                                              bool useLambdas)
     : MemManagerCore(sem, ctx, ptrSz, wordSz,
                      false /* this is a nop since we delegate to RawMemMgr */),
-      m_main(sem, ctx, ptrSz, wordSz, useLambdas),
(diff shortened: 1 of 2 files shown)


lib/seahorn/BvOpSem2TrackingRawMemMgr.hh
```diff
@@ -12,25 +12,48 @@
...
+//
+// Adding a new element (shadow slot) involves adding a new memory manager
+// inside this and in MetadataKind Use of MetadataKind is still needed because
+// setMetadata and getMetadata methods are virtual and thus calls to specific
+// metadata sub managers cannot be resolved at compile-time.
+struct TrackingMemoryTuple {
+  BOOST_HANA_DEFINE_STRUCT(TrackingMemoryTuple, (RawMemManager, m_main),
+                           (RawMemManager, m_r_metadata),
+                           (RawMemManager, m_w_metadata),
+                           (RawMemManager, m_a_metadata));
```
danblitzhou, line 32: to add a new shadow slot, do I just add a new element `(RawMemManager, m_x_metadata)`?

lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
```diff
@@ -70,83 +68,79 @@ Expr TrackingRawMemManager::isDereferenceable(TrackingRawMemManager::PtrTy p,
...
+  return MAIN_MEM_MGR.inttoptr(intVal, intTy, ptrTy);
 }
 TrackingRawMemManager::MemValTy TrackingRawMemManager::MemFill(
     TrackingRawMemManager::PtrTy dPtr, char *sPtr, unsigned int len,
     TrackingRawMemManager::MemValTy mem, uint32_t align) {
-  RawMemValTy rawVal = m_main.MemFill(dPtr, sPtr, len, mem.getRaw(), align);
-  return MemValTy(rawVal, mem.getMetadata(MetadataKind::READ),
-                  mem.getMetadata(MetadataKind::WRITE),
-                  mem.getMetadata(MetadataKind::ALLOC));
+  RawMemValTy rawVal = hana::at_key(m_submgrs, BOOST_HANA_STRING("m_main"))
```
danblitzhou, line 107: can you also use `MAIN_MEM_MGR` here like in other occassions?

lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
```diff
@@ -264,11 +256,24 @@ TrackingRawMemManager::memsetMetaData(MetadataKind kind, PtrTy ptr, Expr len,
   // make sure we can fit the supplied value in metadata memory slot
   assert(llvm::Log2_64(val) + 1 <= g_MetadataBitWidth &&
          "Metadata cannot fit!");
-  return MemValTy(
-      kind, memIn,
-      m_metadata_map[kind]->MemSet(ptr, m_ctx.alu().ui(val, g_MetadataBitWidth),
-                                   len, memIn.getMetadata(kind),
-                                   m_metadata_map[kind]->wordSzInBytes()));
+  auto index = static_cast<size_t>(kind) + 1;
```
danblitzhou, line 259: I am a bit confused for how `kind` and `memIn` are placed into `c`; perhaps this function can be more readable if all 3 elements in c are set explicitly.

lib/seahorn/BvOpSem2TrackingRawMemMgr.cc
```diff
@@ -277,119 +282,149 @@ TrackingRawMemManager::memsetMetaData(MetadataKind kind, PtrTy ptr,
...
   // TODO: expose a method in OpSemMemManager to loadAlignedWordFromMem
-  return m_metadata_map.at(kind)->loadIntFromMem(
-      ptr, memIn.getMetadata(kind), byteSz,
-      m_metadata_map.at(kind)->wordSizeInBytes());
+  auto index = static_cast<size_t>(kind) + 1;
+  auto r =
+      hana::make_range(hana::int_c<0>, TrackingMemoryTuple::GetTupleSize());
+  auto r_t = hana::to_tuple(r);
+  Expr ret;
+  hana::for_each(hana::zip(r_t, hana::keys(m_submgrs)), [&](auto pair) {
```
danblitzhou, line 312: I'm confused why we need to do a linear search here to retrieve a sub mem manager.

danblitzhou (commented):
LGTM, except for `memsetMetaData` and `getMetaData`, need some clarifications on why iteration is required.

danblitzhou:
> Hana has zero-cost abstraction since the loops are unrolled at
> compile time so all iteration should be preprocessed to a bunch of
> hand crafted statements.

I missed your last comment, so the loops would be simplified away at compile time. Still I feel like the code could be more readable if we just use `hana::at_key(m_submgrs, kind)` to retrieve a sub manager; perhaps this is not possible due to hana restrictions? 

> Hana has zero-cost abstraction since the loops are unrolled at
compile time so all iteration should be preprocessed to a bunch of
hand crafted statements.

PR #437, opened 2022-05-25
feat(opsem): added a shadow mem slot

priyasiddharth:
1. Added a shadow memory slot in opsem for tracking user configurable meta information about data in memory.

The API can be seen in the taint_tracking_sat.c and taint_tracking_unsat.c tests.

2. Refactored if..then..else code in BvOpSem2.cc and DfCoiAnalysis.cc into maps to make it easy to add new intrinsics.


include/seahorn/seahorn.h
```diff
@@ -47,7 +47,8 @@ extern void sea_tracking_on(void);
 extern void sea_tracking_off(void);
 /* reset modified metadata for memory pointed to by arg */
 extern void sea_reset_modified(char *);
-
+extern void sea_set_shadowmem(char, char *, char);
```
danblitzhou, line 50: need a brief comment description for `sea_set/get_shadowmem`

priyasiddharth: done

priyasiddharth:
@danblitzhou please let me know if you can review. Thank you.

lib/seahorn/BvOpSem2.cc
```diff
@@ -774,6 +797,55 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemVisitorBase {
...
+    }
+    Expr slot = lookup(*CS.getArgument(0));
+    if (!m_ctx.alu().isNum(slot)) {
+      LOG("opsem", ERR << "Metadata slot should resolve to a number.");
+      assert(false);
+    }
+    size_t slotNum = m_ctx.alu().toNum(slot).get_ui();
+    if (slotNum >= m_ctx.mem().getNumOfMetadataSlots()) {
+      LOG("opsem", ERR << "Metadata slot exceeds number of available slots.");
+      assert(false);
```
danblitzhou, line 815: `assert` only works in debug build I think, so is this the desired behaviour in other build types? or should we use `llvm::unreacheable`?

danblitzhou (approved):
Looks good, would be nice if we can have non-linear look-up in `visitPHINode`, but since number of functions are small it's fine for now.

lib/seahorn/BvOpSem2.cc
priyasiddharth: I think `llvm::unreachable` has undefined behavior and should not be used to describe an exception

https://llvm.org/docs/LangRef.html#unreachable-instruction

lib/seahorn/BvOpSem2.cc
```diff
@@ -610,42 +612,63 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemVisitorBase {
...
+        hana::make_pair(BOOST_HANA_STRING("smt."), &OpSemVisitor::visitSmtCall),
+        hana::make_pair(BOOST_HANA_STRING("sea.set_shadowmem"),
+                        &OpSemVisitor::visitSetShadowMem),
+        hana::make_pair(BOOST_HANA_STRING("sea.get_shadowmem"),
+                        &OpSemVisitor::visitGetShadowMem));
+
+    auto visitFunDecl = [&](StringRef candidate) {
+      auto found = false;
+      // Note: not efficient to loop through all keys.
+      // There is no way to break a hana::for_each loop
```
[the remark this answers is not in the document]
priyasiddharth: I haven't found a good solution. OTOH, having a massive switch is ugly. I will submit this for now and keep thinking about it.

PR #438, opened 2022-06-06
Crab in seapp

LinerSu:
This PR includes these features:

- A class to run crab analysis in `seapp`
- A LLVM pass to lower `sea.is_dereferenceable` in `seapp`
- A pipeline in `seapp` to support above features

### DIFF
--- a/include/seahorn/Transforms/Instrumentation/CrabAnalysis.hh
+++ b/include/seahorn/Transforms/Instrumentation/CrabAnalysis.hh
@@ -36,7 +36,8 @@ public:
   void runCrabAnalysisOnModule(const llvm::Module &M,
                                seadsa::GlobalAnalysis &dsa,
                                llvm::TargetLibraryInfoWrapperPass &tliPass);
-  clam::InterGlobalClam &getCrab() { return *m_crab; }
+  const clam::InterGlobalClam &getCrab() const;
+  clam::InterGlobalClam &getCrab();
 };
 } // namespace seahorn
 
--- a/include/seahorn/Transforms/Instrumentation/CrabAnalysis.hh
+++ b/include/seahorn/Transforms/Instrumentation/CrabAnalysis.hh
@@ -1,0 +1,43 @@
+#ifndef _CRAB_ANALYSIS_HH_
+#define _CRAB_ANALYSIS_HH_
+
+#include "llvm/IR/Module.h"
+
+namespace llvm {
+class TargetLibraryInfoWrapperPass;
+} // namespace llvm
+
+namespace seadsa {
+class ShadowMem;
+} // namespace seadsa
+
+namespace clam {
+class CrabBuilderManager;
+class InterGlobalClam;
+} // namespace clam
+
+namespace seahorn {
+using namespace llvm;
+
+class CrabAnalysis {
+  //// \brief crab's cfg builder manager
+  std::unique_ptr<clam::CrabBuilderManager> m_cfg_builder_man;
+  //// \brief crab instance to solve alloc bounds
+  std::shared_ptr<clam::InterGlobalClam> m_crab;
+
+  /// \brief Creates a crab's cfg builder manager
+  void initCrabAnalysis(const llvm::Module &M, seadsa::ShadowMem &dsaPass,
+                        llvm::TargetLibraryInfoWrapperPass &tliPass);
+  /// \brief Run crab analysis
+  void runCrabAnalysis();
+
+public:
+  CrabAnalysis() {}
+  void runCrabAnalysisOnModule(const llvm::Module &M,
+                               seadsa::ShadowMem &dsaPass,
+                               llvm::TargetLibraryInfoWrapperPass &tliPass);
+  std::shared_ptr<clam::InterGlobalClam> getCrab() { return m_crab; }
+};
+} // namespace seahorn
+
+#endif /* _CRAB_ANALYSIS_HH_ */
--- a/include/seahorn/Transforms/Instrumentation/CrabAnalysis.hh
+++ b/include/seahorn/Transforms/Instrumentation/CrabAnalysis.hh
@@ -22,8 +22,8 @@ using namespace llvm;
 class CrabAnalysis {
   //// \brief crab's cfg builder manager
   std::unique_ptr<clam::CrabBuilderManager> m_cfg_builder_man;
-  //// \brief crab instance to solve alloc bounds
-  std::shared_ptr<clam::InterGlobalClam> m_crab;
+  //// \brief crab instance
+  std::unique_ptr<clam::InterGlobalClam> m_crab;
 
   /// \brief Creates a crab's cfg builder manager
   void initCrabAnalysis(const llvm::Module &M, seadsa::GlobalAnalysis &dsa,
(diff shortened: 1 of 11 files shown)


include/seahorn/Transforms/Instrumentation/CrabAnalysisPass.hh
```diff
@@ -0,0 +1,37 @@
...
+
+namespace clam {
+class CrabBuilderManager;
+class InterGlobalClam;
+} // namespace clam
+
+namespace seahorn {
+using namespace llvm;
+
+class CrabAnalysisPass : public ModulePass {
```
caballa, line 16: Don't set `CrabAnalysisPass` as a LLVM pass. 
Since a while we try to rely on the pass manager as little as possible. 
Most of the code here is still fine. The only changes is to remove is `public Module Pass` and `getAnalysisUsage`. Then, the caller of this code (`CrabLowerIsDeref.cc`) which needs  to be a LLVM pass it will make sure that all the requirements of `CrabAnalysisPass` (if any) are fullfilled.

lib/Transforms/Scalar/CrabLowerIsDeref.cc
```diff
@@ -0,0 +1,115 @@
...
+  StringRef getPassName() const override { return "CrabLowerIsDeref"; }
+};
+char CrabLowerIsDeref::ID = 0;
+
+} // namespace
+
+bool CrabLowerIsDeref::runOnFunction(Function &F) {
+  LOG("crab-isderef", errs()
+                          << "Start Running Crab lowering is_deref checks\n";);
+  m_crab_ptr = getAnalysis<CrabAnalysisPass>().getCrab();
```
caballa, line 46: Don't use the pass manager to get a Crab analysis. (see my comments in `CrabAnalysisPass.cc`

lib/Transforms/Scalar/CrabLowerIsDeref.cc
```diff
@@ -0,0 +1,115 @@
...
+    deadCalls.clear();
+  }
+  LOG("crab-isderef", errs() << "Crab lowering is_deref checks complete\n";);
+  return Changed;
+}
+
+const llvm::ConstantRange
+CrabLowerIsDeref::getCrabInstRng(const llvm::Instruction &I) {
+  unsigned IntWidth = I.getType()->getIntegerBitWidth();
+  if (!m_crab_ptr)
```
caballa, line 79: If `m_crab_ptr` is null we shouldn't  iterate over all instructions anyway.

lib/Transforms/Scalar/CrabLowerIsDeref.cc
```diff
@@ -0,0 +1,115 @@
...
+
+bool CrabLowerIsDeref::runOnFunction(Function &F) {
+  LOG("crab-isderef", errs()
+                          << "Start Running Crab lowering is_deref checks\n";);
+  m_crab_ptr = getAnalysis<CrabAnalysisPass>().getCrab();
+  auto &SBI = getAnalysis<SeaBuiltinsInfoWrapperPass>().getSBI();
+
+  bool Changed = false;
+
+  SmallVector<CallInst *, 32> deadCalls;
```
caballa, line 51: If you want to remove all instructions once you process a block then move this definition after line 52 and then no need to call `deadCalls.clear()`

include/seahorn/Transforms/Instrumentation/CrabAnalysisPass.hh
LinerSu: Yes. I refactor this class to remove the LLVM pass.

include/seahorn/Transforms/Instrumentation/CrabAnalysis.hh
```diff
@@ -0,0 +1,43 @@
...
+} // namespace clam
+
+namespace seahorn {
+using namespace llvm;
+
+class CrabAnalysis {
+  //// \brief crab's cfg builder manager
+  std::unique_ptr<clam::CrabBuilderManager> m_cfg_builder_man;
+  //// \brief crab instance to solve alloc bounds
+  std::shared_ptr<clam::InterGlobalClam> m_crab;
```
caballa, line 26: why a `shared_ptr`? You can use `unique_ptr` and return `clam::InterGlobalClam&` in `getCrab`

lib/Transforms/Scalar/CrabLowerIsDeref.cc
```diff
@@ -0,0 +1,149 @@
...
+    return ConstantInt::getTrue(C);
+  } else if (crabDerefResult.isSingleElement()) {
+    // Crab inferred is_deref call is either true or false
+    Stats::count("crab.pp.isderef.solve");
+    return crabDerefResult.getSingleElement()->getBoolValue()
+               ? ConstantInt::getTrue(C)
+               : ConstantInt::getFalse(C);
+  } else {
+    // Crab cannot know the is_deref result
+    const llvm::DebugLoc &dloc = IsDerefCall->getDebugLoc();
```
caballa, line 135: Move lines 135-138 inside `LOG`

lib/seahorn/BvOpSem2.cc
```diff
@@ -799,7 +799,12 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemVisitorBase {
         crabSolved = true;
       } else {
         Stats::count("crab.isderef.not.solve");
-        LOG("opsem-crab", MSG << "crab cannot solve: " << *inst;);
+        const llvm::DebugLoc &dloc = inst->getDebugLoc();
```
caballa, line 802: Move these lines inside `LOG`

include/seahorn/Transforms/Instrumentation/CrabAnalysis.hh
```diff
@@ -0,0 +1,43 @@
...
+                        llvm::TargetLibraryInfoWrapperPass &tliPass);
+  /// \brief Run crab analysis
+  void runCrabAnalysis();
+
+public:
+  CrabAnalysis() {}
+  void runCrabAnalysisOnModule(const llvm::Module &M,
+                               seadsa::GlobalAnalysis &dsa,
+                               llvm::TargetLibraryInfoWrapperPass &tliPass);
+  clam::InterGlobalClam &getCrab() { return *m_crab; }
```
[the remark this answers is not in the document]
LinerSu: Sorry. For this one I think I cannot return a constant object since querying `range` by clam is not marked as a constant method. See:
https://github.com/seahorn/clam/blob/master/include/clam/ClamQueryAPI.hh#L31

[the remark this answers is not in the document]
caballa: You can still add the `const` version. The compiler will choose the non-const version. 
I think the methods exposed by `ClamQueryAPI` are not marked as `const` because they modify a cache. But C++ provides the `mutable` keyword exactly for those cases. So you add the const version and I will try to make const the methods from `ClamQueryAPI` meanwhile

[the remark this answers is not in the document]
LinerSu: Ok. I made it by using `const_cast` for now.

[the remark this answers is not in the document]
caballa: No, this is not right. You keep both versions the `const` and the `non-const` and the compiler chooses the right one. Please just keep these two:

```
const clam::InterGlobalClam &getCrab() const { return *m_crab;}
clam::InterGlobalClam &getCrab()  { return *m_crab;}

```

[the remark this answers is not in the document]
caballa: So don't use `const_cast`

caballa (approved):
For me it's ready to merge.

PR #443, opened 2022-08-02
feat: flag to print spacer statistics

igcontreras:


lib/seahorn/HornSolver.cc
```diff
@@ -144,7 +148,8 @@ bool HornSolver::runOnModule(Module &M) {
   params.set(":spacer.max_num_contexts", PdrContexts);
   params.set(":spacer.elim_aux", true);
   params.set(":spacer.reach_dnf", true);
-  // params.set ("print_statistics", true);
+  if (PrintSolverStatistics)
+    params.set("print_statistics", true);
```
agurfinkel, line 152: this should be `:spacer.print_statistics` at least to be consistent with the rest of the flags

igcontreras: Do you mean the name of the parameter? With `:spacer.print_statistics` nothing is output in my current version of z3. I think `print_statistics` is global.
```
$ z3 --version
Z3 version 4.8.10 - 64 bit
$ z3 -p | grep statistics
    print_statistics (bool) (default: false)
```

agurfinkel: hmm. then keep it as is

PR #456, opened 2022-10-01
fix(doctests): now compiles with clang-12

priyasiddharth:
upgrade doctests.h and workaround macro redef clash
test: can build seahorn container which depends on doctests building successfully.


priyasiddharth:
@agurfinkel PTAL.

include/seahorn/Support/SeaLog.hh
```diff
@@ -16,6 +16,7 @@
   ::seahorn::warn_ostream(llvm::errs())                                        \
       .prefix("Info: ")                                                        \
       .suffix(FLOC(__FILE__, __LINE__))
+#ifndef DOCTEST_LIBRARY_INCLUDED
```
agurfinkel, line 19: why would we want strange dependence on doctest inside our code?

seems like a wrong way to fix this problem. Just 'undef INFO` in unit tests before doctest is included.

units/doctest.h
```diff
@@ -1,25 +1,33 @@
-// ======================================================================
+// DEAR SEAHORN DEV,
```
agurfinkel, line 1: I see that you updated doctest to the latest version. This is good. However, commit message does not say that.

Do not change this file. At all.
All changes should be outside the file. Talk to me if there is something that you think you cannot do that way.

agurfinkel (changes requested):
the functionality is good, but the way it is done is not

units/doctest.h
priyasiddharth: Couple of points.
1. Doctest does not believe in versions and releases. They ask to `wget` at head. If it helps, I can add the commit-hash info in the commit message.
2. `doctest.h` does not follow this project's clang format. The current pre-merge check uses `clang-format-diff` and does not respect `.clang-format` files. Thus, there is no easy way to exclude this file. I would suggest force merging over the pre-merge failure since this file will be upgraded rarely. The alternative is what I have done here.

I will block further changes to this PR pending reply.

agurfinkel: I would be happy if commit message says "updated doctest to current master". Then I know what to expect.

working around the warning is the way to go. It is our setup issue, not doctest issue. Easier if whenever we use other people code we preserve format and use the code without any changes at all.

priyasiddharth: Removed all changes to doctest (only upgraded to HEAD). 
Made include order consistent as per https://seahornteam.slack.com/archives/DPND2KEJU/p1665394990646389

priyasiddharth:
The clang-format failure is expected since `doctest.h` does not follow project format. We will ignore the clang-format failure as it does not concern project code. 

@agurfinkel PTAL

include/seahorn/Support/SeaLog.hh
priyasiddharth: done

include/seahorn/Support/SeaLog.hh
```diff
@@ -16,11 +16,13 @@
   ::seahorn::warn_ostream(llvm::errs())                                        \
       .prefix("Info: ")                                                        \
       .suffix(FLOC(__FILE__, __LINE__))
+#undef INFO // workaround(test): to avoid name clash with INFO in doctest.h
```
agurfinkel, line 19: I guess this is acceptable, but I would much rather have a simpler fix.

1. no changes in doctest or seahorn files
2. new header 'sea_doctest.h' that
        1. undefs INFO, WARN, ERR
        2. includes doctest
 3. in all tests, include doctest last

priyasiddharth: done

units/units_z3.cpp
```diff
@@ -16,7 +17,8 @@ static Expr mkIntConst(const std::string name, ExprFactory &efac) {
   return bind::intConst(mkTerm(name, efac));
 }
 
-static Expr mkBvConst(const std::string name, ExprFactory &efac, unsigned width) {
+static Expr mkBvConst(const std::string name, ExprFactory &efac,
```
agurfinkel, line 20: why the extra new line?

overall, good, please merge

priyasiddharth: IDE auto formatted presumably using `.clang-format`.

PR #458, opened 2022-10-31
fix(docker): move docker to jammy and llvm12

priyasiddharth:
z3 and yices is not updated in this commit. They will be upgraded separately.

tested: All three containers were built locally. Final check is left for SEAHORN maintener(s) with write access to SEAHORN dockerhub.


agurfinkel (approved):
LGTM.

Can we have container for llvm14?

PR #460, opened 2022-11-03
fix(dfcoi): add back needed instructions in coi

priyasiddharth:
Also add asserts in opsem that help when things go awry

tested: all tests in test/opsem2 now pass


agurfinkel (approved):
LGTM

PR #472, opened 2022-11-28
feat(z3_solver): print reason for unknown

priyasiddharth:


include/seahorn/Expr/Smt/Z3.hh
```diff
@@ -612,9 +611,17 @@ public:
...
 
+  std::string getReasonUnknown() {
+    return std::string(Z3_solver_get_reason_unknown(ctx, solver));
+  }
+
   boost::tribool solve() {
     boost::tribool res = z3l_to_tribool(Z3_solver_check(ctx, solver));
     ctx.check_error();
+    if (indeterminate(res)) {
+      llvm::errs() << "Reason (for unknown status) : " << getReasonUnknown()
```
agurfinkel, line 622: should be under `LOG`. Probably don't want to print this from inside solve. This is client dependent.

agurfinkel (approved):

include/seahorn/Expr/Smt/Z3.hh
agurfinkel: is this being addressed and PR merged?

PR #477, opened 2022-12-16
Isalloc2

priyasiddharth:


lib/Transforms/Instrumentation/FatBufferBoundsCheck.cc
```diff
@@ -75,6 +75,10 @@ static cl::opt<bool> UseFatVac("horn-bnd-chk-vac",
                                cl::desc("Use vacuity check for bounds checks"),
                                cl::init(true));
 
+static cl::opt<bool> addIsAllocCheck("add-is-alloc-check",
```
agurfinkel, line 78: option should start with `horn-` or some other prefix that is passed through the cli scripts.

I would use `horn-` because of the other options in this file. We can normalize option prefixes later on.

agurfinkel (approved):
Except for option name, this looks great.

lib/Transforms/Instrumentation/FatBufferBoundsCheck.cc
priyasiddharth: Fixed

priyasiddharth:
Ignoring `crab_is_deref_2.c` failure since it is unrelated and being tracked at https://github.com/seahorn/seahorn/issues/478

PR #479, opened 2023-03-03
feat: add flag to UNignore defined fn in promote verifier

priyasiddharth:
Defined fns are ignored by default for lifting to verifier calls


agurfinkel (approved):
LGTM

PR #484, opened 2023-05-10
Updated README, set LLVM version to 14, added instruction on how to f…

BorisJancicUWaterloo:
…ix problem with libpolly when using apt.LLVM.org version 14

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,8 +245,8 @@ are:
 The easiest way to get started with SeaHorn is via a docker distribution. 
 
 ```shell
-$ docker pull seahorn/seahorn-llvm10:nightly
-$ docker run --rm -it seahorn/seahorn-llvm10:nightly
+$ docker pull seahorn/seahorn-llvm14:nightly
+$ docker run --rm -it seahorn/seahorn-llvm14:nightly
 ```
 
 Start with exploring what the `sea` command can do:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -267,6 +267,18 @@ We at least will update the blog post to indicate that it is not expected to
 work with the latest version of the code.
 
 
+You can also manually install by:
+
+Following the instructions in the Docker file
+Dockerfile: [`docker/seahorn-builder.Dockerfile`](docker/seahorn-builder.Dockerfile).
+
+If using LLVM-14 from apt.LLVM.org (version with package bugs)
+Fist running the below commnad to fix an issue with LLVM-14 and libpolly-14
```
igcontreras, line 276: Fist -> First, commnad -> command

agurfinkel, line 276: This will not work. Give complete sequence of commands. It would be something along the line of
```
./llvm.sh 14 && apt download ... && dpkg ... 
```
First install `clang-14` and the rest of llvm, then force install missing `libpolly` library.

README.md
```diff
@@ -267,6 +267,18 @@ We at least will update the blog post to indicate that it is not expected to
...
+And then run the commands in the docker above
```
agurfinkel, line 277: do not start a sentence with an *And*

agurfinkel (changes requested):
typos, commit message, and a little issue with instructions

PR #485, opened 2023-05-11
fix: updated installation instructions in READMEInstructions f…

BorisJancicUWaterloo:
…or mannual installation \n \n Instructions to fix potential issue with libpolly


agurfinkel (changes requested):
the text in README is good. 

The commit message still needs to be fixed. Somehow new lines got changed to `\n`

PR #486, opened 2023-05-11
fix: updated installation instructions in README

BorisJancicUWaterloo:
Instructions for manual installation
Instructions to fix potential issue with libpolly

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,6 +266,21 @@ phased out, new things are added. If you find problems in the blog, let us know.
 We at least will update the blog post to indicate that it is not expected to
 work with the latest version of the code.
 
+You can also manually install by:
+
+Following the instructions in the Docker file
+Dockerfile: [`docker/seahorn-builder.Dockerfile`](docker/seahorn-builder.Dockerfile).
+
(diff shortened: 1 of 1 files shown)


agurfinkel (approved):

agurfinkel:
Unfortunately, we are only familiar with Ubuntu and cannot provide installation instructions on other platforms.
If installation is a challenge, we suggest to use our nightly docker container that has the tool installed. Note that there is also a docker container of https://github.com/seahorn/verify-c-common that comes with seahorn and a verification case study.

Installation on other platforms should be similar. We require LLVM14. You can compile LLVM14 from sources, or use a binary distribution. One of the binary distributions for Ubuntu has an issue and the PR above provides instructions on how to fix it. The fix, however, is not necessary for the default LLVM14 packages on Ubuntu. So it might not be necessary for other distributions as well.

SeaHorn itself does not depend on libpolly, but seems that LLVM does. It was easier to make the dependency available then trace how and where it is used.

aliraeisdanaei:
That's fair. Thank you.

PR #498, opened 2023-07-18
Print placeholder for large alloca expressions

thomashart17:


agurfinkel:
@caballa can you check that crab test failure is not related to this commit

caballa:
Looking at it. The problem is not related to this commit.

PR #504, opened 2023-07-25
fix(gsa): generate undef value when if both incoming undef

priyasiddharth:
Currently we assert that only one of the incoming values must be undef, this is only caught in debug mode. Rust compiler appears to generate undef values on both incoming edges. As such, we should deal with gracefully. This is the simplest attempt at doing so.


agurfinkel:
In Rust, undef is used much more than in C to represent not yet set value. It flows through the program, but is guaranteed to not be used. 

This causes issues like this. 

The fix is good.

PR #520, opened 2023-09-05
feat(ci): add nightly test of seahorn main branch

priyasiddharth:
don't enable push to docker for main branch

### DIFF
--- a/.github/workflows/seahorn-main.yml
+++ b/.github/workflows/seahorn-main.yml
@@ -1,7 +1,7 @@
-# workflow to build and test seahorn dev14 every night.
-# This version of seahorn will be pushed to docker.
+# workflow to build and test seahorn main branch every night.
+# This version of seahorn will NOT be pushed to docker.
 
-name: Nightly
+name: Nightly-main-nopush
 
 # Controls when the action will run.
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/seahorn-main.yml
```diff
@@ -0,0 +1,52 @@
...
+    steps:
+      - name: Log manual trigger info
+        if: ${{ github.event_name == 'workflow_dispatch' }}  # only if triggered manually
+        run: |
+          echo "Reason for manual trigger: ${{ github.event.inputs.TriggerReason }}"
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - name: Check Out Repo
+        uses: actions/checkout@v2
+        with:
+          ref: dev14  # only checkout dev14
```
caballa, line 36: Not sure what I does but shouldn't be `main` ?

agurfinkel (approved):

.github/workflows/seahorn-main.yml
priyasiddharth: I think @caballa is right. We want to test `main`. `dev14` is already tested. However, recently someone complained that `main` wasn't building -- hence this change. I'll update the branch.

agurfinkel:
is this stuck or should be merged?

PR #522, opened 2023-09-08
fix(opsem): fix ptrEq in extrawidememmgr

priyasiddharth:
ptrEq should also take ptr size into account. Now it does.


LinerSu:
LGTM.

agurfinkel (approved):

Issue #529, opened 2023-09-24
Compilation error on main branch


Nivous:
**Describe the bug**
Compilation error of seahorn on main branch

**To Reproduce**
Just compile normally like described in the readme

**Logs, Stacktraces, and Screenshots**
These are the compilation errors I recieve:

[142/280] Building CXX object lib/seahorn/CMakeFiles/seahorn.LIB.dir/Houdini.cc.o
FAILED: lib/seahorn/CMakeFiles/seahorn.LIB.dir/Houdini.cc.o 
[...]

PR #530, opened 2023-09-24
Bug Fix: Including list in houdini header

Nivous:
Fix to issue #529

### DIFF
--- a/include/seahorn/Houdini.hh
+++ b/include/seahorn/Houdini.hh
@@ -12,6 +12,8 @@
 #include "seahorn/Expr/Smt/EZ3.hh"
 #include "seahorn/HornClauseDBWto.hh"
 
+#include <list>
+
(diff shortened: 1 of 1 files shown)


caballa:
Thank you for opening a PR!

lint doesn't like the formatting for your commit message. 
Maybe something like "fix: include header file" will make it happy.
Also, would you mind to re-open the PR against `dev14`? `dev14` is our development branch. `main` gets updated from `dev14` once in a while.

Nivous:
> Thank you for opening a PR!
> 
> lint doesn't like the formatting for your commit message. Maybe something like "fix: include header file" will make it happy. Also, would you mind to re-open the PR against `dev14`? `dev14` is our development branch. `main` gets updated from `dev14` once in a while.

Hi
no problem, will do

PR #536, opened 2023-10-01
feat: Added resetDB function to HornClauseDB

Nivous:
Done in order to allow reseting the DB while keeping the same object.
I need this function as part of my efforts to support k-saftey.
I want to reset the DB for a run I do after HornifyModule but Horn write takes it's DB from HornifyModule, so I need to reset this DB while keeping the same object.


Nivous:
is it possible to add it?

Nivous:
Ping

PR #543, opened 2023-10-23
feat(opsem): symb memcpy now simplifies len

priyasiddharth:
Simplify len expression and if found to be a number, delegate to a concrete mempy, else continue with symbolic one. Also do not add alignment constraint on symb len if ignoreAlignment is set.

### DIFF
--- a/lib/seahorn/BvOpSem2RawMemMgr.cc
+++ b/lib/seahorn/BvOpSem2RawMemMgr.cc
@@ -642,6 +642,17 @@ RawMemManagerCore::MemValTy RawMemManagerCore::MemSet(PtrTy ptr, Expr _val,
 RawMemManagerCore::MemValTy RawMemManagerCore::MemSet(PtrTy ptr, Expr _val,
                                                       Expr len, MemValTy mem,
                                                       uint32_t align) {
+  auto simplifiedLength = len;
+  if (m_ctx.shouldSimplify()) {
+    simplifiedLength = m_ctx.simplify(len);
+  }
+  if (m_ctx.alu().isNum(simplifiedLength)) {
+    unsigned numBytes = m_ctx.alu()
+                            .toNum(simplifiedLength)
+                            .get_ui(); /* assume: len should >= 0 */
+    return MemSet(ptr, _val, numBytes, mem, align);
+  }
+
   auto *memset = dyn_cast<llvm::MemSetInst>(&m_ctx.getCurrentInst());
   LOG(
       "opsem", if (memset == NULL) {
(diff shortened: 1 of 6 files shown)


priyasiddharth:
@agurfinkel FYI,

In fixing the CI breakage, i discovered that we were delegating to symbolic memcpy for simple numbers wrapped in expressions. 

Thus I have added a simplification step for symbolic `memcpy` and `memset` that delegates to a concrete version of the functions if the length can be simplified to a number.

lib/seahorn/BvOpSem2RawMemMgr.cc
```diff
@@ -642,6 +642,17 @@ RawMemManagerCore::MemValTy RawMemManagerCore::MemSet(PtrTy ptr, Expr _val,
 RawMemManagerCore::MemValTy RawMemManagerCore::MemSet(PtrTy ptr, Expr _val,
                                                       Expr len, MemValTy mem,
                                                       uint32_t align) {
+  auto simplifiedLength = len;
+  if (m_ctx.shouldSimplify()) {
```
agurfinkel, line 646: not sure the simplification should depend on this flag.

This flag is for aggressive simplification of memory state during vcgen. I would imagine that simplification in this step should be done aggressively at all time.

If this is a problem, might restrict simplifications only to cases where `len` has small dag size. 

However, I do believe that simplifying all the time should be doable.

priyasiddharth: I see what you mean. However, currently `shouldSimplify` is also used elsewhere. For example in simplifying isDeref: https://sourcegraph.com/github.com/seahorn/seahorn@dev14/-/blob/lib/seahorn/BvOpSem2ExtraWideMemMgr.cc?L122

Having a single flag to turn simplification on/off is a useful tool for debugging. If we want to keep a memory-state only simplification flag then should be break it into two flags -- `shouldSimplify` and `shouldSimplifyMemory` and only simplify memory if both are on.

agurfinkel: Yes, that's another incorrect use of that option. The external documentation for the option is what you call `shouldSimplifyMemory`. 

So perhaps you need to add another flag. 

We find that in some Rust verification aggressive simplification is very very expensive. So we turn it off. However, we still would like is_deref and memcpy checks to be simplified!

priyasiddharth: Added a new flag `shouldSimplifyNonMem`.

PR #544, opened 2023-11-21
feat: add support for btor2 witness generation

jetafese:
Adding a set of special functions for witness generation from seahorn cex

### DIFF
--- a/include/seahorn/seahorn.h
+++ b/include/seahorn/seahorn.h
@@ -22,7 +22,7 @@ extern void __VERIFIER_error(void);
 extern void __VERIFIER_assume(int);
 extern void __SEA_assume(bool);
 
-extern void __VERIFIER_assert(bool);
+extern void __VERIFIER_assert(bool, int property);
 extern void __VERIFIER_assert_not(bool);
 extern void __VERIFIER_assert_if(bool, bool);
 /**
--- a/sea-rt/seahorn.cpp
+++ b/sea-rt/seahorn.cpp
@@ -29,6 +29,11 @@ void __VERIFIER_assume(int x) {
   assert(x);
 }
 
+void __VERIFIER_assert(bool x, int property) {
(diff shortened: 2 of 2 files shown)


include/seahorn/seahorn.h
```diff
@@ -22,7 +22,7 @@ extern void __VERIFIER_error(void);
 extern void __VERIFIER_assume(int);
 extern void __SEA_assume(bool);
 
-extern void __VERIFIER_assert(bool);
+extern void __VERIFIER_assert(bool, int property);
```
agurfinkel, line 25: I think is is not backward compatible with existing clients of assert. 

You need to add a new function with new name. Not just change an existing one. 

Also this is probably the wrong place to add it unless you also change seahorn to recognize and treat it in some way.

sea-rt/seahorn.cpp
```diff
@@ -29,10 +29,23 @@ void __VERIFIER_assume(int x) {
...
+void __VERIFIER_assert(bool x, int property) {
+  sealog("[sea] __VERIFIER_assert was called for property: %d,\n", property);
+  assert(x);
+}
+
 void __SEA_assume(bool x) {
   assert(x);
 }
 
+void btor2mlir_print_input_num(unsigned num, unsigned value, unsigned width) {
```
agurfinkel, line 41: this definitely seems btor2mlir related and not seahorn related. 

A better solution is to move these functions into btor2mlir and provide them as extra library to seahorn during cex generation. 

exe cex command should be able to take multiple files and link them together. 

you are relying on seahorn linking it's rt by default. But this is not the right way to link desired code in

agurfinkel (changes requested):
Don't think this is the right place for this functionality

sea-rt/seahorn.cpp
jetafese: Sounds good. I'll do that instead.

PR #547, opened 2024-02-01
fix expr typechecker and add errorbinding to record error details

priyasiddharth:


include/seahorn/Expr/ExprOpBinder.hh
```diff
@@ -145,7 +145,8 @@ static inline bool binderCheck(Expr exp, TypeChecker &tc,
   // makes sure that all of the binders arguments are constants and store their
   // types
   for (auto b = exp->args_begin(), e = exp->args_end() - 1; b != e; b++) {
-    if (!op::bind::IsConst()(*b))
+    (*b)->dump();
```
agurfinkel, line 148: remove debug `dump`

lib/seahorn/Smt/TypeChecker.cc
```diff
@@ -37,6 +39,10 @@ class TCVR {
     if (isOp<ERROR_TY>(type)) {
       foundError(exp);
       m_errorMap.insert({exp, m_errorExp});
+    } else if (isOp<ERRORBINDER>(type)) {
```
agurfinkel, line 42: I would avoid calling an Error type `ERRORBINDER` if only to avoid confusion between `OpBind` and `OpBinder`. We use the root `Bind` too much in the code and it has already confused you in the past.

agurfinkel (approved):
LGTM. 

My only comments is about the name of the error type. I would have used a simpler, less ambiguous name.

include/seahorn/Expr/ExprOpBinder.hh
priyasiddharth: done

lib/seahorn/Smt/TypeChecker.cc
priyasiddharth: I couldn't find a better name. I will give it some thought.

PR #575, opened 2025-05-26
Lower fshl and fshr intrinsics

LinerSu:
Add missing `fshl` and `fshr` intrinsic handlers in bvopsem2.

### DIFF
--- a/lib/seahorn/BvOpSem2.cc
+++ b/lib/seahorn/BvOpSem2.cc
@@ -1849,16 +1849,16 @@ public:
     case Intrinsic::fshl:
     case Intrinsic::fshr: {
       // fshl, fshr
-      // <ty> @llvm.fshl.<ty>(<ty> %low, <ty> %high, <ty> %shift)
-      // result = ( (%low  << %shift)        // left-shift low
-      //      | (%high >> (width-%shift)) )  // right-shift high to fill in
-      // <ty> @llvm.fshr.<ty>(<ty> %low, <ty> %high, <ty> %shift)
-      // result = ( (%high >> %shift)        // right-shift high
-      //      | (%low << (width − %shift)) ) // left-shift low to fill in
-      Expr low = lookup(*I.getOperand(0));   // low operand
-      Expr high = lookup(*I.getOperand(1));  // high operand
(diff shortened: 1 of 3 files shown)


lib/seahorn/BvOpSem2.cc
```diff
@@ -1846,6 +1846,40 @@ class OpSemVisitor : public InstVisitor<OpSemVisitor>, OpSemVisitorBase {
         setValue(I, res);
       }
     } break;
+    case Intrinsic::fshl:
+    case Intrinsic::fshr: {
+      // fshl, fshr
+      // <ty> @llvm.fshl.<ty>(<ty> %low, <ty> %high, <ty> %shift)
```
priyasiddharth: should this be `(<ty> %high, <ty> %low, <ty> %shift)`

https://releases.llvm.org/14.0.0/docs/LangRef.html#llvm-fshl-intrinsic

priyasiddharth:
It is also interesting why the program is generating these instructions. For example, one use `fshl` is for unaligned `memcpy`. I am wondering if this is for `memcpy` why the `memcpy` is unaligned in the first place. This comment is not blocking the PR.

lib/seahorn/BvOpSem2.cc
LinerSu: Yes, but the implementation is correct; only the parameter naming is incorrect. Just fixed it.

LinerSu:
> It is also interesting why the program is generating these instructions. For example, one use `fshl` is for unaligned `memcpy`. I am wondering if this is for `memcpy` why the `memcpy` is unaligned in the first place. This comment is not blocking the PR.

I did not observe the `fshl` instruction, but I did see `fshr`. Here is a sample program:
```c
#include <stdio.h>

typedef unsigned long ulong;
#define ROTATE_LEFT(x,r) (((x)<<(r)) | ((x)>>(64-(r))))

int main() {
    ulong a = 10;
    a = ROTATE_LEFT(a, 31);
    return 0;
}
```
The expression `(x) >> (64 - (r))` compiles to `%shr = lshr i64 %x_r, 33`.

priyasiddharth (approved):

PR #579, opened 2026-06-23
test: migrate regression tests to filecheck

priyasiddharth:
OutputCheck (PyPI 0.4.1, 2014) is unmaintained and breaks on Python 3.11+ because it uses the argparse FileType('rU') mode removed in 3.11. Replace it with the maintained, pip-installable filecheck package (a Python-native clone of LLVM's FileCheck) and convert all check directives to FileCheck syntax.

Conversion (default CHECK prefix and custom prefixes OLD, OLDO0, OLDO3, XFAIL_UNKNOWN, XFAIL_NOT_UNKNOWN):

- literal -L / -NOT-L directives become plain CHECK / CHECK-NOT, since FileCheck patterns are literal by default
- regex CHECK / CHECK-NOT / CHECK-NEXT patterns are wrapped in {{...}} so anchors and escapes keep their regex meaning
- RUN-line OutputCheck becomes filecheck; the %oc substitution is updated and the OutputCheck-only --comment flag dropped
- README.md, deps/Dockerfile, test/readme.md install filecheck

Verified on the LLVM-14 build: all CI suites pass (opsem, opsem2, mcfuzz, smc, cex, devirt, crab, bmc, plus widemem/extra-widemem/cexgen variants).

Claude-Session: https://claude.ai/code/session_0136umANGeF5NyZaU8aZYDMT


priyasiddharth:
@agurfinkel just FYI

PR #580, opened 2026-06-24
feat(llvm15): port SeaHorn to LLVM 15 (dev15)

priyasiddharth:
# Summary

Ports SeaHorn from LLVM 14 (`dev14`) to **LLVM 15**, establishing the `dev15` line. The bulk of the change is adapting to LLVM 15's **opaque pointers** and removed/renamed APIs, plus a new opt-in struct-`memcpy` lowering pass, and the CI/Docker move to clang-15 + GHCR.

Validated end-to-end against the **verify-c-common** unit-proof suite: **228/228 passing, 0 crashes** — up from 16/228 before the port (all 212 prior failures were `seapp` segfaults on opaque pointers).

## Porting changes (LLVM 15)

Split into focused, by-kind commits:

**Opaque pointers**
- `GlobalVariable`/unique-scalar types via `getValueType()` (pointee types are gone)
- `AllocaInst::getAllocatedType()` instead of `getType()->getElementType()`
- `getAlign().value()` for load/store alignment (`getAlignment()` removed)
- devirt: key alias sets by `FunctionType` (a function pointer has no pointee type)

**Removed / changed APIs**
- `getFreedOperand` instead of the removed `isFreeCall`
- drop the legacy `createWholeProgramDevirtPass()` (new-PM-only in LLVM 15; SeaHorn's own devirt does the work)
- add headers LLVM 15 no longer pulls in transitively

**Build**
- guard all Clam/Crab code behind `HAVE_CLAM` so SeaHorn builds without Clam
- configure for LLVM 15 / clang-15
- de-duplicate `sea` subcommands for Python 3.11 argparse

**Transform passes (opaque pointers)**
- `PromoteMalloc`: promote `malloc`/`new` to `alloca i8 x size` directly (malloc always returned `i8*`)
- `PromoteMemcpy`: lower whole-struct `memcpy` into field-wise loads/stores, recovering the struct type **only** from a GEP that indexes the exact memcpy pointer (sound). Gated behind a new flag **`--horn-promote-memcpy`** (off by default; verify-c-common opts in). `sea` forwards the flag to `seapp` and swallows it before `seahorn` (which doesn't define it).

### Why `--horn-promote-memcpy` matters

Field-wise lowering is a large BMC-backend speedup — raw `memcpy` produces hard array-copy SMT. verify-c-common (228 jobs, `ctest -j8`):

| | with flag | without (raw memcpy) |
|---|---|---|
| total wall | 469 s | 2104 s |
| byte_buf_eq | 2.3 s | timeout (>2000 s) |
| byte_buf_write_from_whole_string | 14.6 s | 427 s |
| result | **228/228** | 227/228 (1 timeout) |

## Follow-ups / notes

- The `buildpack-deps-seahorn:jammy-llvm15` base must be **published to GHCR once** (run the manual *Buildpack deps* workflow) before CI / Nightly / Coverage can go green — they now `FROM` it.
- That builder is `workflow_dispatch`, so it must live on the repo's **default branch** to be triggerable; its checkout should target `dev15` so it builds the LLVM-15 Dockerfile.
- Clam/Crab is compiled out for now (`HAVE_CLAM` off) — a separate Clam port is still needed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


agurfinkel (approved):
very nice! thank you

priyasiddharth:
@copilot resolve the merge conflicts in this pull request

priyasiddharth:
commitlint refused to run during laste push, therefore overriding and submitting.

